PR #4, opened 2018-04-10
rpc_address: remove c interface of dsn_address_t

shengofsun:
merge the structure of dsn_address_t to rpc_address.h, make code
less confused


include/dsn/tool-api/rpc_address.h
```diff
@@ -0,0 +1,243 @@
...
+
+namespace dsn {
+
+class rpc_group_address;
+class rpc_uri_address;
+
+class rpc_address
+{
+public:
+    static const rpc_address sInvalid;
```
neverchanje: Why use camel case naming here? May `invalid_address` be a preferred name?

include/dsn/tool-api/rpc_address.h
```diff
@@ -0,0 +1,243 @@
...
+    static uint32_t ipv4_from_host(const char *hostname);
+    static uint32_t ipv4_from_network_interface(const char *network_interface);
+    ~rpc_address();
+
+    rpc_address() = default;
```
neverchanje: Use constexpr here, and mark sInvalid as constexpr.

include/dsn/tool-api/rpc_address.h
```diff
@@ -0,0 +1,243 @@
...
+    uint32_t ip() const { return (uint32_t)_addr.v4.ip; }
+    uint16_t port() const { return (uint16_t)_addr.v4.port; }
+    void set_port(uint16_t port) { _addr.v4.port = port; }
+    rpc_group_address *group_address() const
+    {
+        return (rpc_group_address *)(uintptr_t)_addr.group.group;
+    }
+    rpc_uri_address *uri_address() const { return (rpc_uri_address *)(uintptr_t)_addr.uri.uri; }
+    bool is_invalid() const { return _addr.v4.type == HOST_TYPE_INVALID; }
+    void set_invalid() { clear(); }
```
neverchanje: Remove this?

include/dsn/tool-api/rpc_address.h
```diff
@@ -0,0 +1,243 @@
...
+        case HOST_TYPE_URI:
+            return strcmp(to_string(), r.to_string()) == 0;
+        case HOST_TYPE_GROUP:
+            return _addr.group.group == r._addr.group.group;
+        default:
+            return true;
+        }
+    }
+
+    bool operator!=(::dsn::rpc_address r) const { return !(*this == r); }
```
neverchanje: Remove this. Compiler automatically generates the inequality operator if equality has defined.

src/core/core/address.cpp
```diff
@@ -153,8 +155,91 @@ DSN_API uint32_t dsn_ipv4_local(const char *network_interface)
...
+{
+    clear();
+    _addr.group.type = HOST_TYPE_GROUP;
+    dsn::rpc_group_address *addr = new dsn::rpc_group_address(name);
+    // take the lifetime of rpc_uri_address, release_ref when change value or call deconstruct
+    addr->add_ref();
+    _addr.group.group = (uint64_t)addr;
+}
+
+rpc_address rpc_address::clone() const
```
neverchanje: I didn't see any difference between `clone` and `operator=`. Even if they are implemented differently, they are supposed to be the same. Remove `clone`?

src/core/core/address.cpp
```diff
@@ -57,15 +57,16 @@
...
 #include <dsn/tool-api/task.h>
-#include "group_address.h"
-#include "uri_address.h"
 
 namespace dsn {
-const rpc_address rpc_group_address::_invalid;
-}
+const rpc_address rpc_address::sInvalid;
 
 #ifdef _WIN32
```
neverchanje: We should no longer support win32. Remove this.

src/core/core/rpc_message.cpp
```diff
@@ -214,6 +214,13 @@ message_ex::message_ex()
 
 message_ex::~message_ex()
 {
+    // when recv a request, message_header is hidden ahead of buffer(see@create_receive_message
```
neverchanje: I don't understand what does "hidden ahead of buffer" mean? Could you make it clear? 
BTW, don't use abbreviation like "recv" in your comment. Always be careful of your grammar.

src/core/core/address.cpp
neverchanje: We should no longer support win32. See L71, L247, L256 and L36 of this file, L81 of src/core/tests/address.cpp.

src/core/core/address.cpp
shengofsun: clone is used for uri_resolver in the old code, currently it's indeed useless. Let's remove it.

src/core/core/rpc_message.cpp
```diff
@@ -214,6 +214,18 @@ message_ex::message_ex()
...
 {
+    // when receiving a request, the lifetime of message_header object is managed by
+    // vector "buffers" (see@create_receive_message function).
+    // if you call copy_and_prepare_send() function, then a new request will be created,
+    // but new request shares the same header with the old_request.
+    // so if old_request releases header,
+    // then new request's header will be invalid.
+    //
+    // so we don't release_buffer_header().
+    // however, this won't lead to any memory leak problems as long as the header is a POD type
```
qinzuoyan: 这里说"the header is a POD type"

src/core/core/rpc_message.cpp
```diff
@@ -214,6 +214,18 @@ message_ex::message_ex()
...
+    // see@ Attention of message_header
+
+    // release_buffer_header();
```
qinzuoyan: 这个函数没有真正被使用过，所以只是为了帮助写注释？

src/core/core/rpc_message.cpp
```diff
@@ -465,9 +478,18 @@ void message_ex::prepare_buffer_header()
...
 
+    // here, we should call placement new, because message_header contain variable that is not
+    // POD-type, such as rpc_address, so we should call these variable's construct
+    new (ptr)(message_header);
     header = (message_header *)ptr;
 }
 
+void message_ex::release_buffer_header()
+{
+    // message_header contain variable that is not POD-type, so we call these variable's destructor
```
qinzuoyan: 这里也说“message_header contain variable that is not POD-type”

src/core/core/rpc_message.cpp
```diff
@@ -465,9 +478,18 @@ void message_ex::prepare_buffer_header()
     this->_rw_offset = (int)sizeof(message_header);
     this->buffers.push_back(buffer);
 
+    // here, we should call placement new, because message_header contain variable that is not
```
qinzuoyan: 这里又说“message_header contain variable that is not POD-type”，与上面的注释不一致，容易confused

src/core/core/rpc_message.cpp
shengofsun: 对的

src/core/core/rpc_message.cpp
qinzuoyan: 还是写清楚点吧：however, this won't lead to any memory leak problems as long as the header can be thought of as POD type (because the from_address must be IPv4 address), refer to the ATTENTION of message_header.

shengofsun:
@qinzuoyan comments are fixed

qinzuoyan:
@shengofsun more comments.

src/core/core/rpc_message.cpp
shengofsun: 后来这个地方改过了，你看看新的注释，应该也能说清楚了

src/core/core/rpc_message.cpp
```diff
@@ -465,9 +479,20 @@ void message_ex::prepare_buffer_header()
...
     this->buffers.push_back(buffer);
 
+    // here we should call placement new,
+    // so the gpid & rpc_address can be initialized
+    new (ptr)(message_header);
+
     header = (message_header *)ptr;
 }
 
+void message_ex::release_buffer_header()
```
neverchanje: Remove this, I don't see any place using it.

src/core/core/rpc_message.cpp
```diff
@@ -214,6 +214,19 @@ message_ex::message_ex()
 
 message_ex::~message_ex()
 {
+    // when receiving a request, the lifetime of message_header object is managed by
```
neverchanje: 1. 建议把这段注释全删了
2. 把 `copy_and_prepare_send` 的 clone_content 参数也删了，我搜了一下没有哪里有用除了 false 以外的参数。并且改名叫 `forward_to_send`，因为它只在 `rpc_engine::forward` 里用。
3. 加以下注释在 `message_ex::buffers` 上面：

```
// For sending, buffers[0] = (memory of message_header), buffer[1] = (memory of request content).
// For receiving, buffers[0] = (memory of response content).
// In the case where request is forwarded from one server to another, the received request
// as well as its message header will be entirely moved into the new request to send.
// \see message_ex::copy_and_prepare_send and rpc_engine::forward
```

原来底下这段注释的目的是解释 message_header 使用 dsn::blob 来共享内存的原因，但如果只是为了 forward，我没有理解为什么还要用 blob？可不可以用 unique_ptr<char[]>?

shengofsun: 再提个pr来改吧，这个pr是rpc_address相关的，以及由于rpc_address改动而对潜在问题的相关改动。
在message上再纠结下去，超纲了。

src/core/core/rpc_message.cpp
shengofsun: 当前用这种方式注释这段代码没问题，你有了placement new，当然得有显式的析构函数。不加这段代码才容易让人感到疑惑

qinzuoyan:
LGTM

PR #22, opened 2018-04-19
simple_load_balancer: support black list

shengofsun:
add black list in simple_load_balancer, so as to
skip some nodes when adding secondary, which is very
useful for offline some nodes

### DIFF
--- a/src/dist/replication/meta_server/server_load_balancer.cpp
+++ b/src/dist/replication/meta_server/server_load_balancer.cpp
@@ -786,7 +786,7 @@ void simple_load_balancer::register_ctrl_commands()
 
     _ctrl_assign_secondary_black_list = dsn::command_manager::instance().register_app_command(
         {"lb.assign_secondary_black_list"},
-        "lb.assign_secondary_black_list [ip_list|clear]",
+        "lb.assign_secondary_black_list [<ip:port,ip:port,ip:port>|clear]",
         "control the assign secondary black list",
         [this](const std::vector<std::string> &args) {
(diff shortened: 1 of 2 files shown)


src/dist/replication/meta_server/server_load_balancer.cpp
```diff
@@ -786,35 +783,88 @@ void simple_load_balancer::register_ctrl_commands()
...
             }
         }
     }
     return result;
 }
 
+std::string
+simple_load_balancer::ctrl_assign_secondary_black_list(const std::vector<std::string> &args)
+{
+    std::string invalid_arguments("invalid_arguments");
```
qinzuoyan: invalid arguments

src/dist/replication/meta_server/server_load_balancer.cpp
```diff
@@ -786,35 +783,88 @@ void simple_load_balancer::register_ctrl_commands()
         "lb.assign_delay_ms [num | DEFAULT]",
         "control the replica_assign_delay_ms_for_dropouts config",
         [this](const std::vector<std::string> &args) { return ctrl_assign_delay_ms(args); });
+
+    _ctrl_assign_secondary_black_list = dsn::command_manager::instance().register_app_command(
+        {"lb.assign_secondary_black_list"},
+        "lb.assign_secondary_black_list [ip_list|clear]",
```
qinzuoyan: 不是ip_list，而是address_list。
改成： lb.assign_secondary_black_list [clear | address_list]  (where address_list is in format of 'ip1:port1,ip2:port2...')

PR #37, opened 2018-04-28
pegasus_server: cancel or wait background checkpoint when close replica

qinzuoyan:


src/dist/replication/lib/replica.cpp
```diff
@@ -380,7 +380,25 @@ void replica::close()
             name(),
             enum_to_string(status()));
 
-    if (nullptr != _checkpoint_timer) {
+    while (true) {
```
neverchanje: 加段注释吧：
```
// Wait until no background checkpoint tasks are scheduled.
```

PR #98, opened 2018-06-11
json: parse int as double; support true/false in json standard

shengofsun:


include/dsn/cpp/json_helper.h
```diff
@@ -207,19 +207,32 @@ inline bool json_decode(const JsonObject &in, std::string &str)
 inline void json_encode(JsonWriter &out, bool t) { out.Int(t ? 1 : 0); }
 inline bool json_decode(const JsonObject &in, bool &t)
 {
-    dverify(in.IsInt());
-    int ans = in.GetInt();
-    t = (ans != 0);
-    return true;
+    if (in.IsInt()) {
```
qinzuoyan: 也补上bool的单元测试吧

PR #119, opened 2018-07-05
perf_counter: code refactor

shengofsun:
1. merge code pieces from different position of source tree into directory of perf_counter.
2. remove the factory registry's mechanism, which is useless and subtle. Accordingly, several implements of perf_counter are removed, only the atomic one remains, which is the default implementation we used in production.


src/core/tools/common/shared_io_service.h
```diff
@@ -56,7 +56,7 @@ class shared_io_service : public utils::singleton<shared_io_service>
         _io_service_worker_count =
             (int)dsn_config_get_value_uint64("core",
                                              "timer_service_worker_count",
-                                             1,
+                                             2,
```
neverchanje: 这个为什么要修改默认值?

shengofsun: 线上的默认值是这个

PR #136, opened 2018-07-17
utility: refactor on blob

neverchanje:
- `blob::create_from_bytes`: Support zero-copy construction of blob based on implementation 
https://github.com/XiaoMi/pegasus/blob/master/src/base/pegasus_value_schema.h#L55

- In addition we also support a copy-version of `blob::create_from_bytes`, which memcpy an allocated `char*` into the buffer of blob. This is widely used in rDSN (cases: [fds_service.cpp](https://github.com/XiaoMi/rdsn/blob/master/src/dist/block_service/fds/fds_service.cpp#L699), [thrift_message_parser.cpp](https://github.com/XiaoMi/rdsn/blob/master/src/core/tools/common/thrift_message_parser.cpp#L128)) but not generalized as utility.

### DIFF
--- a/include/dsn/utility/blob.h
+++ b/include/dsn/utility/blob.h
@@ -72,7 +72,7 @@ public:
     }
 
     /// Create shared buffer without copying data.
-    static blob create_from_bytes(std::string bytes)
+    static blob create_from_bytes(std::string &&bytes)
     {
         auto s = new std::string(std::move(bytes));
         std::shared_ptr<char> buf(const_cast<char *>(s->data()), [s](char *) { delete s; });
--- a/include/dsn/utility/blob.h
+++ b/include/dsn/utility/blob.h
@@ -41,22 +41,12 @@ class blob
 public:
     constexpr blob() = default;
 
-    blob(const std::shared_ptr<char> &buffer, unsigned int length)
-        : _holder(buffer), _buffer(_holder.get()), _data(_holder.get()), _length(length)
-    {
-    }
-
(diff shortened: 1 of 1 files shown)


include/dsn/utility/blob.h
```diff
@@ -72,6 +62,23 @@ class blob
...
+    /// NOTE: this operation is not efficient since it involves a memory copy.
+    static blob create_from_bytes(const char *s, size_t len)
+    {
+        std::shared_ptr<char> s_arr(new char[len], std::default_delete<char[]>());
+        memcpy(s_arr.get(), s, len);
+        return blob(std::move(s_arr), 0, static_cast<unsigned int>(len));
+    }
+
+    /// Create shared buffer without copying data.
+    static blob create_from_bytes(std::string bytes)
```
shengofsun: 为啥不用const std::string&

neverchanje: 因为我们的目的是把这个 string move 到 blob 里面, 从而做到 zero-copy, 如果你传的是 const std::string& 那它就是 copy 的了, 那其实你应该用 

```cpp
blob create_from_bytes(const char *s, size_t len)
```

shengofsun: 如果是zero copy的话，不是(std::string &&)更合适一点？因为外面一个string在调用(std::string bytes)的版本时候，也要copy一遍的。

neverchanje: 已改

PR #156, opened 2018-09-09
cli: refactor cli related code

shengofsun:
这个pr主要干的事情：
1. 以前cli是集成在core里面的。但这只是一个简单的rpc service，所以直接挪出来了。这样配置和core层的代码都好理解一些。
2. 以前command manager和rpc的处理是耦合的。现在解耦了，这样用别的方式来调用这些command也容易很多。（比如可以加个http的？)
3. cli的头文件和实现以前散落在不同地方，现在统一到了dist下。特别的，src/tools下面cli的一个小shell也放回到和cli模块一起。
4. 把cli_remote/cli_local这两个配置删了，相应的配置以后在新版本集群配置里也没必要了。

另外：
1. src/tools这个目录只剩下没用的repli了，这个pr就不动这个地方了。谁有空可以单独提个pr删一下。


include/dsn/dist/cli/cli.server.h
```diff
@@ -0,0 +1,60 @@
...
+ * THE SOFTWARE.
+ */
+
+#pragma once
+#include <iostream>
+#include <dsn/cpp/serverlet.h>
+#include <dsn/dist/cli/cli.code.definition.h>
+#include <dsn/dist/cli/cli.types.h>
+
+namespace dsn {
```
neverchanje: namespace 应该是 dsn::dist，不应该是 dsn？

neverchanje:
really nice job! but there're still something i doubt that need to fix: 

first, since the `_http_server` is placed in `meta_service_app`, could the `cli_service` be placed there too? could you adapt the code style with the `_http_server` so we can make further refactoring easier?

second, why the `cli_service` have to call `open_service` after its constructor, like:

```
    // start cli service before acquiring leader lock,
    // so that the command line call can be handled
    _cli_service = std::move(dsn::cli_service::create_service());
    _cli_service->open_service();
```

why not just:

```
: _cli_service(new cli_service())
```

and in the constructor you can "open the service" and don't have to split up the logic into two different functions: 

```
cli_service() : ::dsn::serverlet<cli_service>("cli") 
{
  register_async_rpc_handler(RPC_CLI_CLI_CALL, "call", &cli_service::on_call);
}
~cli_service() {
  unregister_rpc_handler(RPC_CLI_CLI_CALL);
}
```

src/dist/cli/shell/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-set(MY_PROJ_NAME dsn.cli.tool)
+set(MY_PROJ_NAME cli_sh)
```
neverchanje: call this dsn_cli_shell?

shengofsun:
@neverchanje 
1. "open_service" function for "xxx_service" is generated by the code generator, here we just keep the same with old conventions. We can modify this in a seperated pr after the code generator is refactored.
2. “putting the cli_service into meta_service and replica_stub” follows the old conventions, too. As all the services are handled by replica_stub & meta_service, only the "http_service" is an exception:-(

src/dist/cli/shell/CMakeLists.txt
shengofsun: 这个晚上我抽空改一下吧，白天没时间……

shengofsun:
@all has fixed based on comments

PR #165, opened 2018-09-23
misc: refactor cmake scripts to support gcov

neverchanje:
we refactor several places on cmake scripts to make the whole building process much simpler:
- remove `WARNING_ALL` option, because whatever the value is, it always add `-Wall` flag.
- remove `scripts/linux/misc.sh,` this is useless.
- don't generate `.matchfile`, useless.
- remove `MY_PROJ_BINDIRS` argument, useless.
- don't install useless stuff in rdsn/bin to rdsn/builder/output
- extract the installation process to `dsn_install_library` and `dsn_install_executable`, remove `MY_DO_INSTALL` argument, which makes things complicated.

 - Formerly we had a unfinished support for gcov, however we used lcov to generate report. lcov is useful but it is an old tool and compared to gcovr, it's not convenient enough. Now I change our tool to gcovr, which do everything only in one statement.

- unset `CMAKE_RUNTIME_OUTPUT_DIRECTORY`, `CMAKE_ARCHIVE_OUTPUT_DIRECTORY` and `CMAKE_LIBRARY_OUTPUT_DIRECTORY`. The problem is, when test executable, `dsn_nfs_test` for example, is copied to `builder/bin/dsn_nfs_test` (by `CMAKE_RUNTIME_OUTPUT_DIRECTORY`), gcovr can not find the corresponding `*.gcno` file because they are not in the same directory. 
More specifically, `dsn_nfs_test` is in `builder/bin/dsn_nfs_test`, while `main.cpp.gcno` is in `builder/src/dist/nfs/test/CMakeFiles/dsn_nfs_test.dir`. gcovr failed to find gcno file of `dsn_nfs_test` in `builder/bin/dsn_nfs_test` so it generated empty result for it.
To fix this we make `builder/bin/dsn_nfs_test` a soft link to `builder/src/dist/nfs/test`, and it works now.


qinzuoyan:
good job, looks nice to me

PR #174, opened 2018-10-15
valgrind: use function-local initialization to implement singleton

neverchanje:
Valgrind warns that the internal pointer of singleton is not deleted (`singleton::_instance`), to fix this we rewrite the singleton implementation.

function-local initialization: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2008/n2660.htm#Local

See this if performance issue is your concern: https://stackoverflow.com/questions/23829389/does-a-function-local-static-variable-automatically-incur-a-branch?noredirect=1&lq=1


include/dsn/utility/singleton.h
```diff
@@ -45,48 +37,12 @@ template <typename T>
...
-    static T &fast_instance() { return *_instance; }
-
-    static bool is_instance_created() { return nullptr != _instance; }
-
-protected:
-    static T *_instance;
-    static std::atomic<int> _l;
-
-private:
-    singleton(const singleton &);
```
shengofsun: nocopyable的feature没了

neverchanje: 已改

include/dsn/utility/singleton.h
```diff
@@ -24,19 +24,11 @@
...
- *
- * Revision history:
- *     xxxx-xx-xx, author, first version
- *     xxxx-xx-xx, author, fix bug about xxx
- */
-
 #pragma once
 
 #include <mutex>
 #include <atomic>
```
acelyc111: 这些include删了吧

PR #188, opened 2018-11-01
replica: miscellaneous refactoring and fixes

neverchanje:
- let user of prepare_list know its gpid and name. this knowledge will help debugging via printing gpid into log.
- remove useless codes
- add some utilities (pipeline, `dcheck_xxx` macros in fmt_logging, comparison operator for `rpc_holder`)


src/dist/replication/lib/prepare_list.cpp
```diff
@@ -70,7 +63,7 @@ void prepare_list::truncate(decree init_decree)
 error_code prepare_list::prepare(mutation_ptr &mu, partition_status::type status)
 {
     decree d = mu->data.header.decree;
-    dassert(d > last_committed_decree(), "%" PRId64 " VS %" PRId64 "", d, last_committed_decree());
+    dcheck_ge_replica(d, last_committed_decree());
```
vagetablechicken: 原来的代码是d大于last_committed_decree，这里是大于等于？

PR #189, opened 2018-11-07
scripts: if 64-bit, find_library() should search the lib64

vagetablechicken:


vagetablechicken:
if the system has both lib32 & lib64, find_library() may find the lib32 first and link the lib32 into our program.
So if the system is 64-bit, we should set FIND_LIBRARY_USE_LIB64_PATHS ON.

PR #239, opened 2019-04-01
fix perf_counter_info::timestamp

zhangyifan27:
fix perf_counter_info::timestamp to the timestamp when perf_counters::take_snapshot()

### DIFF
--- a/include/dsn/perf_counter/perf_counters.h
+++ b/include/dsn/perf_counter/perf_counters.h
@@ -147,7 +147,8 @@ private:
     };
     std::unordered_map<std::string, counter_snapshot> _snapshots;
 
-    int64_t _timestamp;  // in seconds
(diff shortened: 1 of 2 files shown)


include/dsn/perf_counter/perf_counters.h
```diff
@@ -146,6 +146,9 @@ class perf_counters : public dsn::utils::singleton<perf_counters>
         bool updated_recently{false};
     };
     std::unordered_map<std::string, counter_snapshot> _snapshots;
+
+    // timestamp in seconds when take snapshot of current counters
+    int64_t _timestamp;
```
neverchanje: 建议 `_timestamp_in_sec`

neverchanje:
Relates to https://github.com/XiaoMi/pegasus/issues/235

PR #249, opened 2019-05-06
dup: add duplication recovery on meta server

neverchanje:
MetaServer recovery loads metadata from zookeeper, and initializes internal data structure.

For duplication (`recover_from_meta_state`), it first reads zk on path ` <app_path>/duplication/<dup_id>` for full detail of duplication (`do_restore_duplication`), and if the information is valid, then it tries to restore every partition's progress (`do_restore_duplication_progress`), ie. `confirmed_decree`, concurrently.

Before it's allowed to serve request, meta-server must wait until the recovery completes, when all partitions has been restored.


src/dist/replication/meta_server/duplication/meta_duplication_service.cpp
```diff
@@ -231,5 +231,106 @@ meta_duplication_service::new_dup_from_init(const std::string &remote_cluster_na
...
+
+    // /<app>/duplication/<dupid>/<partition_idx>
+    //                       |         |-> confirmed_decree
+    //                       |
+    //                       |-> json of dup info
+
+    for (const auto &kv : _state->_exist_apps) {
+        std::shared_ptr<app_state> app = kv.second;
+        if (app->status != app_status::AS_AVAILABLE) {
+            return;
```
hycdong: Why return here?

src/dist/replication/meta_server/duplication/meta_duplication_service.cpp
```diff
@@ -231,5 +231,106 @@ meta_duplication_service::new_dup_from_init(const std::string &remote_cluster_na
...
+        }
+
+        _meta_svc->get_meta_storage()->get_children(
+            get_duplication_path(*app),
+            [this, app](bool node_exists, const std::vector<std::string> &dup_id_list) {
+                if (!node_exists) {
+                    // if there's no duplication
```
hycdong: I suggest printing log here, ddebug or dinfo

src/dist/replication/meta_server/duplication/meta_duplication_service.cpp
```diff
@@ -231,5 +231,106 @@ meta_duplication_service::new_dup_from_init(const std::string &remote_cluster_na
...
+{
+    for (int partition_idx = 0; partition_idx < app->partition_count; partition_idx++) {
+        std::string str_pidx = std::to_string(partition_idx);
+
+        // <app_path>/duplication/<dup_id>/<partition_index>
+        std::string partition_path = get_partition_path(dup, str_pidx);
+
+        _meta_svc->get_meta_storage()->get_data(
+            std::move(partition_path), [dup, partition_idx](const blob &value) {
+                if (value.size() == 0) {
```
hycdong: I guess that 'value.size() == 0' means that you can't find partition's duplication progress on zk. The type of 'value' is 'int64_t' indicating duplication progress.  I don't know if I understand it correct.

src/dist/replication/meta_server/duplication/meta_duplication_service.cpp
neverchanje: Should perform no additional behavior when the app does not have duplication.

src/dist/replication/meta_server/duplication/meta_duplication_service.cpp
neverchanje: right. I've added some comments here.

src/dist/replication/test/meta_test/unit_test/meta_duplication_service_test.cpp
```diff
@@ -45,6 +45,9 @@ class meta_duplication_service_test : public ::testing::Test
     void SetUp() override
     {
         _ms = make_unique<fake_receiver_meta_service>();
+        _ms->_failure_detector.reset(new meta_server_failure_detector(_ms.get()));
```
hycdong: refactor test according to meta_test_base.h

src/dist/replication/meta_server/duplication/meta_duplication_service.cpp
hycdong: I recommend add log here and return directly

PR #259, opened 2019-05-31
http: bugfix on http_message_parser

neverchanje:
bugfix:

- fix the case where URL is very long, `message->buffers.size()` will unexpectedly be 4, but not 3.
This case has been checked in `http_message_parser_test.parse_long_url`.

- fix the assertion where `message->buffers.size() == 3`. This is always true if everything works fine, but for robustness, the server will handle this case instead of being killed. 
This case has been checked in `http_message_parser_test.parse_bad_request`.

Check Lists:

- [x] Unit tests

- [x] Works fine on pprof heap profiling.

### DIFF
--- a/src/core/tools/http/http_server.cpp
+++ b/src/core/tools/http/http_server.cpp
@@ -74,7 +74,7 @@ void http_server::add_service(http_service *service)
 
 /*static*/ error_with<http_request> http_request::parse(message_ex *m)
 {
-    if (m->buffers.size() < 3) {
+    if (m->buffers.size() != 3) {
         return error_s::make(ERR_INVALID_DATA,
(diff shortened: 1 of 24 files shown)


src/core/tools/http/http_server.cpp
```diff
@@ -73,7 +74,10 @@ void http_server::add_service(http_service *service)
 
 /*static*/ error_with<http_request> http_request::parse(message_ex *m)
 {
-    dassert(m->buffers.size() == 3, "");
+    if (m->buffers.size() < 3) {
```
hycdong: 这里没有处理 m->buffers.size() > 3的情况，是在哪里处理这种情况的呢？

neverchanje: 我写错了，应该是 !=

PR #262, opened 2019-06-17
Revert "use c++14 mode for compilation (#257)"

neverchanje:
This reverts commit 89eb36095d63e6046e902cb97968c45d9939675f.

It's no good timing for this commit. And it also affects the client ABI compatibility with older gcc version.


bin/dsn.cmake
```diff
@@ -344,8 +336,14 @@ function(dsn_common_setup)
...
+    if(NOT ${COMPILER_SUPPORTS_CXX1Y})
+        message(FATAL_ERROR "You need a compiler with C++1y support.")
+    endif()
+
     dsn_setup_system_libs()
     dsn_setup_compiler_flags()
     dsn_setup_include_path()
     dsn_setup_link_path()
-endfunction(dsn_common_setup)
+endfunction(dsn_common_setup)
```
qinzuoyan: 行尾缺换行

PR #267, opened 2019-07-09
replication: fix cold backup bug & config loading

vagetablechicken:
1. cold backup will return err code instead of assert, when the
block_provider is unknown
2. if duplication and non-idempotent write are enabled together, abort
the process


vagetablechicken:
由于禁止duplication和non idempotent write同时打开，pegasus的配置和test还需要修改一下

vagetablechicken:
pegasus部分已经完成，这个pr可以合了

PR #272, opened 2019-07-18
fd: fix ingeter overflow

hycdong:
We reproduce failure detection bug (issue: https://github.com/XiaoMi/pegasus/issues/315) this week, it is caused by integer overflow. 
Master(meta server) will check lease periodically, like codes below:
```
now = dsn_now_ms();
{
    zauto_lock l(_lock);
    if(now - last_beacon_recv_time > grace_time){
        // master consider worker disconnected
    }
}
```

I found that last_beacon_recv_time might be larger than now during getting the lock, logs are below:
> **replica send beacon succeed**
> D2019-07-16 11:00:23.345 (1563246023345980449 20459) replica.     fd1.04080001000000bf: failure_detector.cpp:542:send_beacon(): send ping message, from[<replica_server_ip>], to[<meta_server_ip>], time[1563246023345]

> **meta receive beacon succeed**
> D2019-07-16 11:00:23.346 (1563246023346024792 7fe2)   meta.     fd1.02007fce003daeb6: failure_detector.cpp:344:on_ping_internal(): master <meta_server_ip> update **last_beacon_recv_time=1563246023346**

> **meta check error**
> E2019-07-16 11:00:23.346 (1563246023346609355 7fe1)   meta.     fd0.0201000000000003: failure_detector.cpp:280:check_all_records(): worker <worker_server_ip> disconnected, now=1563246023343, last_beacon_recv_time=1563246023346, **now-last_recv=-3**

> **replica receive meta reply succeed**
> D2019-07-16 11:00:23.346 (1563246023346207585 20458) replica.     fd0.0408000100069100: failure_detector.cpp:436:end_ping_internal(): worker <replica_server_ip> send beacon succeed, update last_send_time=1563246023345

As the logs show, the heart beat is normal, but the check is incorrect.

`now` and `last_beacon_recv_time` both are `uint64_t` type, if last_beacon_recv_time is larger than now, `now-last_beacon_recv_time` will be a very large integer. 

Add a comparison between `now` and `last_beacon_recv_time` is the simplest way to fix this integer overflow bug.

### DIFF
--- a/src/dist/failure_detector/failure_detector.cpp
+++ b/src/dist/failure_detector/failure_detector.cpp
@@ -283,7 +283,8 @@ void failure_detector::check_all_records()
             worker_record &record = itq->second;
 
             if (record.is_alive != false &&
-                now - record.last_beacon_recv_time > _grace_milliseconds) {
+                now - record.last_beacon_recv_time > _grace_milliseconds &&
+                is_time_greater_than(now, record.last_beacon_recv_time)) {
                 derror("worker %s disconnected, now=%" PRId64 ", last_beacon_recv_time=%" PRId64
                        ", now-last_recv=%" PRId64,
                        record.node.to_string(),
--- a/src/dist/failure_detector/failure_detector.cpp
+++ b/src/dist/failure_detector/failure_detector.cpp
@@ -282,9 +282,11 @@ void failure_detector::check_all_records()
         for (; itq != _workers.end(); itq++) {
             worker_record &record = itq->second;
 
+            // we should ensure now is greater than record.last_beacon_recv_time to aviod integer
+            // overflow
             if (record.is_alive != false &&
-                now - record.last_beacon_recv_time > _grace_milliseconds &&
-                is_time_greater_than(now, record.last_beacon_recv_time)) {
+                is_time_greater_than(now, record.last_beacon_recv_time) &&
+                now - record.last_beacon_recv_time > _grace_milliseconds) {
                 derror("worker %s disconnected, now=%" PRId64 ", last_beacon_recv_time=%" PRId64
                        ", now-last_recv=%" PRId64,
                        record.node.to_string(),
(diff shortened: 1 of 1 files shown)


src/dist/failure_detector/failure_detector.cpp
```diff
@@ -283,7 +283,8 @@ void failure_detector::check_all_records()
             worker_record &record = itq->second;
 
             if (record.is_alive != false &&
-                now - record.last_beacon_recv_time > _grace_milliseconds) {
+                now - record.last_beacon_recv_time > _grace_milliseconds &&
+                is_time_greater_than(now, record.last_beacon_recv_time)) {
```
qinzuoyan: is_time_greater_than(now, record.last_beacon_recv_time) 放在 now - record.last_beacon_recv_time > _grace_milliseconds 的前面比较好吧？

另外if前面要加注释，说明为什么要加is_time_greater_than检查。这种tricky的情况一定要通过注释记录下来，不然过段时间就忘记了。

hycdong: 嗯，好的，我改一下

neverchanje:
Can you reproduce this bug in unit-test? I still have no idea why

src/dist/failure_detector/failure_detector.cpp
```diff
@@ -282,7 +282,10 @@ void failure_detector::check_all_records()
         for (; itq != _workers.end(); itq++) {
             worker_record &record = itq->second;
 
+            // we should ensure now is greater than record.last_beacon_recv_time to aviod integer
+            // overflow
             if (record.is_alive != false &&
```
neverchanje: 这里的 `record.is_alive != false` 顺手改成 `record.is_alive` 吧。

hycdong:
> Can you reproduce this bug in unit-test? I still have no idea why `now` is larger than `last_beacon_recv_time` because of the lock. Well, it may be possible, but the reason why is the question.

Consider the situation below:
1. meta received replica beacon, get lock (in fd thread1)
2. meta is checking lease, get current time = now, and try to get lock (in fd thread2)
3. update last_recv_time, release lock (in fd thread1)
4. meta server get lock, continue to check lease, and it cost a little time for thread to get the lock, last_recv_time is greater than now (in fd thread2)

This condition is hard to reproduce in unit-test, because it is caused by concurrence and lock occasionally.

neverchanje:
Reasonable! LGTM.

PR #273, opened 2019-07-18
duplication: bugfix on pause_dup

neverchanje:
What this PR does:

- Fix the bug where:

Change a PAUSED duplication to PAUSED, since the status is not altered, after
meta server successfully persist the change to zookeeper, `dup->persist_status` will
unexpectedly fail because of the assertion:

```cpp
void duplication_info::persist_status()
{
    zauto_write_lock l(_lock);
   // this line fails
    dassert_dup(_is_altering,
                this,
                "impossible, callers never write a duplication that is not altering to meta store");
```

Solution:

If the status is not altered, meta server return ERR_OK instead of writing this change to zookeeper.

```cpp
+    if (!dup->is_altering()) {
+        return;
+    }
```

- The removed duplication is still visible to users, for ease of debugging.

```cpp
-        // the invalid duplication is not visible to user.
-        if (is_valid()) {
-            entry_list.emplace_back(to_duplication_entry());
-        }
+        entry_list.emplace_back(to_duplication_entry());
```

### DIFF
--- a/src/dist/replication/meta_server/duplication/duplication_info.h
+++ b/src/dist/replication/meta_server/duplication/duplication_info.h
@@ -88,10 +88,7 @@ public:
     void persist_status();
 
     // if this duplication is in valid status.
-    bool is_valid() const
-    {
-        return _status == duplication_status::DS_START || _status == duplication_status::DS_PAUSE;
-    }
+    bool is_valid() const { return is_duplication_status_valid(_status); }
 
     ///
     /// alter_progress -> persist_progress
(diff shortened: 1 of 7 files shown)


src/dist/replication/meta_server/duplication/duplication_info.h
```diff
@@ -88,10 +88,7 @@ class duplication_info
     void persist_status();
 
     // if this duplication is in valid status.
-    bool is_valid() const
-    {
-        return _status == duplication_status::DS_START || _status == duplication_status::DS_PAUSE;
-    }
+    bool is_valid() const { return is_duplication_status_valid(_status); }
```
hycdong: The function of 'is_duplication_status_valid'  is just as same as original 'is_valid', why would you write a new function instead of using original 'is_valid'?

neverchanje: for a shorter name, and no argument needed.

hycdong: ok

PR #289, opened 2019-08-02
utils: add test utils on meta server

neverchanje:
Move the common test-utils of meta-server (setup remote storage, delete all metadata after test, e.g.) into one single class `meta_test_base`.


vagetablechicken:
`error: ‘void dsn::replication::server_state::initialize_node_state()’ is private`

PR #304, opened 2019-08-29
feat(dup): add interface mutation_duplicator & duplication procedure

neverchanje:
This PR added interface `mutation_duplicator` and pipeline of duplication procedure.

## mutation_duplicator

This is the interface of data "sender" to be extensible for multiple endpoint types.
For example, if the remote endpoint is a pegasus cluster, the `pegasus_mutation_duplicator` will
transfer client RPCs to the remote pegasus.
If there's a further requirement to duplicate data to HBase/MySQL/etc., we can implement this interface.

## Duplication Pipeline

The other part of this PR is simple. In each round of duplication, a batch of mutations is loaded
from private-log, and is sent via `mutation_duplicator` to remote endpoint.

So the pipeline can be declared like the following (in bash convention):

```bash
while true
do
  load_from_private_log | ship_mutations
done
```

Each procedure in our implementation is declared as a **class**:

-  `load_from_private_log`, run in thread pool, `REPLICATION_LONG`
- `ship_mutations`, run in thread pool, `REPLICATION`

In our real implementation, we make this design slightly complicated:

```bash
while true
do
  load_mutations | ship_mutations
done

# load_mutations will fork a new task to load from private-log, aka `load_from_private_log`
```

- `load_mutations` runs in thread pool  `REPLICATION`, because we need to obtain the latest information of this replica. Leave to future refactoring if we find this unnecessary and we can delete this class.

## replica_duplicator

`replica_duplicator` is the basement of this pipeline, so it extends `pipeline::base`.

```cpp
thread_pool(LPC_REPLICATION_LOW).task_tracker(tracker()).thread_hash(get_gpid().thread_hash());
```
It means that the pipeline is running in the environment of:

- `LPC_REPLICATION_LOW`: in low priority of thread pool `REPLICATION`.
- using the task tracker of replica, which means when the replica destructs, the pipeline will be detroyed.
- using the thread hash of this replica.

### DIFF
--- a/src/dist/replication/lib/duplication/replica_duplicator.cpp
+++ b/src/dist/replication/lib/duplication/replica_duplicator.cpp
@@ -3,6 +3,8 @@
 // can be found in the LICENSE file in the root directory of this source tree.
 
 #include "replica_duplicator.h"
+#include "load_from_private_log.h"
+#include "duplication_pipeline.h"
 #include "dist/replication/lib/replica_stub.h"
 
 #include <dsn/dist/replication/replication_app_base.h>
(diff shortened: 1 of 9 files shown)


src/dist/replication/lib/duplication/replica_duplicator.cpp
```diff
@@ -30,12 +32,45 @@ replica_duplicator::replica_duplicator(const duplication_entry &ent, replica *r)
     }
     ddebug_replica(
         "initialize replica_duplicator [dupid:{}, meta_confirmed_decree:{}]", id(), it->second);
+    thread_pool(LPC_REPLICATION_LOW).task_tracker(tracker()).thread_hash(get_gpid().thread_hash());
```
hycdong: 这句是在做什么呢？

neverchanje: 补充了PR描述，你可以参照一下。

PR #307, opened 2019-09-04
build: use official thrift, cmake set default 3rdlibs, curl without idn

vagetablechicken:
https://github.com/XiaoMi/pegasus/issues/361#issuecomment-515656652

It's unnecessary to build curl with [libidn](https://www.gnu.org/software/libidn/manual/libidn.html#Introduction).


vagetablechicken:
building thrift-bin needs bison & flex
`apt -y install bison flex`

PR #317, opened 2019-09-19
feat(dup): implement procedure load_from_private_log

neverchanje:
`load_from_private_log` performs the heaviest procedure among the duplication:
**loading mutation-logs from disk**.

The entire process:

1. `load_from_private_log::run`:
ensure the logs from `start_decree` are available on disk.

2. `load_from_private_log::find_log_file_to_start`:
select the log file to load.

3. `load_from_private_log::replay_log_block`:
loading in a block by block way. once EOF is hit, switch to the next file.

4. Once a log block is loaded, the mutations will be passed down to the next stage, aka `ship_mutations`.

The unit test of this procedure is instrumented by some testing methods:
 
1. **write and load**: write a number of fake mutations to disk and load. The next stage after `load_from_private_log` is mocked as verification of the loaded logs.

2. **handle real-world logs, not only fake logs**: given files that were generated from a real pegasus server.

3. **fault injection**: the replaying of logs and file opening are injected with faults in 25% probability. This helps us to test the corner cases.


src/dist/replication/lib/duplication/load_from_private_log.h
```diff
@@ -31,11 +32,42 @@ class load_from_private_log : public replica_base,
     // The loaded mutations will be passed down to `ship_mutation`.
     void run() override;
 
+    void set_start_decree(decree start_decree);
+
+    /// ==== Implementation ==== ///
+
+    void find_log_file_to_start();
+
+    /// Find the log file that contains decree `d`.
```
hycdong: 这个注释中提到的decree是什么呢？

src/dist/replication/lib/duplication/load_from_private_log.cpp
```diff
@@ -2,22 +2,172 @@
...
+{
+    if (dsn_unlikely(log_file_map.empty())) {
+        derror_replica("unable to start duplication since no log file is available");
+        return;
+    }
+
+    for (auto it = log_file_map.begin(); it != log_file_map.end(); it++) {
+        auto next_it = std::next(it);
+        if (next_it == log_file_map.end()) {
+            // use the last file if no file to read
```
hycdong: 这里特殊处理了log_file只有一个文件的情况，不考虑_start_decree和这个log file的关系了嘛？以及什么时候调用set_start_decree给_start_decree赋值的呢？

neverchanje: 在 load 开始的时候给 set_start_decree 赋值，这个在后面的 PR 会很清楚的看到。

src/dist/replication/lib/duplication/load_from_private_log.h
neverchanje: fixed

src/dist/replication/lib/duplication/load_from_private_log.cpp
neverchanje: `find_log_file_to_start` 的目标是找到一个可以读的文件。意思是 “如果前 n-1 个文件都确定不可读，那就读第 n 个”，因为最后一个文件我们不知道它的 end_decree 是多少，所以就啥也不管，直接读。

PR #350, opened 2019-12-06
fix: fix the bug that threads don't stop when shared_io_service is released

levy5307:
In shared_io_service, there are some threads running. But shared_io_service doesn't have a destructor, so these threads can't be stopped.
So this pull request add a destructor for it to stop these threads.

### DIFF
--- a/src/core/tools/common/shared_io_service.h
+++ b/src/core/tools/common/shared_io_service.h
@@ -69,8 +69,8 @@ public:
     ~shared_io_service()
     {
         ios.stop();
-        for (int i = 0; i < _io_service_worker_count; i++) {
-            _workers[i]->join();
+        for (auto worker : _workers) {
+            worker->join();
         }
     }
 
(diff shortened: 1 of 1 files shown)


vagetablechicken:
这个和timer的联系是什么？

levy5307:
> 这个和timer的联系是什么？

timer用到了shared_io_service做io，然后创建了一个shared_io_service：
```
_timer.reset(new boost::asio::deadline_timer(tools::shared_io_service::instance().ios));
```
所以即使调用了_timer->cancel，shared_io_service里的thread也没有停止

src/core/tools/common/shared_io_service.h
```diff
@@ -66,6 +66,14 @@ class shared_io_service : public utils::singleton<shared_io_service>
         }
     }
 
+    ~shared_io_service()
+    {
+        ios.stop();
+        for (int i = 0; i < _io_service_worker_count; i++) {
```
acelyc111: 使用range based loop吧

PR #354, opened 2019-12-10
fix: fix meta response state to client 

Shuo-Jia:
#332 resolve the core dump when assert the meta state, but the response state to client is not reasonable. This pr fix the bug.


src/dist/replication/test/meta_test/unit_test/state_sync_test.cpp
```diff
@@ -288,6 +288,30 @@ void meta_service_test_app::state_sync_test()
         req.app_name = "make_no_sense";
         ss2->query_configuration_by_index(req, resp);
         ASSERT_EQ(dsn::ERR_OBJECT_NOT_FOUND, resp.err);
+
+        // 3.3 app is dropping/creating/recalling
+        std::shared_ptr<app_state> app = ss2->get_app(15);
```
levy5307: 我感觉这里用了好多15数字常量不太好，能不能统一到一个变量里？要不然把15改的话需要改一大片

neverchanje: 不相干的重构可以后续统一改

PR #368, opened 2019-12-24
refactor: move primary's learning preparation of cache into another function

neverchanje:


src/replica/replica_learn.cpp
```diff
@@ -1004,6 +956,69 @@ void replica::on_learn_reply(error_code err, learn_request &&req, learn_response
...
+        int count = 0;
+        for (decree d = learn_start_decree; d < response.prepare_start_decree; d++) {
+            auto mu = _prepare_list->get_mutation_by_decree(d);
+            dassert(mu != nullptr, "mutation must not be nullptr, decree = %" PRId64 "", d);
+            mu->write_to(writer, nullptr);
+            count++;
+        }
+        response.type = learn_type::LT_CACHE;
+        response.state.meta = writer.get_buffer();
+        ddebug("%s: on_learn[%016" PRIx64 "]: learner = %s, learn mutation cache succeed, "
```
levy5307, line 1007: ddebug_f

neverchanje: I don't refactor the original code at this time. It was automatically modified by clang-format, not me. I'd like to modify learn-related code as little as possible.

src/replica/replica_learn.cpp
```diff
@@ -437,62 +437,15 @@ void replica::on_learn(dsn::message_ex *msg, const learn_request &request)
...
     // in this case, the state on the PS is still incomplete
-    else {
+    bool should_learn_cache = prepare_cached_learn_state(request,
+                                                         learn_start_decree,
+                                                         local_committed_decree,
+                                                         learner_state,
+                                                         response,
+                                                         delayed_replay_prepare_list);
+    if (!should_learn_cache) {
         if (learn_start_decree > _app->last_durable_decree()) {
```
hycdong, line 449: It seems that the comments are for condition `if (learn_start_decree > _app->last_durable_decree())` in line 449, it is better to move comments between line 448 and line 449.
```
// learn delta state or checkpoint
// in this case, the state on the PS is still incomplete
```

neverchanje: I think "delta state" stands for log and cache. So it's still reasonable to see this comment stands above L442. I don't know what's the initial intention of this comment.

hycdong: Sounds reasonable~ I recommend removing comment `// in this case, the state on the PS is still incomplete` which is useless and confusing.

hycdong (approved):

Shuo-Jia (approved):

PR #387, opened 2020-02-04
fix(thrift): add start_time_ns for structure mutation_update in thrift file

hycdong:
Pull request #336 added a new filed called `start_time_ns` for structure `mutation_update`. `mutation_update` is defined in `replication.thrift`, and we generate `.h` and `.cpp` files atomically by thrift. However, original pr directly added this filed in `.h` and `.cpp` file, which might be difficult to maintain. 

This pr add a new filed `start_time_ns` for structure `mutation_update` in thrift file, and initialize it when creating this structure in function `add_client_request`.

### DIFF
--- a/src/dist/replication/replication.thrift
+++ b/src/dist/replication/replication.thrift
@@ -20,7 +20,7 @@ struct mutation_update
     //the serialization type of data, this need to store in log and replicate to secondaries by primary
     2:i32            serialization_type;
     3:dsn.blob       data;
-    4:i64	      start_time_ns;
+    4:i64            start_time_ns;
 }
 
 struct mutation_data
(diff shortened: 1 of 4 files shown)


src/dist/replication/replication.thrift
```diff
@@ -20,6 +20,7 @@ struct mutation_update
     //the serialization type of data, this need to store in log and replicate to secondaries by primary
     2:i32            serialization_type;
     3:dsn.blob       data;
+    4:i64	      start_time_ns;
```
acelyc111: tab替换为空格并对齐

neverchanje: 这个建议是 optional 的

neverchanje: 然后set的时候使用 __set_start_time_ns

PR #395, opened 2020-02-13
fix(utils): fix bug in time_utils

levy5307:
There is a function in time_utils, which name is ```get_localtime```. It calls ```localtime_r(const time_t *timep, struct tm *result)``` and return the return value of this function.
While ```localtime_r``` needs the second parameter, the struct tm *result, as a buffer to store the return value. But in our ```get_localtime``` function, it pass a local variable to ```localtime_r```. And after it returns, the local variable is destroyed. So this leads to some undefined errors.

### DIFF
--- a/include/dsn/utility/time_utils.h
+++ b/include/dsn/utility/time_utils.h
@@ -32,17 +32,17 @@
 namespace dsn {
 namespace utils {
 
-static struct tm *get_localtime(uint64_t ts_ms)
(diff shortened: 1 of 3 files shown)


levy5307:
本来为```time_ms_to_string/time_ms_to_date/time_ms_to_date_time/time_ms_to_date_time```这四个function添加了单测的，后来发现由于github上ci和本地时区不一样，导致转换出来的时间不一样，ci就报错了

PR #399, opened 2020-02-14
feat(split): re-implement function split_replica_exec

hycdong:
This pull request is re-implementing function `split_replica_exec`. During partition split, parent/child may want child/parent to execute another function, `split_replica_exec` is used for such function. 

To simplify description, we assume that parent want child to execute function A. Parent replica calls function `split_replica_exec` directly,  if child replica can be found on this replica_stub, child will execute function A, otherwise, parent will execute another function, usually is an error handler.

In current design, replica should execute functions in a same thread, whose id is calculated by its pid. However, `split_replica_exec` will let child replica execute function A in parent's thread, whose id is usually different from child's.

As a result,  I re-implement `split_replica_exec`. Parent replica still calls function `split_replica_exec` directly,  if child replica can be found on this replica_stub, replica_stub use `tasking::enqueue` to guarantee child replica execute function A in its own thread, and return ERR_OK, otherwise, return ERR_OBJECT_NOT_FOUND. Parent will execute the error handler if it got `ERR_OBJECT_NOT_FOUND` from `split_replica_exec`.

This pull request also updates unit tests and removes some useless cases.


src/dist/replication/lib/replica_stub.h
```diff
@@ -186,18 +186,14 @@ class replica_stub : public serverlet<replica_stub>, public ref_counter
...
-
-    // This function is used for partition split error handler, caller(replica)
-    // if partition split meet error, parent/child may want child/parent execute error handler
-    // if replica <pid> valid and exist, execute <handler>, otherwise return
+    // parent/child may want child/parent to execute function during partition split
+    // for parent, if replica existed, will execute function {handler} and return ERR_OK, else
+    // return ERR_OBJECT_NOT_FOUND
+    // for child, if pid.app_id > 0 and existed, will execute function {handler} and return ERR_OK,
+    // else return ERR_OBJECT_NOT_FOUND
+    dsn::error_code split_replica_exec(dsn::task_code code, gpid pid, local_execution handler);
```
neverchanje: This comment makes me more confused. I still don't know why it makes a difference whether the replica is child or parent.

```
// If replica `pid` exists, will execute function `handler` and return ERR_OK, 
// return ERR_OBJECT_NOT_FOUND otherwise.
```

PR #401, opened 2020-02-20
feat(dup): support HTTP API to inspect duplication state

neverchanje:
## What problem I solved in this PR?

API Example:

```
http://0.0.0.0:34602/meta/app/duplication?name=temp
```

```json
{
    "1": {
        "create_ts": "2020-03-04 15:24:13",
        "dupid": 1583306653,
        "not_confirmed_mutations_num": {
            "0": 1,
            "1": 0,
            "2": 1,
            "3": 0,
            "4": 1,
            "5": 1,
            "6": 0,
            "7": 0
        },
        "remote": "onebox2",
        "status": "DS_START"
    },
    "appid": 2
}
```

```json
{
    "1": {
        "create_ts": "2020-03-09 18:13:50",
        "dupid": 1583748830,
        "not_confirmed_mutations_num": {
            "0": 4964,
            "1": 5144,
            "2": 5123,
            "3": 5148,
            "4": 5208,
            "5": 5289,
            "6": 5253,
            "7": 5148
        },
        "remote": "onebox2",
        "status": "DS_PAUSE"
    },
    "appid": 2
}
```

"not_confirmed" is the number of mutations that are not confirmed yet. not_confirmed=0 means all written data has been duplicated and confirmed.

## How did I solve it?

Like usual, I added an interface at `meta_http_service`. But this time I felt strongly that it's not easy to use `table_printer` correctly for serialization from thrift to JSON.

So I introduced a new library called [nlohmann_json](https://github.com/nlohmann/json) in order to speed up development. Since our HTTP API returns only JSON text, using a JSON library directly instead of `table_printer` is more appropriate.

### About nlohmann_json

For now, we have two JSON libraries, "rapidjson", and "nlohmann_json".

- Why do I add a new JSON lib?

From my experiences, nlohmann_json is **much more user-friendly**. It doesn't need to declare a "buffer" or "allocator" before constructing a JSON. Of course, usability costs performance. In [a benchmark by miloyip](https://github.com/miloyip/nativejson-benchmark#parsing-time), rapidjson 8-times outperforms nlohmann_json in JSON construction. However, the performance advantage is trivial in almost every case of pegasus. We should care more about development speed.

As declared in README of nlohmann_json
> Speed. There are certainly faster JSON libraries out there. However, if your goal is to speed up your development by adding JSON support with a single header, then this library is the way to go. If you know how to use a std::vector or std::map, you are already set.

- Should we keep the libs both?

Yes, I suggest so. Perhaps we can remove rapidjson in future refactoring.

### DIFF
--- a/src/dist/http/http_server.cpp
+++ b/src/dist/http/http_server.cpp
@@ -127,23 +127,29 @@ void http_server::add_service(http_service *service)
         unresolved_query = decoded_unresolved_query.get_value();
     }
 
+    // remove tailing '\0'
+    if (!unresolved_path.empty() && *unresolved_path.crbegin() == '\0') {
+        unresolved_path.pop_back();
+    }
     std::vector<std::string> args;
     boost::split(args, unresolved_path, boost::is_any_of("/"));
-    std::vector<string_view> real_args;
-    for (string_view arg : args) {
-        if (!arg.empty() && strlen(arg.data()) != 0) {
-            real_args.emplace_back(string_view(arg.data()));
+    std::vector<std::string> real_args;
+    for (std::string &arg : args) {
+        if (!arg.empty()) {
+            real_args.emplace_back(std::move(arg));
         }
     }
-    if (real_args.size() > 2) {
-        return error_s::make(ERR_INVALID_PARAMETERS);
-    }
-    if (real_args.size() == 1) {
-        ret.service_method = {std::string(real_args[0]), ""};
-    } else if (real_args.size() == 0) {
+    if (real_args.size() == 0) {
         ret.service_method = {"", ""};
+    } else if (real_args.size() == 1) {
+        ret.service_method = {std::move(real_args[0]), ""};
     } else {
-        ret.service_method = {std::string(real_args[0]), std::string(real_args[1])};
+        std::string method = std::move(real_args[1]);
+        for (int i = 2; i < real_args.size(); i++) {
+            method += '/';
+            method += real_args[i];
+        }
+        ret.service_method = {std::move(real_args[0]), std::move(method)};
     }
 
     // find if there are method args (<ip>:<port>/<service>/<method>?<arg>=<val>&<arg>=<val>)
--- a/src/dist/http/test/http_server_test.cpp
+++ b/src/dist/http/test/http_server_test.cpp
@@ -27,7 +27,8 @@ TEST(http_server, parse_url)
         {"http://127.0.0.1:34601/threads/", ERR_OK, {"threads", ""}},
         {"http://127.0.0.1:34601//pprof/heap/", ERR_OK, {"pprof", "heap"}},
         {"http://127.0.0.1:34601//pprof///heap", ERR_OK, {"pprof", "heap"}},
-        {"http://127.0.0.1:34601/pprof/heap/arg/", ERR_INVALID_PARAMETERS, {}},
+        {"http://127.0.0.1:34601/pprof/heap/arg/", ERR_OK, {"pprof", "heap/arg"}},
+        {"http://127.0.0.1:34601/pprof///heap///arg/", ERR_OK, {"pprof", "heap/arg"}},
     };
 
     for (auto tt : tests) {
--- a/src/dist/replication/meta_server/duplication/duplication_info.cpp
+++ b/src/dist/replication/meta_server/duplication/duplication_info.cpp
@@ -191,5 +191,25 @@ duplication_info_s_ptr duplication_info::decode_from_blob(dupid_t dup_id,
     return dup;
 }
 
+void duplication_info::append_if_valid_for_query(
+    const app_state &app,
+    /*out*/ std::vector<duplication_entry> &entry_list) const
+{
+    zauto_read_lock l(_lock);
+
+    entry_list.emplace_back(to_duplication_entry());
+    duplication_entry &ent = entry_list.back();
+    ent.__isset.pending = true;
+    // the confirmed decree is not useful for displaying
+    // the overall state of duplication, instead we show pending mutations.
+    ent.__isset.progress = false;
+    for (const partition_configuration &part : app.partitions) {
+        int pid = part.pid.get_partition_index();
+        auto it = _progress.find(pid);
+        int64_t pending = part.last_committed_decree - it->second.stored_decree;
+        ent.pending[pid] = std::max(pending, int64_t(0));
+    }
+}
+
 } // namespace replication
 } // namespace dsn
--- a/src/dist/replication/meta_server/duplication/duplication_info.cpp
+++ b/src/dist/replication/meta_server/duplication/duplication_info.cpp
@@ -199,7 +199,7 @@ void duplication_info::append_if_valid_for_query(
 
     entry_list.emplace_back(to_duplication_entry());
     duplication_entry &ent = entry_list.back();
-    ent.__isset.pending = true;
+    ent.__isset.not_confirmed = true;
     // the confirmed decree is not useful for displaying
     // the overall state of duplication, instead we show pending mutations.
     ent.__isset.progress = false;
@@ -207,7 +207,7 @@ void duplication_info::append_if_valid_for_query(
         int pid = part.pid.get_partition_index();
         auto it = _progress.find(pid);
         int64_t pending = part.last_committed_decree - it->second.stored_decree;
-        ent.pending[pid] = std::max(pending, int64_t(0));
+        ent.not_confirmed[pid] = std::max(pending, int64_t(0));
     }
 }
 
--- a/src/dist/replication/meta_server/meta_http_service.cpp
+++ b/src/dist/replication/meta_server/meta_http_service.cpp
@@ -8,6 +8,7 @@
 #include <dsn/cpp/json_helper.h>
 #include <dsn/cpp/serialization_helper/dsn.layer2_types.h>
 #include <dsn/dist/replication/replication_types.h>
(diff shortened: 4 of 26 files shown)


src/dist/replication/test/meta_test/unit_test/meta_duplication_service_test.cpp
```diff
@@ -650,5 +651,34 @@ TEST_F(meta_duplication_service_test, recover_from_corrupted_meta_data)
...
+    auto dup = duplications.begin()->second;
+
+    fake_req.query_args["name"] = test_app;
+    mhs.query_duplication_handler(fake_req, fake_resp);
+    ASSERT_EQ(fake_resp.status_code, http_status_code::ok);
+    ASSERT_EQ(
+        fake_resp.body,
+        std::string("{\"") + std::to_string(dup->id) + "\":{\"create_ts\":" +
+            std::to_string(dup->create_timestamp_ms) +
+            ",\"progress\":{\"0\":-1,\"1\":-1,\"2\":-1,\"3\":-1,"
```
acelyc111: Using https://en.cppreference.com/w/cpp/language/string_literal is more readable

src/dist/http/test/http_server_test.cpp
```diff
@@ -27,7 +27,8 @@ TEST(http_server, parse_url)
         {"http://127.0.0.1:34601/threads/", ERR_OK, {"threads", ""}},
         {"http://127.0.0.1:34601//pprof/heap/", ERR_OK, {"pprof", "heap"}},
         {"http://127.0.0.1:34601//pprof///heap", ERR_OK, {"pprof", "heap"}},
-        {"http://127.0.0.1:34601/pprof/heap/arg/", ERR_INVALID_PARAMETERS, {}},
+        {"http://127.0.0.1:34601/pprof/heap/arg/", ERR_OK, {"pprof", "heap/arg"}},
+        {"http://127.0.0.1:34601/pprof///heap///arg/", ERR_OK, {"pprof", "heap/arg"}},
```
acelyc111: Why there are so many `/`? What's the different with a single `/`?

src/dist/http/http_server.cpp
```diff
@@ -127,23 +127,29 @@ void http_server::add_service(http_service *service)
...
+            real_args.emplace_back(std::move(arg));
         }
     }
-    if (real_args.size() > 2) {
-        return error_s::make(ERR_INVALID_PARAMETERS);
-    }
     if (real_args.size() == 1) {
-        ret.service_method = {std::string(real_args[0]), ""};
+        ret.service_method = {std::move(real_args[0]), ""};
     } else if (real_args.size() == 0) {
```
acelyc111: I think a more natural order would like size == 0, size == 1, other

src/dist/replication/meta_server/meta_http_service.cpp
```diff
@@ -573,8 +575,54 @@ void meta_http_service::query_backup_policy_handler(const http_request &req, htt
...
+        resp.body = "name should not be empty";
+        resp.status_code = http_status_code::bad_request;
+        return;
+    }
+    rpc_req.app_name = it->second;
+    duplication_query_response rpc_resp;
+    _service->_dup_svc->query_duplication_info(rpc_req, rpc_resp);
+    if (rpc_resp.err != ERR_OK) {
+        resp.body = rpc_resp.err.to_string();
+        if (rpc_resp.err == ERR_APP_NOT_EXIST) {
```
acelyc111: What's the return code if a app is exist but not in duplication?

src/dist/http/http_server.cpp
```diff
@@ -127,23 +127,29 @@ void http_server::add_service(http_service *service)
         unresolved_query = decoded_unresolved_query.get_value();
     }
 
+    // remove tailing '\0'
+    if (!unresolved_path.empty() && *unresolved_path.crbegin() == '\0') {
+        unresolved_path.pop_back();
+    }
```
Shuo-Jia: why remove? is it bug?

src/dist/http/test/http_server_test.cpp
Shuo-Jia: I test `multiple slashes` in browser, It is same to `single slash`.

src/dist/http/http_server.cpp
neverchanje: because we don't want the url path includes '\0'

acelyc111:
```
        "progress": {
            "0": -1,
            "1": -1,
            "2": -1,
            "3": -1,
            "4": -1,
            "5": -1,
            "6": -1,
            "7": -1
        },
```
What does `-1` here means?

neverchanje:
> ```
>         "progress": {
>             "0": -1,
>             "1": -1,
>             "2": -1,
>             "3": -1,
>             "4": -1,
>             "5": -1,
>             "6": -1,
>             "7": -1
>         },
> ```
> 
> What does `-1` here mean?

means invalid_decree, it will immediately turn into a valid decree after replica start syncing duplication progress with meta.

acelyc111:
> > ```
> >         "progress": {
> >             "0": -1,
> >             "1": -1,
> >             "2": -1,
> >             "3": -1,
> >             "4": -1,
> >             "5": -1,
> >             "6": -1,
> >             "7": -1
> >         },
> > ```
> > 
> > 
> > What does `-1` here mean?
> 
> means invalid_decree, it will immediately turn into a valid decree after replica start syncing duplication progress with meta.

For easily reading, could you modify the sample to change some `-1` to other values?
Another question, even thoungh you show me the current **duplication** decree, how can I know the current decree of partitions **on master** cluster? That is to say, I still don't know the progress.

neverchanje:
> For easily reading, could you modify the sample to change some `-1` to other values?
> Another question, even thoungh you show me the current **duplication** decree, how can I know the current decree of partitions **on master** cluster? That is to say, I still don't know the progress.

I modified the HTTP result to show "pending mutations" instead of  "confirmed_decree".

src/dist/replication/meta_server/duplication/duplication_info.cpp
```diff
@@ -191,5 +191,25 @@ duplication_info_s_ptr duplication_info::decode_from_blob(dupid_t dup_id,
...
+    duplication_entry &ent = entry_list.back();
+    ent.__isset.pending = true;
+    // the confirmed decree is not useful for displaying
+    // the overall state of duplication, instead we show pending mutations.
+    ent.__isset.progress = false;
+    for (const partition_configuration &part : app.partitions) {
+        int pid = part.pid.get_partition_index();
+        auto it = _progress.find(pid);
+        int64_t pending = part.last_committed_decree - it->second.stored_decree;
+        ent.pending[pid] = std::max(pending, int64_t(0));
```
levy5307: Why the variable of ent.progress is not set here?

src/dist/replication/replication.thrift
```diff
@@ -678,7 +678,10 @@ struct duplication_entry
     4:i64                  create_ts;
 
     // partition_index => confirmed decree
-    5:map<i32, i64>        progress;
+    5:optional map<i32, i64> progress;
```
acelyc111: How about rename to `confirmed_decree`, and `pending ` rename to `pending_mutations ` , it would be more clearly.

src/dist/replication/meta_server/duplication/duplication_info.cpp
neverchanje: Because progress (confirmed_decree) does not give useful information in terms of querying duplication state. We need "pending" to tell whether the duplication is healthy or blocked due to duplication failure, eg. intra-cluster network failure.

src/dist/replication/replication.thrift
neverchanje: I think the meaning of 'pending' is too long to use variable name to describe.
If we want to name it completely, its name may be "non_confirmed_mutations_count".
So I think it's fine to use a shorter name but with some documents describing it in detail.

neverchanje: Or I can rename `pending` to `not_confirmed`.

acelyc111: It's OK to use `not_confirmed ` as a internal variable name, but not suitable for user faced name, such as http interface, or metric name, I might consider it's not confirmed K-V pairs, decrees, etc. I suggest you to make it more clearly for http interface, at least.

neverchanje: ok, so which do you think the suitable, 'not_confirmed_mutations' or 'not_confirmed_mutations_num'?

src/dist/replication/meta_server/duplication/duplication_info.cpp
```diff
@@ -191,5 +191,25 @@ duplication_info_s_ptr duplication_info::decode_from_blob(dupid_t dup_id,
...
+
+    entry_list.emplace_back(to_duplication_entry());
+    duplication_entry &ent = entry_list.back();
+    ent.__isset.not_confirmed = true;
+    // the confirmed decree is not useful for displaying
+    // the overall state of duplication, instead we show pending mutations.
+    ent.__isset.progress = false;
+    for (const partition_configuration &part : app.partitions) {
+        int pid = part.pid.get_partition_index();
+        auto it = _progress.find(pid);
```
levy5307: Does each pid has a corresponding entry in _progress?

neverchanje: Yes, `progress` is used to sync duplication confirmed decree between meta and replica. Each partition within the duplicating app reports its confirmed decree to meta, so that during failure recovery, the partition can restore the former progress and restart duplication.

PR #403, opened 2020-02-24
refactor(utils): make logger decoupled from dsn runtime

levy5307:
The logger API, is tightly coupled in rdsn runtime for historical issues. As a part of the long-term plan of rdsn (see #141 ) we decoupled it to make further refactoring possible.


levy5307:
logger_proxy没有使用继承singleton来实现单例（懒汉式），因为这种通过测试发现，这种方式有问题：https://github.com/XiaoMi/rdsn/issues/404

但是目前这种实现方案（饿汉式）也有问题，即：静态变量的初始化问题。例如在pegasus的function test代码中，global_env也是个单例，由于global_env的创建时间比logger_proxy要早，所以在调用global_env的构造函数时，logger_proxy的单例还没有初始化。而global_env的构造函数里却调用了日志打印，这样会导致coredump。静态变量初始化问题的解决方案就是使用懒汉式，在使用时通过instance()函数去初始化该静态变量，这样可以保证在用的时候一定初始化了。
目前采用的方法是获取logger_proxy的单例时先判断一下其是否为nullptr，避免发生coredump

同理，clock的重构也是采用的这种饿汉式的方式，也有同样的问题

src/core/tools/common/providers.common.cpp
```diff
@@ -39,19 +39,20 @@
 #include "native_aio_provider.linux.h"
 #include "simple_task_queue.h"
 #include "network.sim.h"
-#include "simple_logger.h"
 #include "dsn_message_parser.h"
 #include "thrift_message_parser.h"
 #include "raw_message_parser.h"
+#include "../../core/simple_logger.h"
```
neverchanje: don't use relative-path.
```suggestion
#include "core/simple_logger.h"
```

src/core/core/simple_logger.cpp
```diff
@@ -0,0 +1,182 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+#include <sstream>
+#include <dsn/utility/filesystem.h>
+#include <dsn/utility/time_utils.h>
+#include <dsn/tool_api.h>
+
+#include "simple_logger.h"
```
neverchanje: src/core/tools/common/simple_logger.h → src/core/core/simple_logger.h 
Why move this file? simple_logger is a tool, it's one specific implmentation of the provider interface.
So it's suitable being under /core/tools.

src/core/core/logger.cpp
```diff
@@ -0,0 +1,137 @@
...
+/*
+ * Description:
+ * Description:
+ *     What is this file about?
+ *
+ * Revision history:
+ *     xxxx-xx-xx, author, first version
+ *     xxxx-xx-xx, author, fix bug about xxx
+ */
+
```
acelyc111: 删除

src/core/tools/common/providers.common.cpp
acelyc111: 可以不用这种形式的相对路径吗

src/core/tests/logger.cpp
```diff
@@ -33,12 +33,15 @@
  *     xxxx-xx-xx, author, fix bug about xxx
  */
```
acelyc111: 改到的文件就把这种描述section删除掉吧

src/core/core/screen_logger.h
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+#pragma once
+
+#include <dsn/utility/logger.h>
+
+namespace dsn {
+namespace utils {
+
+class screen_logger : public logger
```
acelyc111: 加个描述吧，说明这个类的用处

src/core/core/simple_logger.h
```diff
@@ -35,75 +35,48 @@
...
-
-    virtual void flush();
-
-private:
-    ::dsn::utils::ex_lock_nr _lock;
-    bool _short_header;
-};
-
-class simple_logger : public logging_provider
+class simple_logger : public logger
```
acelyc111: 也加个类的描述

src/core/core/logger_proxy.h
```diff
@@ -0,0 +1,68 @@
...
+ */
+
+#pragma once
+
+#include <dsn/utility/logger.h>
+
+namespace dsn {
+namespace utils {
+
+class logger_proxy : public logger
```
acelyc111: 添加类描述

src/core/core/logger_proxy.h
```diff
@@ -0,0 +1,68 @@
...
+    void set_stderr_start_level(dsn_log_level_t stderr_start_level) override;
+
+private:
+    static std::unique_ptr<logger_proxy> _instance;
+    std::unique_ptr<logger> _logger;
+    dsn_log_level_t _log_start_level;
+};
+
+} // namespace utils
+} // namespace dsn
```
acelyc111: 末尾空行

acelyc111:
> logger_proxy没有使用继承singleton来实现单例（懒汉式），因为这种通过测试发现，这种方式有问题：#404
> 
> 但是当前的实现方案（饿汉式）也有问题，即：静态变量的初始化问题。例如在pegasus的function test代码中，global_env也是个单例，由于global_env的创建时间比logger_proxy要早，所以在调用global_env的构造函数时，logger_proxy的单例还没有初始化。而global_env的构造函数里却调用了日志打印，这样会导致coredump。静态变量初始化问题的解决方案就是使用懒汉式，在使用时通过instance()函数去初始化该静态变量，这样可以保证在用的时候一定初始化了。但是如上所述，懒汉式的方案也有问题。
> 目前采用的方法是获取logger_proxy的单例时先判断一下其是否为nullptr，避免发生coredump，如下所示：
> 
> ```
> DSN_API void dsn_logv(const char *file,
>                       const char *function,
>                       const int line,
>                       dsn_log_level_t log_level,
>                       const char *fmt,
>                       va_list args)
> {
>     ::dsn::utils::logger *log = dsn::utils::logger_proxy::instance();
>     if (nullptr != log) {
>         log->logv(file, function, line, log_level, fmt, args);
>     } else {
>         printf("%s:%d:%s():", file, line, function);
>         vprintf(fmt, args);
>         printf("\n");
>     }
> }
> ```
> 
> 但是其实这样代码是不太优雅的。不过重构前这段代码也是这么写的，可能当时也是遇到了相同的问题。
> 
> 同理，clock的重构也是采用的这种饿汉式的方式，也有同样的问题，即启动时如果有某个单例比其初始化早，并且调用clock::instance()，将会coredump

So the problem is the destruct order of singletons, is it able to

src/core/tools/common/simple_logger.cpp
```diff
@@ -86,14 +86,7 @@ static void print_header(FILE *fp, dsn_log_level_t log_level)
     }
 }
 
-screen_logger::screen_logger(const char *log_dir) : logging_provider(log_dir)
-{
-    _short_header =
```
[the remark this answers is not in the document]
levy5307: All of our config.ini set short_header = false, so I think it is useless

src/core/tools/common/logger_proxy.h
```diff
@@ -0,0 +1,73 @@
...
+ */
+
+#pragma once
+
+#include <dsn/tool-api/logging_provider.h>
+
+namespace dsn {
+namespace tools {
+
+class logger_proxy : public logging_provider
```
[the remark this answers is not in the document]
levy5307: I think let the logger depends on config reader is a best practise, because log reader needs to be initialized by dsn runtime.
The logger_proxy is a interface of all of child logger type. It is used to control and manage access to an object of logger child type, so these logger child type don't have to be exposed outside.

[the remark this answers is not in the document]
neverchanje: config reader belongs to dsn_utils, logger belongs to dsn_utils too.
So why do you think that logger depends on config reader is not reasonable?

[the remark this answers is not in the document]
levy5307: Here is my thought: 
If the config file is not load by dsn runtime, we can get nothing from it. In other words, if the rdsn runtime is not running, the config reader doesn`t work. So If we let logger depends on the config reader, it depends on the dsn runtime indirectly.

neverchanje:
> If the config file is not loaded by dsn_runtime, we can get nothing from it

dsn_runtime depends on config-lib, you misunderstood that config-lib depends on dsn_runtime. Every application can use `dsn_config_load` to read config, it's not limited to dsn_runtime only.

```cpp
    if (!dsn_config_load(config_file, config_arguments)) {
        printf("Fail to load config file %s\n", config_file);
        return false;
    }
```

So logger depends on config-lib is reasonable.
We can first decouple config-lib from dsn_runtime, then in PR afterward we can decouple logger with config-lib, if it's worthwhile.

src/core/core/logging.cpp
```diff
@@ -70,16 +60,18 @@ void dsn_log_init()
         ::dsn::tools::sys_exit.put_back(log_on_sys_exit, "log.flush");
     }
 
+    dsn::logging_provider *logger = dsn::utils::factory_store<dsn::logging_provider>::create(
+        logging_factory_name.c_str(), ::dsn::PROVIDER_TYPE_MAIN, dir_log.c_str());
```
acelyc111: `::dsn`, remove the leading colons

src/core/core/logging.cpp
```diff
@@ -148,10 +134,26 @@ DSN_API void dsn_log(const char *file,
...
 }
+
+namespace dsn {
+
+std::unique_ptr<logging_provider> logging_provider::_logger =
+    std::unique_ptr<logging_provider>(nullptr);
+
+logging_provider *logging_provider::instance()
+{
+    static std::unique_ptr<logging_provider> default_logger =
```
acelyc111: Why `default_logger` defined as static? its lifetime will be kept until process exit.

src/core/tools/common/simple_logger.h
```diff
@@ -26,11 +26,9 @@
 
 /*
  * Description:
```
acelyc111: `Description` section should be removed completly, and move the class description closer to the class definition.

src/core/core/logging.cpp
levy5307: because if not use static, a new logger will be created by each call to logging_provider::instance().
Yet its lifetime will be kept untiil process exit, but there is extra one static logger only, so I think it`s ok.

acelyc111: It's because that `logging_provider::instance()` may be called befaore `_logger ` has been created, we have to use `default_logger` instead that time?

neverchanje: Yes. `_logger` is the user-customized logger. For now, dsn_runtime is the "user", which reads the configuration and `set_logger` to replace default_logger (`dsn_log_init`). But if an application (like unit-test) without binding the runtime wants to print log, it doesn't have to set a logger of its own.

PR #409, opened 2020-02-27
feat: add query_disk_info support for replica_server

Shuo-Jia:
The https://github.com/XiaoMi/pegasus/issues/261 describes the design of the disk rebalancer, this pr adds the surpport for `query_disk_info`.


src/dist/replication/lib/replica_stub.cpp
```diff
@@ -889,6 +889,87 @@ void replica_stub::on_query_replica_info(const query_replica_info_request &req,
...
+        for (auto it = _replicas.begin(); it != _replicas.end(); ++it) {
+            replica_ptr &r = it->second;
+            const app_info &a_info = *r->get_app_info();
+            if (app_ids.find(a_info.app_id) == app_ids.end()) {
+                app_ids.insert(a_info.app_id);
+            }
+        }
+    }
+
+    for (auto &n : _fs_manager._dir_nodes) {
```
acelyc111: `n` is meanless, rename to `node`, and also const reference

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -889,6 +889,87 @@ void replica_stub::on_query_replica_info(const query_replica_info_request &req,
...
+        disk_info d_info;
+        for (auto app_id : app_ids) {
+            const auto &iter = n->holding_replicas.find(app_id);
+            if (iter == n->holding_replicas.end()) {
+                dwarn("disk(disk_dir = %s) hasn't been allocated space for app(app_id = %d)",
```
acelyc111: It's normal that a dir not contain an app, why logging as warning? Suppose a cluster with many 4-partion apps, there will be many WARN level log.

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -889,6 +889,87 @@ void replica_stub::on_query_replica_info(const query_replica_info_request &req,
...
+                      n->full_dir.c_str(),
+                      app_id);
+                continue;
+            } else {
```
acelyc111: The former case is `continue`, you can remove `else` and the following indent

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -889,6 +889,87 @@ void replica_stub::on_query_replica_info(const query_replica_info_request &req,
...
+                std::set<dsn::gpid> &replicas_for_app = iter->second;
```
acelyc111: Move it closer to its using place, L944

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -889,6 +889,87 @@ void replica_stub::on_query_replica_info(const query_replica_info_request &req,
...
+
+                const auto &primary_iter = n->holding_primary_replicas.find(app_id);
+                if (primary_iter == n->holding_primary_replicas.end()) {
+                    dwarn("this is the first time to calc primary count or the disk(disk_dir = %s) "
+                          "hasn't been "
+                          "allocated primary space for app(app_id = %d)",
+                          n->full_dir.c_str(),
+                          app_id);
+                } else {
+                    primary_iter->second.clear();
```
acelyc111: `on_query_disk_info ` is just a getter, why you intent to change any content?

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -889,6 +889,87 @@ void replica_stub::on_query_replica_info(const query_replica_info_request &req,
...
+    ddebug("got query disk info request from (%s)", req.node.to_string());
+    _fs_manager.compute_disk_density();
+
+    std::set<app_id> app_ids;
+    {
+        zauto_read_lock l(_replicas_lock);
+        for (auto it = _replicas.begin(); it != _replicas.end(); ++it) {
+            replica_ptr &r = it->second;
+            const app_info &a_info = *r->get_app_info();
+            if (app_ids.find(a_info.app_id) == app_ids.end()) {
```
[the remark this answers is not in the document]
Shuo-Jia: different `replica_ptr` may get same `app_id`, so `app_ids` may contain some `app_id` from `replica_ptr`.

src/dist/replication/lib/replica_stub.cpp
Shuo-Jia: I remove `continue` and move some code into `else`

src/dist/replication/lib/replica_stub.cpp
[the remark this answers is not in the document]
Shuo-Jia: New commit has modify the style of getting `app_id` to frome the `holding_replica`

src/dist/replication/lib/replica_stub.h
```diff
@@ -95,6 +95,8 @@ class replica_stub : public serverlet<replica_stub>, public ref_counter
                          /*out*/ query_replica_decree_response &resp);
     void on_query_replica_info(const query_replica_info_request &req,
                                /*out*/ query_replica_info_response &resp);
+    void on_query_disk_info(const query_disk_info_request &req,
```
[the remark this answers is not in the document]
Shuo-Jia: New commit add `register_rpc_handler(RPC_QUERY_DISK_INFO, "query_disk_info", &replica_stub::on_query_disk_info);` and 2th argument don't  match the "const member function"

src/dist/replication/common/fs_manager.h
```diff
@@ -91,18 +105,45 @@ class fs_manager
     void update_disk_stat();
 
 private:
+    void reset_disk_stat()
+    {
+        _capacity_total_mb = 0;
```
acelyc111: `_total_capacity_mb`, the same to the others

src/dist/replication/common/fs_manager.h
```diff
@@ -91,18 +105,45 @@ class fs_manager
...
     perf_counter_wrapper _counter_available_total_ratio;
     perf_counter_wrapper _counter_available_min_ratio;
     perf_counter_wrapper _counter_available_max_ratio;
+
+    friend class replica_stub;
+    friend class mock_replica_stub;
+    friend class replica_disk_test;
 };
 }
 }
```
acelyc111: namespace

src/dist/replication/common/fs_manager.h
```diff
@@ -91,18 +105,45 @@ class fs_manager
...
-    std::vector<std::unique_ptr<dir_node>> _dir_nodes;
+
+    int64_t _capacity_total_mb = 0;
+    int64_t _available_total_mb = 0;
+    int64_t _available_total_ratio = 0;
+    int64_t _available_min_ratio = 100;
+    int64_t _available_max_ratio = 0;
+
+    std::vector<std::shared_ptr<dir_node>> _dir_nodes;
+    std::set<std::shared_ptr<dir_node>> _sorted_dir_nodes;
```
acelyc111: Elements in `std::set<std::shared_ptr<dir_node>>` are not sorted by `dir_node` but sorted by `std::shared_ptr<>`

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -889,6 +889,41 @@ void replica_stub::on_query_replica_info(const query_replica_info_request &req,
     resp.err = ERR_OK;
 }
 
+void replica_stub::on_query_disk_info(const query_disk_info_request &req,
+                                      /*out*/ query_disk_info_response &resp)
+{
+    ddebug("got query disk info request from (%s)", req.node.to_string());
```
acelyc111: This log is useless, remove it.

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -889,6 +889,41 @@ void replica_stub::on_query_replica_info(const query_replica_info_request &req,
     resp.err = ERR_OK;
 }
 
+void replica_stub::on_query_disk_info(const query_disk_info_request &req,
```
acelyc111: Define it as const member functon to make sure you won't change anything.

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -889,6 +889,41 @@ void replica_stub::on_query_replica_info(const query_replica_info_request &req,
...
+                                      /*out*/ query_disk_info_response &resp)
+{
+    ddebug("got query disk info request from (%s)", req.node.to_string());
+
+    for (const auto &dir_node : _fs_manager._dir_nodes) {
+        disk_info disk_info;
+        for (auto primary_iter = dir_node->holding_primary_replicas.begin();
```
acelyc111: Use range loop instead.

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -889,6 +889,41 @@ void replica_stub::on_query_replica_info(const query_replica_info_request &req,
     resp.err = ERR_OK;
 }
 
+void replica_stub::on_query_disk_info(const query_disk_info_request &req,
+                                      /*out*/ query_disk_info_response &resp)
+{
+    ddebug("got query disk info request from (%s)", req.node.to_string());
+
+    for (const auto &dir_node : _fs_manager._dir_nodes) {
+        disk_info disk_info;
```
acelyc111: struct and object do not use the same name.

src/dist/replication/replication.thrift
```diff
@@ -436,6 +436,30 @@ struct query_replica_info_response
...
 }
 
+struct disk_info
+{
+    1:string tag;
+    2:string full_dir;
+    3:i64 disk_capacity_mb;
+    4:i64 disk_available_mb;
+    5:i64 disk_density;
+    6:map<i32,i32> holding_primary_replicas;
```
acelyc111: Comment what key and value mean.

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -889,6 +889,41 @@ void replica_stub::on_query_replica_info(const query_replica_info_request &req,
...
+        for (auto primary_iter = dir_node->holding_primary_replicas.begin();
+             primary_iter != dir_node->holding_primary_replicas.end();
+             primary_iter++) {
+            disk_info.holding_primary_replicas[primary_iter->first] =
```
acelyc111: `=` should be `+=`, right?

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -889,6 +889,41 @@ void replica_stub::on_query_replica_info(const query_replica_info_request &req,
...
+                static_cast<int>(dir_node->holding_primary_replicas[primary_iter->first].size());
+        }
+
+        for (auto secondary_iter = dir_node->holding_secondary_replicas.begin();
+             secondary_iter != dir_node->holding_secondary_replicas.end();
+             secondary_iter++) {
+            disk_info.holding_secondary_replicas[secondary_iter->first] = static_cast<int>(
+                dir_node->holding_secondary_replicas[secondary_iter->first].size());
+        }
```
acelyc111: Same.

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -889,6 +889,41 @@ void replica_stub::on_query_replica_info(const query_replica_info_request &req,
...
+
+        disk_info.tag = dir_node->tag;
+        disk_info.full_dir = dir_node->full_dir;
+        disk_info.disk_capacity_mb = dir_node->disk_capacity_mb;
+        disk_info.disk_available_mb = dir_node->disk_available_mb;
+        disk_info.disk_density = dir_node->disk_density;
+
+        resp.disk_infos.push_back(disk_info);
```
acelyc111: Use `emplace_back`

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -2512,5 +2548,28 @@ void replica_stub::split_replica_error_handler(gpid pid, local_execution handler
...
+void replica_stub::update_disk_holding_replicas()
+{
+    for (const auto &dir_node : _fs_manager._dir_nodes) {
+        // clear the holding_primary_replicas/holding_secondary_replicas and re-calculate it from
+        // holding_replica
+        dir_node->holding_primary_replicas.clear();
+        dir_node->holding_secondary_replicas.clear();
+        for (auto iter = dir_node->holding_replicas.begin();
+             iter != dir_node->holding_replicas.end();
+             iter++) {
```
acelyc111: range loop

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -2512,5 +2548,28 @@ void replica_stub::split_replica_error_handler(gpid pid, local_execution handler
...
+            const std::set<dsn::gpid> &replicas_for_app = iter->second;
+            for (const auto &pid : replicas_for_app) {
+                replica_ptr r = get_replica(pid);
+                if (r->status() == partition_status::PS_PRIMARY) {
+                    dir_node->holding_primary_replicas[iter->first].emplace(pid);
+                } else if (r->status() == partition_status::PS_SECONDARY) {
+                    dir_node->holding_secondary_replicas[iter->first].emplace(pid);
+                }
```
acelyc111:
```suggestion
                } else {
                    dassert(r->status() == partition_status::PS_SECONDARY), "");
                    dir_node->holding_secondary_replicas[iter->first].emplace(pid);
                }
```

src/dist/replication/common/fs_manager.h
acelyc111: And `_sorted_dir_nodes ` is only used for test?

Shuo-Jia: No, It's for findind the max and min dir_node(sorted by denisty ) for balance

Shuo-Jia: > Elements in `std::set<std::shared_ptr<dir_node>>` are not sorted by `dir_node` but sorted by `std::shared_ptr<>`

really? I should test it aganin and modify it. if that, should I only use std::set<dir_node>?

src/dist/replication/lib/replica_stub.cpp
Shuo-Jia: New commit add `register_rpc_handler(RPC_QUERY_DISK_INFO, "query_disk_info", &replica_stub::on_query_disk_info);` and 2th argument don't  match the "const member function"

src/dist/replication/lib/replica_stub.cpp
Shuo-Jia: No, `dir_node->holding_primary_replicas` is map<app_id, set<gpid>>,`disk_info.holding_primary_replicas` is `map<app_id, gpids_size>`

src/dist/replication/common/fs_manager.h
acelyc111: This is test only, you'd better remove it.

src/dist/replication/lib/replica_stub.cpp
Shuo-Jia: `r->status() ` also contains `PS_POTENTIAL_SECONDARY` ,`PS_INVALID` and so on, this only focuses on `PS_PRIMARY` and `PS_SECONDARY`, `dassert ` may result coredump.

src/dist/replication/common/fs_manager.h
Shuo-Jia: > No, It's for find the max and min dir_node(sorted by denisty ) for balance

Not only for test. And new commit have been modified to use:
```
static bool sorted_by_density(const std::shared_ptr<dir_node> &left,
                                  const std::shared_ptr<dir_node> &right)
    {
        return left->disk_density < right->disk_density;
    };
```
to compare, you can review it.

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -889,6 +889,35 @@ void replica_stub::on_query_replica_info(const query_replica_info_request &req,
     resp.err = ERR_OK;
 }
 
+void replica_stub::on_query_disk_info(const query_disk_info_request &req,
+                                      /*out*/ query_disk_info_response &resp)
+{
+    for (const auto &dir_node : _fs_manager._dir_nodes) {
+        disk_info info;
+        for (const auto &holding_primary_replica : dir_node->holding_primary_replicas) {
```
acelyc111: There may be several replicas, iteration variable should be plural.

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -889,6 +889,35 @@ void replica_stub::on_query_replica_info(const query_replica_info_request &req,
...
+            info.holding_primary_replicas[holding_primary_replica.first] = static_cast<int>(
+                dir_node->holding_primary_replicas[holding_primary_replica.first].size());
+        }
+
+        for (const auto &holding_secondary_replica : dir_node->holding_secondary_replicas) {
+            info.holding_secondary_replicas[holding_secondary_replica.first] = static_cast<int>(
+                dir_node->holding_secondary_replicas[holding_secondary_replica.first].size());
+        }
```
acelyc111: Same.

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -2512,5 +2542,26 @@ void replica_stub::split_replica_error_handler(gpid pid, local_execution handler
...
 }
 
+void replica_stub::update_disk_holding_replicas()
+{
+    for (const auto &dir_node : _fs_manager._dir_nodes) {
+        // clear the holding_primary_replicas/holding_secondary_replicas and re-calculate it from
+        // holding_replica
+        dir_node->holding_primary_replicas.clear();
+        dir_node->holding_secondary_replicas.clear();
+        for (const auto &holding_replica : dir_node->holding_replicas) {
```
acelyc111: `holding_replica ` -> `holding_replicas `

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -2512,5 +2542,26 @@ void replica_stub::split_replica_error_handler(gpid pid, local_execution handler
...
+            const std::set<dsn::gpid> &replicas_for_app = holding_replica.second;
```
acelyc111: `replicas_for_app ` -> `pids`

src/dist/replication/test/replica_test/unit_test/mock_utils.h
```diff
@@ -211,6 +211,30 @@ class mock_replica_stub : public replica_stub
         return rep;
     }
 
+    void generate_replicas_base_dir_nodes_for_app(app_info mock_app,
+                                                  int primary_count_for_disk = 1,
+                                                  int secondary_count_for_disk = 2)
+    {
+        std::vector<std::shared_ptr<dir_node>> &dir_nodes = _fs_manager._dir_nodes;
+        for (auto &n : dir_nodes) {
```
acelyc111: `n` -> `dir_node`

src/dist/replication/test/replica_test/unit_test/mock_utils.h
```diff
@@ -211,6 +211,30 @@ class mock_replica_stub : public replica_stub
...
+            for (auto holding_replica = n->holding_replicas.begin();
+                 holding_replica != n->holding_replicas.end();
+                 holding_replica++) {
```
acelyc111: range loop

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -889,6 +889,35 @@ void replica_stub::on_query_replica_info(const query_replica_info_request &req,
     resp.err = ERR_OK;
 }
 
+void replica_stub::on_query_disk_info(const query_disk_info_request &req,
+                                      /*out*/ query_disk_info_response &resp)
+{
+    for (const auto &dir_node : _fs_manager._dir_nodes) {
+        disk_info info;
+        for (const auto &holding_primary_replica : dir_node->holding_primary_replicas) {
+            info.holding_primary_replicas[holding_primary_replica.first] = static_cast<int>(
```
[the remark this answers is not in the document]
Shuo-Jia: This is only one disk replica info.

src/dist/replication/test/replica_test/unit_test/replica_disk_test.cpp
```diff
@@ -0,0 +1,121 @@
...
+public:
+    std::unique_ptr<mock_replica_stub> _stub;
+    int app_id = 1;
+    int dir_nodes_count = 2;
+    int primary_count_for_disk = 1;
+    int secondary_count_for_disk = 2;
+    dsn::app_info app_info;
+
+public:
+    void SetUp()
```
acelyc111: you'd better add `override`

src/dist/replication/test/replica_test/unit_test/replica_disk_test.cpp
```diff
@@ -0,0 +1,121 @@
...
+            app_info, primary_count_for_disk, secondary_count_for_disk);
+        _stub->on_disk_stat();
+    }
+
+    std::vector<std::shared_ptr<dir_node>> get_fs_manager_nodes()
+    {
+        return _stub->_fs_manager._dir_nodes;
+    }
+
+    void TearDown() {}
```
acelyc111: remove it

src/dist/replication/test/replica_test/unit_test/replica_disk_test.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+namespace dsn {
+namespace replication {
+
+class replica_disk_test : public testing::Test
+{
+public:
+    std::unique_ptr<mock_replica_stub> _stub;
+    int app_id = 1;
+    int dir_nodes_count = 2;
```
acelyc111: Could you set some large value to this variable for test? like 7, 12.

src/dist/replication/test/replica_test/unit_test/replica_disk_test.cpp
```diff
@@ -0,0 +1,121 @@
...
+                                               "full_dir_" + std::to_string(num),
+                                               disk_capacity_mb,
+                                               disk_available_mb,
+                                               disk_available_ratio);
+            int replica_index = 0;
+            int disk_holding_replica_count = primary_count_for_disk + secondary_count_for_disk;
+            while (disk_holding_replica_count-- > 0) {
+                node_disk->holding_replicas[app_id].emplace(gpid(app_id, replica_index++));
+            }
+            _stub->_fs_manager._dir_nodes.emplace_back(node_disk);
```
acelyc111: `node_disk`s are put into `_dir_nodes` regularly, i.e. in inverted order of `disk_available_mb`, could you shuffle the order?

Shuo-Jia: Only `_stub->on_disk_stat()` can trigger shuffle, In there, `_stub->on_disk_stat()`  is excuted in `SetUp()` in `replica_disk_test.cpp`

src/dist/replication/test/replica_test/unit_test/replica_disk_test.cpp
Shuo-Jia: Yes, test 10~

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -1712,6 +1740,7 @@ void replica_stub::on_disk_stat()
     _counter_replicas_garbage_replica_dir_count->set(garbage_replica_dir_count);
 
     _fs_manager.update_disk_stat();
+    update_disk_holding_replicas();
```
hycdong: `on_disk_stat` is called by timer, time interval is `options.disk_stat_interval_seconds`, whose default value is 600 seconds. Is this time interval suitable for updating disk primary and secondary stat?

src/dist/replication/common/fs_manager.h
```diff
@@ -90,19 +103,52 @@ class fs_manager
     bool for_each_dir_node(const std::function<bool(const dir_node &)> &func) const;
     void update_disk_stat();
 
+    // when executing "disk rebalance", we need find min and max node(sorted_by_density)
+    static bool sorted_by_density(const std::shared_ptr<dir_node> left,
+                                  const std::shared_ptr<dir_node> right)
```
hycdong: `disk_density` is not always greater than zero. For example, disk_ratio is 8%, 11%, total_disk_ratio is 10%, so the density will be -2 and 1. I suggest you to consider this condition, or guarantee `disk_density` greater than zero.

src/dist/replication/common/fs_manager.h
```diff
@@ -50,23 +50,36 @@ struct dir_node
...
+          disk_density(disk_density_)
     {
     }
     unsigned replicas_count(app_id id) const;
     unsigned replicas_count() const;
     bool has(const dsn::gpid &pid) const;
     unsigned remove(const dsn::gpid &pid);
     void update_disk_stat();
+
+    void set_disk_density(int density) { disk_density = density; }
```
hycdong: Is this function only used for function `compute_disk_density`?

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -889,6 +889,34 @@ void replica_stub::on_query_replica_info(const query_replica_info_request &req,
...
+        info.full_dir = dir_node->full_dir;
+        info.disk_capacity_mb = dir_node->disk_capacity_mb;
+        info.disk_available_mb = dir_node->disk_available_mb;
+
+        resp.disk_infos.emplace_back(info);
+    }
+
+    resp.total_capacity_mb = _fs_manager._total_capacity_mb;
+    resp.total_available_mb = _fs_manager._total_available_mb;
+    resp.err = ERR_OK;
```
hycdong: Is this function always got ERR_OK in response?

src/dist/replication/replication.thrift
```diff
@@ -436,6 +436,30 @@ struct query_replica_info_response
     2:list<replica_info>  replicas;
 }
 
+struct disk_info
+{
+    1:string tag;
```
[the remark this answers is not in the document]
Shuo-Jia: Indeed, I haven't accurate anwser. I will test it in test-cluster after related pr complete.

src/dist/replication/lib/replica_stub.cpp
Shuo-Jia: I don't think it makes much difference, because it won't report falcon for "health monitor", but only for `shell` query some time.

src/dist/replication/common/fs_manager.h
Shuo-Jia: Yeah, we just used it to judge whether it is balanced.

src/dist/replication/common/fs_manager.h
Shuo-Jia: Yes

src/dist/replication/lib/replica_stub.cpp
Shuo-Jia: So far it seems, because the code don't have other error such as `ERR_OBJECT_NOT_FOUND`. Indeed, I refer to https://github.com/XiaoMi/rdsn/blob/f6604cbda06cd5aa2ac9d64ca299ef0d98751f04/src/dist/replication/lib/replica_stub.cpp#L859

src/dist/replication/common/fs_manager.h
hycdong: I don't understand how can `disk_density` identify balance, could you please explain about it?

src/dist/replication/common/fs_manager.h
hycdong: If this function is only used for `compute_disk_density`, and this function is simple, there is no need to write a function for only one sentence.

src/dist/replication/lib/replica_stub.cpp
hycdong: oKay~

src/dist/replication/replication.thrift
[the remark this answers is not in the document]
hycdong: Ok, I suggest adding a TODO here.

src/dist/replication/common/fs_manager.h
Shuo-Jia: It's actually the concept of `range` which used to calculate the `standard deviatio`, Negative means the disk is below average, Positive means above. Well, why you expect it should be positive?

src/dist/replication/common/fs_manager.cpp
```diff
@@ -80,38 +81,39 @@ void dir_node::update_disk_stat()
...
-        derror("update disk space failed: dir = %s", full_dir.c_str());
+        derror_f("update disk space failed: dir = {}", full_dir.c_str());
     }
 }
 
 fs_manager::fs_manager(bool for_test)
 {
     if (!for_test) {
-        _counter_capacity_total_mb.init_app_counter("eon.replica_stub",
+        _counter_total_capacity_mb.init_app_counter("eon.replica_stub",
```
neverchanje: why change this name? you make this PR change too bloated. if you do feel this name is not appropriate, make this rename another PR.

Shuo-Jia: acelyc111 suggest modify

src/dist/replication/common/fs_manager.h
hycdong: The comment `it's used to evaluate whether a disk is balanced` raise me a question that how this variety evaluate a disk is balanced or not. 
Just like my previous comment, for example, disk A ratio is 8%, disk B ratio is 11%, total_disk_ratio is 10%, so the density A will be -2, density B will be 1, how can -2 and 1 to evaluate A and B disk balance?
As a result, I thought you would like to use `density` to distinguish which disk is much more not-balanced, just like the example above, disk A is much more not-balanced, and its `density` is expected to be greater than B, but -2 < 1. So I suggest you to guarantee `density` be greater than zero.
However, the meaning of `density` is not what I thought, it can not be used to `evaluate whether a disk is balanced`,  your comment in code confused me.
I suggest you update your comment in code to explain what `disk_density` exactly used for.

Shuo-Jia: Yeah, good advice, the comment should be comprehensible to every one.

src/dist/replication/common/fs_manager.h
```diff
@@ -90,19 +105,52 @@ class fs_manager
     bool for_each_dir_node(const std::function<bool(const dir_node &)> &func) const;
     void update_disk_stat();
 
+    // when executing "disk rebalance", we need find min and max node(sorted_by_density)
+    static bool sorted_by_density(const std::shared_ptr<dir_node> left,
```
neverchanje:
```suggestion
    static bool sorted_by_density(const std::shared_ptr<dir_node>& left, const std::shared_ptr<dir_node>& right)
```
use reference '&'

src/dist/replication/replication.thrift
```diff
@@ -436,6 +436,30 @@ struct query_replica_info_response
...
+}
+
+struct query_disk_info_request
+{
+    1:dsn.rpc_address node;
+}
+
+struct query_disk_info_response
+{
+    1:dsn.error_code err;
```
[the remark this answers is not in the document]
Shuo-Jia: So far, It hasn't other possible error. see https://github.com/XiaoMi/rdsn/pull/409#discussion_r387597036

src/dist/replication/common/fs_manager.cpp
```diff
@@ -80,38 +81,39 @@ void dir_node::update_disk_stat()
...
         disk_available_mb = info.available / 1024 / 1024;
         disk_available_ratio =
-            disk_capacity_mb == 0 ? 0 : disk_available_mb * 100 / disk_capacity_mb;
-        ddebug("update disk space succeed: dir = %s, capacity_mb = %" PRId64
-               ", available_mb = %" PRId64 ", available_ratio = %" PRId64 "%%",
-               full_dir.c_str(),
-               disk_capacity_mb,
-               disk_available_mb,
-               disk_available_ratio);
+            static_cast<int>(disk_capacity_mb == 0 ? 0 : std::round((double)disk_available_mb *
```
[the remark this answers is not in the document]
Shuo-Jia: 1、Why don‘t use `expression? expression 1: expression 2`?
2、Is result of std::lldiv(x,y).quot not equal std::round()? for example:std::round(19.0/4)=5, but std::lldiv(19,4).quot = 4?  std::round is suggested by acelyc111 and expected result more precisely

src/dist/replication/common/fs_manager.h
```diff
@@ -50,16 +50,31 @@ struct dir_node
     std::string full_dir;
     int64_t disk_capacity_mb;
     int64_t disk_available_mb;
-    int64_t disk_available_ratio;
+    int disk_available_ratio;
+    // disk_density = disk_available_ratio - total_available_ratio, it's used to evaluate whether a
```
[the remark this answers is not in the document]
Shuo-Jia: Copy from hdfs.

src/dist/replication/common/fs_manager.h
Shuo-Jia: Oh, levy5307 suggest it's not necessary for share_ptr.

src/dist/replication/common/fs_manager.h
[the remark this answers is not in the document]
neverchanje: Where is the source? Can you copy the link here?

src/dist/replication/common/fs_manager.h
neverchanje: he's wrong. Copying shared_ptr increases the internal atomic counter. It's not free.

Shuo-Jia: Yeah, I have modified it

src/dist/replication/common/fs_manager.h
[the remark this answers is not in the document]
Shuo-Jia: [DiskBalancerVolume.java#L50](https://github.com/apache/hadoop/blob/trunk/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/diskbalancer/datamodel/DiskBalancerVolume.java#L50)
[DiskBalancerVolumeSet.java#L155](https://github.com/apache/hadoop/blob/trunk/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/diskbalancer/datamodel/DiskBalancerVolumeSet.java#L155)

src/dist/replication/replication.thrift
```diff
@@ -436,6 +436,33 @@ struct query_replica_info_response
...
+    6:map<i32,i32> holding_secondary_replica_counts;
+}
+
+// This request is sent from client to replica_server.
+struct query_disk_info_request
+{
+    1:dsn.rpc_address node;
+}
+
+// This response is recieved from client for replica_server.
```
neverchanje:
```suggestion
// This response is received from replica_server.
```
Actually no comment is OK... Because the comment on `query_disk_info_request` is clear enough.

src/dist/replication/test/replica_test/unit_test/mock_utils.h
```diff
@@ -211,6 +211,28 @@ class mock_replica_stub : public replica_stub
...
     }
 
+    void generate_replicas_base_dir_nodes_for_app(app_info mock_app,
+                                                  int primary_count_for_disk = 1,
+                                                  int secondary_count_for_disk = 2)
+    {
+        const auto &dir_nodes = _fs_manager._dir_nodes;
+        for (auto &dir_node : dir_nodes) {
+            for (const auto &holding_replicas : dir_node->holding_replicas) {
+                const std::set<gpid> pids = holding_replicas.second;
```
levy5307: ```const std::set<gpid> &pids = holding_replicas.second;```

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -902,6 +902,83 @@ void replica_stub::on_query_replica_info(const query_replica_info_request &req,
     resp.err = ERR_OK;
 }
 
+void replica_stub::on_query_disk_info(const query_disk_info_request &req,
```
hycdong: I suggest add comment for this function to show it executes in default thread pool, like this:
`// ThreadPool: THREAD_POOL_DEFAULT`

src/dist/replication/test/replica_test/unit_test/replica_disk_test.cpp
```diff
@@ -0,0 +1,167 @@
...
+            }
+
+            stub->_fs_manager._dir_nodes.emplace_back(node_disk);
+            num--;
+        }
+    }
+};
+
+TEST_F(replica_disk_test, on_query_disk_info)
+{
```
hycdong: Could you please refactor this unit test? This test contains too many test cases and has many duplicated code.
You can write test in different TEST_F, such as `TEST_F(replica_disk_test, query_app_not_existed)`, `TEST_F(replica_disk_test, query_all_app)`. 
Besides, you can also define a test_struct to reduce duplicated code.
You can see more in [http_server_test.cpp](https://github.com/XiaoMi/rdsn/blob/master/src/dist/http/test/http_server_test.cpp)

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -902,6 +902,83 @@ void replica_stub::on_query_replica_info(const query_replica_info_request &req,
...
+                    const app_info &info = *(replica.second)->get_app_info();
+                    if (info.app_id == req.app_id) {
+                        app_existed = true;
+                        break;
+                    }
+                }
+                if (!app_existed) {
+                    for (const auto &closing_replica : _closing_replicas) {
+                        const app_info &info = std::get<2>(closing_replica.second);
+                        if (info.app_id == req.app_id) {
```
acelyc111: Too many if-statements, could you do some refactor?

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -2540,5 +2618,28 @@ void replica_stub::on_notify_primary_split_catch_up(const notify_catch_up_reques
...
+            const std::set<dsn::gpid> &pids = holding_replicas.second;
+            for (const auto &pid : pids) {
+                replica_ptr replica = get_replica(pid);
+                if (replica != nullptr) {
+                    if (replica->status() == partition_status::PS_PRIMARY) {
+                        dir_node->holding_primary_replicas[holding_replicas.first].emplace(pid);
+                    } else if (replica->status() == partition_status::PS_SECONDARY) {
+                        dir_node->holding_secondary_replicas[holding_replicas.first].emplace(pid);
+                    }
+                }
```
acelyc111:
```suggestion
                if (!replica != nullptr) {
                    continue;
                }
                if (replica->status() == partition_status::PS_PRIMARY) {
                    dir_node->holding_primary_replicas[holding_replicas.first].emplace(pid);
                } else if (replica->status() == partition_status::PS_SECONDARY) {
                    dir_node->holding_secondary_replicas[holding_replicas.first].emplace(pid);
                }
```

src/dist/replication/test/replica_test/unit_test/mock_utils.h
```diff
@@ -211,6 +211,29 @@ class mock_replica_stub : public replica_stub
...
+                    // generate primary replica and secondary replica.
+                    if (primary_count_for_disk-- > 0) {
+                        add_replica(generate_replica(
+                            mock_app, pid, partition_status::PS_PRIMARY, mock_app.app_id));
+                    } else if (secondary_count_for_disk-- > 0) {
+                        add_replica(generate_replica(
+                            mock_app, pid, partition_status::PS_SECONDARY, mock_app.app_id));
+                    }
+                }
+            }
```
acelyc111:
```suggestion
            if (replica_iter == dir_node->holding_replicas.end()) {
                continue;
            }
            const std::set<gpid> &pids = replica_iter->second;
            for (const gpid &pid : pids) {
                // generate primary replica and secondary replica.
                if (primary_count_for_disk-- > 0) {
                    add_replica(generate_replica(
                        mock_app, pid, partition_status::PS_PRIMARY, mock_app.app_id));
                } else if (secondary_count_for_disk-- > 0) {
                    add_replica(generate_replica(
                        mock_app, pid, partition_status::PS_SECONDARY, mock_app.app_id));
                }
            }
```

src/dist/replication/lib/replica_stub.cpp
Shuo-Jia: Yeah, I also think so. but I have it but not good result. I will re-consider it.

src/dist/replication/lib/replica_stub.cpp
Shuo-Jia: is
```
if (!replica != nullptr) {
                    continue;
                }
```
redundant code？I don't think it necessary.

neverchanje: Of course it's necessary. `replica` can be null.

src/dist/replication/common/fs_manager.h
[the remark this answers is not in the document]
neverchanje: If `disk_densty` is not included in `query_disk_info_response`, we can add this attribute later, when we find it necessary, maybe when we have a more detailed strategy of how to use it.

src/dist/replication/lib/replica_stub.cpp
Shuo-Jia: I know, I means use "continue" is necessary?

neverchanje: I think what @acelyc111 means is to reduce the embed levels of your codes.

Compare to:
```
if(true) {
  if(true) {
  else if(true) {
  }
}
```

apparently the following is better and clearer:

```
if(false) {
  continue;
}
if(true) {
} else if(true) {
}
```

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -902,6 +902,61 @@ void replica_stub::on_query_replica_info(const query_replica_info_request &req,
...
+
+            for (const auto &holding_secondary_replicas : dir_node->holding_secondary_replicas) {
+                info.holding_secondary_replica_counts[holding_secondary_replicas.first] =
+                    static_cast<int>(holding_secondary_replicas.second.size());
+            }
+        } else {
+            int app_id = 0;
+            {
+                zauto_read_lock l(_replicas_lock);
+                if (!(app_id = get_app_id_from_replicas(req.app_name)) &&
```
acelyc111: You will acquire the lock several times, why not move this judgement outside of the loop?

src/dist/replication/replication.thrift
```diff
@@ -436,6 +436,35 @@ struct query_replica_info_response
...
+    2:string full_dir;
+    3:i64 disk_capacity_mb;
+    4:i64 disk_available_mb;
+    // map<i32,i32> means map<app_id, replica_counts>
+    5:map<i32,i32> holding_primary_replica_counts;
+    6:map<i32,i32> holding_secondary_replica_counts;
+}
+
+// This request is sent from client to replica_server.
+struct query_disk_info_request
```
acelyc111: So although we separate shell cmd to 2 cmds, we still share the same RPC?

src/dist/replication/common/fs_manager.h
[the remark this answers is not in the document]
Shuo-Jia: `is not included` is because acelyc111 suggest that if value can be calculated by existed value, it is no need to via `rpc`. so I delete it and re-calc on command side.

src/dist/replication/replication.thrift
Shuo-Jia: Share same rpc. for query info command, I think we can share same rpc, otherwise, it will need too many rpc api. actually, the existed query commands usually only share `list_node` and `list_app` `rpc` api.

PR #411, opened 2020-02-28
feat(dup): handle non-idempotent writes during duplication

neverchanje:
## What problem does this PR solve?

We used to configure the entire cluster to forbid  'non-idempotent writes' like 'incr', 'check_and_set', and 'check_and_mutate', by configuration:

```
[replication]
allow_non_idempotent_write = false
```

Given that once non-idempotent-write is enabled, write-write duplication cannot guarantee eventual-consistency, which is achieved by NTP-timestamp + unique cluster_id. For example:

- a=1
- A: incr a 1 [timestamp=100, cluster_id=1]
- B: incr a 1 [timestamp=100, cluster_id=2]
- A duplicates incr to B, but fails because its cluster_id is lower.
- B duplicates incr to A, succeeds.
- The result is A: a=3, B: a=2

The result is not the same. Take CAS as another example:

- a=1
- A: check if a=1 set a=2 [timestamp=100, cluster_id=1]
- B: check if a=1 set a=3 [timestamp=100, cluster_id=2]
- A duplicates CAS to B, but fails because its cluster_id is lower, so B's value is not changed.
- B duplicates CAS to A, but fails since A's a=2.
- The result is A: a=2, B: a=3

Certainly there's some method to solve this problem (i.e transaction support), but for now it's OK to leave it there. In this PR I remove `allow_non_idempotent_write`, and forbid a duplicating table to accepts non-idempotent writes:

```
>>> add_dup temp onebox2
adding duplication succeed [app: temp, remote: onebox2, appid: 2, dupid: 1582889807, freeze: false]

>>> incr a a 1
ERROR: operation disabled

app_id          : -1
partition_index : -1
decree          : -1
server          : 
```

## New perf-counter

- `dup.disabled_non_idempotent_write_count@<app_name>`

The number of non-idempotent writes that were rejected by replica-server due to duplication.


src/dist/replication/lib/duplication/mutation_batch.cpp
```diff
@@ -103,7 +103,14 @@ add_mutation_if_valid(mutation_ptr &mu, mutation_tuple_set &mutations, decree st
...
             continue;
         }
-
+        // Ignore non-idempotent writes.
+        // Normally a duplicating replica will reply non-idempotent writes with
+        // ERR_OPERATION_DISABLED, but there could still be a mutation written
+        // before the duplication was added.
+        // To ignore means this write will be lost, which is acceptable under this rare case.
+        if (!task_spec::get(update.code)->rpc_request_is_write_idempotent) {
+            continue;
```
acelyc111: I suggest you to add a metric a here, make it 'observable'.

src/dist/replication/lib/replica_2pc.cpp
```diff
@@ -58,7 +58,9 @@ void replica::on_client_write(dsn::message_ex *request, bool ignore_throttling)
     }
 
     task_spec *spec = task_spec::get(request->rpc_code());
-    if (!_options->allow_non_idempotent_write && !spec->rpc_request_is_write_idempotent) {
+    if (is_duplicating() && !spec->rpc_request_is_write_idempotent) {
+        // Ignore non-idempotent write, because duplication provides no guarantee of atomicity to
```
acelyc111: Same.

acelyc111:
Update doc later.

src/dist/replication/common/replication_common.h
```diff
@@ -51,7 +51,6 @@ class replication_options
     bool verbose_commit_log_on_start;
     bool delay_for_fd_timeout_on_start;
     bool empty_write_disabled;
-    bool allow_non_idempotent_write;
```
hycdong: We used `allow_non_idempotent_write` to config whether a cluster support non-idempotent operation. Removing this option means all clusters will not restrict non-idempotent operation. Is it safe to do this? Is there other condition that should not coexistent with non-idempotent except duplication?

src/dist/replication/lib/duplication/test/duplication_test_base.h
```diff
@@ -13,6 +13,9 @@
 namespace dsn {
 namespace replication {
 
+DEFINE_STORAGE_WRITE_RPC_CODE(RPC_DUPLICATION_IDEMPOTENT_WRITE, false, true)
```
hycdong: Suggest update definition like below which is clearer.
`DEFINE_STORAGE_WRITE_RPC_CODE(RPC_DUPLICATION_IDEMPOTENT_WRITE, NOT_ALLOW_BATCH, IS_IDEPOTENT)`

src/dist/replication/lib/duplication/test/load_from_private_log_test.cpp
```diff
@@ -17,7 +17,7 @@
 namespace dsn {
 namespace replication {
 
-DEFINE_TASK_CODE_RPC(RPC_RRDB_RRDB_PUT, TASK_PRIORITY_COMMON, ::dsn::THREAD_POOL_DEFAULT);
+DEFINE_STORAGE_WRITE_RPC_CODE(RPC_RRDB_RRDB_PUT, true, true)
```
hycdong: Same suggestion

src/dist/replication/common/replication_common.h
neverchanje: > Is there other condition that should not coexistent with non-idempotent except duplication?

No. If there is, we should forbid them all.

src/dist/replication/lib/duplication/test/duplication_test_base.h
neverchanje: Reasonable. We can do this refactoring for all our usages of `DEFINE_STORAGE_WRITE_RPC_CODE`, but later, not in this PR.

src/dist/replication/lib/duplication/mutation_batch.cpp
neverchanje: This may be a little bit useless. I've added the counter on 2PC.

PR #414, opened 2020-03-11
feat: forbid large-size-value writes to Pegasus

Shuo-Jia:
related issue: https://github.com/XiaoMi/pegasus/issues/490

New Configuration:

```ini
[replication]
  max_allowed_write_size = 1048576 # default = 1MB
```

New perf-counter

```
eon.replica_stub*recent_write_size_exceed_threshold_count
```


src/dist/replication/lib/replica_2pc.cpp
```diff
@@ -44,6 +44,16 @@ void replica::on_client_write(dsn::message_ex *request, bool ignore_throttling)
         return;
     }
 
+    if (_stub->_max_allowed_write_size &&
```
acelyc111: Add `dsn_unlikely` on this condition.

src/dist/replication/lib/replica_2pc.cpp
```diff
@@ -44,6 +44,16 @@ void replica::on_client_write(dsn::message_ex *request, bool ignore_throttling)
...
+        request->body_size() > _stub->_max_allowed_write_size) {
+        dwarn_replica("client from {} write request body size exceed threshold = {}, it will be "
```
acelyc111: Is it able to logging more info, like gpid. Of cource, logging key would be more helpful, but it seems impossible here.

src/dist/replication/lib/replica_2pc.cpp
```diff
@@ -44,6 +44,16 @@ void replica::on_client_write(dsn::message_ex *request, bool ignore_throttling)
...
+                      "reject!",
+                      request->header->from_address.to_string(),
+                      _stub->_max_allowed_write_size);
+        _stub->_counter_recent_write_size_exceed_threshold_count->increment();
+        response_client_write(request, ERR_INVALID_DATA);
```
acelyc111: Should return here.

Shuo-Jia: Yeah, I add it but del it accidentally.

src/dist/replication/lib/replica_2pc.cpp
Shuo-Jia: 1. The log refer to “abnormal get” log. I think it not more than later.
2.The response will result in client exception, we can get the more exception such as key (https://github.com/XiaoMi/pegasus-java-client/issues/92).

src/dist/replication/lib/replica_2pc.cpp
```diff
@@ -44,6 +44,19 @@ void replica::on_client_write(dsn::message_ex *request, bool ignore_throttling)
         return;
     }
 
+    if (dsn_unlikely(_stub->_max_allowed_write_size &&
+                     request->body_size() > _stub->_max_allowed_write_size)) {
+        dwarn_replica("client from {} write request body size exceed threshold = {}, gpid = "
+                      "({}.{}), it will be reject!",
+                      request->header->from_address.to_string(),
+                      _stub->_max_allowed_write_size,
+                      get_gpid().get_app_id(),
```
levy5307: get_gpid().to_string()

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -85,6 +85,12 @@ replica_stub::replica_stub(replica_state_subscriber subscriber /*= nullptr*/,
     _log = nullptr;
     _primary_address_str[0] = '\0';
     install_perf_counters();
+
+    _max_allowed_write_size = dsn_config_get_value_uint64(
+        "pegasus.server",
+        "max_allowed_write_size",
+        1000000,
```
levy5307: I think use 1e6 is better. Because use 1000000, we must count the number of '0' to get the value.

src/dist/replication/test/replica_test/unit_test/replica_test.cpp
```diff
@@ -0,0 +1,62 @@
...
+    int count = 100;
+    task_code default_code;
+    struct dsn::message_header header;
+    header.body_length = 10000000;
+
+    auto write_request = dsn::message_ex::create_request(default_code);
+    write_request->header = &header;
+    while (count-- > 0) {
+        stub->on_client_write(pid, write_request);
+        calc_write_size_exceed_threshold_count();
```
levy5307: I think you can move ```calc_write_size_exceed_threshold_count()``` out of while

src/dist/replication/lib/replica_stub.cpp
acelyc111: `1<<20` would be better, it represent 1M accurately

src/dist/replication/test/replica_test/unit_test/replica_test.cpp
Shuo-Jia: No, I have test it. `_counter_recent_write_size_exceed_threshold_count` value is not "accumulative value"(it equal 0 or 1), you need accumulate it every time.

src/dist/replication/lib/replica_2pc.cpp
Shuo-Jia: I found `dwarn_replica` will auto add `gpid@ip` info, such as `[2.1@127.0.0.1]`, so it can be del

src/dist/replication/lib/replica_2pc.cpp
```diff
@@ -44,6 +44,17 @@ void replica::on_client_write(dsn::message_ex *request, bool ignore_throttling)
         return;
     }
 
+    if (dsn_unlikely(_stub->_max_allowed_write_size &&
+                     request->body_size() > _stub->_max_allowed_write_size)) {
+        dwarn_replica(
+            "client from {} write request body size exceed threshold = {}, it will be reject!",
+            request->header->from_address.to_string(),
+            _stub->_max_allowed_write_size);
```
acelyc111: Log request->body_size()  too

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -85,6 +85,13 @@ replica_stub::replica_stub(replica_state_subscriber subscriber /*= nullptr*/,
     _log = nullptr;
     _primary_address_str[0] = '\0';
     install_perf_counters();
+
+    _max_allowed_write_size = dsn_config_get_value_uint64("pegasus.server",
+                                                          "max_allowed_write_size",
+                                                          1048576,
```
acelyc111: Use `1<<20` directly.

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -322,6 +329,12 @@ void replica_stub::install_perf_counters()
                                                       "recent.write.busy.count",
                                                       COUNTER_TYPE_VOLATILE_NUMBER,
                                                       "write busy count in the recent period");
+
+    _counter_recent_write_size_exceed_threshold_count.init_app_counter(
+        "eon.replica_stub",
+        "recent.write.size.exceed.threshold.count",
```
acelyc111: `.` is a special character in regular expression, we'd better not use it, use `_` instead.
Some old perf_counter names contain `.`, but it's not a good naming.

src/dist/replication/test/replica_test/unit_test/replica_test.cpp
```diff
@@ -0,0 +1,60 @@
...
+        _app_info.app_type = "replica";
+        _app_info.is_stateful = true;
+        _app_info.max_replica_count = 3;
+        _app_info.partition_count = 8;
+    }
+};
+
+TEST_F(replica_test, write_size_limited)
+{
+    int count = 100;
```
levy5307: I think use loop is better than count

Shuo-Jia: It‘s corresponding to `count` of `write_size_exceed_threshold_count()` , and it actually it is inconsequential.

levy5307: OK, I got it. And you can write these code like this:
```
int count = 100;
for (int i =0; i < count; i++) {
        stub->on_client_write(pid, write_request);
 }
ASSERT_EQ(get_write_size_exceed_threshold_count(), count);
```

src/dist/replication/test/replica_test/unit_test/replica_test.cpp
```diff
@@ -0,0 +1,60 @@
...
+    task_code default_code;
+    struct dsn::message_header header;
+    header.body_length = 10000000;
+
+    auto write_request = dsn::message_ex::create_request(default_code);
```
levy5307: write_request should be deleted after use. Or it will produce memory leak.

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -85,6 +85,13 @@ replica_stub::replica_stub(replica_state_subscriber subscriber /*= nullptr*/,
     _log = nullptr;
     _primary_address_str[0] = '\0';
     install_perf_counters();
+
+    _max_allowed_write_size = dsn_config_get_value_uint64("pegasus.server",
```
neverchanje: section name should be replication, not pegasus.server

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -322,6 +329,12 @@ void replica_stub::install_perf_counters()
                                                       "recent.write.busy.count",
                                                       COUNTER_TYPE_VOLATILE_NUMBER,
                                                       "write busy count in the recent period");
+
+    _counter_recent_write_size_exceed_threshold_count.init_app_counter(
+        "eon_replica_stub",
```
neverchanje: eon_replica_stub => eon.replica_stub

src/dist/replication/test/replica_test/unit_test/replica_test.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include <gtest/gtest.h>
+
+#include <dsn/utility/fail_point.h>
+#include "replica_test_base.h"
+#include <dsn/utility/defer.h>
+
+namespace dsn {
+namespace replication {
+
+class replica_test : public replica_test_base
```
hycdong: `replica_test` seems too common, could you please rename this class name?

Shuo-Jia: The class aims to test some common unit in future

src/dist/replication/lib/replica_2pc.cpp
```diff
@@ -44,6 +45,18 @@ void replica::on_client_write(dsn::message_ex *request, bool ignore_throttling)
...
 
+    if (dsn_unlikely(_stub->_max_allowed_write_size &&
+                     request->body_size() > _stub->_max_allowed_write_size)) {
+        dwarn_replica("client from {} write request body size exceed threshold, request_body_size "
+                      "= {}, max_allowed_write_size = {}, it will be rejected!",
+                      request->header->from_address.to_string(),
+                      request->body_size(),
+                      _stub->_max_allowed_write_size);
+        _stub->_counter_recent_write_size_exceed_threshold_count->increment();
+        response_client_write(request, ERR_INVALID_DATA);
```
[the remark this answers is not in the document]
Shuo-Jia: It will throw exception like `ERR_BUSY` log.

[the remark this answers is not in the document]
hycdong: Will client try to query meta server if it received `ERR_INVALID_DATA`?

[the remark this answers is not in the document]
Shuo-Jia: No, It only throw exception.

PR #415, opened 2020-03-11
feat(dup): support replica http api for duplication state query

neverchanje:
## API Example

```
http://0.0.0.0:34801/replica/duplication?appid=2
```

```json
{
    "1583820008": {
        "2.1": {
            "duplicating": true,
            "not_confirmed_mutations_num": 3,
            "not_duplicated_mutations_num": 3
        },
        "2.3": {
            "duplicating": true,
            "not_confirmed_mutations_num": 2,
            "not_duplicated_mutations_num": 1
        },
        "2.4": {
            "duplicating": true,
            "not_confirmed_mutations_num": 2,
            "not_duplicated_mutations_num": 2
        },
        "2.7": {
            "duplicating": true,
            "not_confirmed_mutations_num": 3,
            "not_duplicated_mutations_num": 2
        }
    }
}
```

This PR provides the duplication state query on replica-server. 

We already expose the duplication query API on meta-server. But the metadata which is pushed from replica-server may not be completely reliable, in case of some bug, or network failure happening. Directly getting the data from replica-server is certainly more accurate and reliable for this purpose.

Because secondaries do not participate in duplication, only the primaries are printed in JSON.


src/dist/replication/lib/duplication/test/replica_http_service_test.cpp
```diff
@@ -0,0 +1,60 @@
...
+              R"({)"
+              R"("1583306653":)"
+              R"({"1.1":{"confirmed_decree":1000,"duplicating":false,"last_decree":1000}},)"
+              R"("non-primaries":)"
+              R"({"1.2":{"confirmed_decree":899}})"
+              R"(})");
+}
+
+} // namespace replication
+} // namespace dsn
```
hycdong: new line

src/dist/replication/lib/duplication/duplication_sync_timer.cpp
```diff
@@ -153,5 +153,29 @@ void duplication_sync_timer::start()
                                          DUPLICATION_SYNC_PERIOD_SECOND * 1_s);
 }
 
+std::multimap<dupid_t, duplication_sync_timer::replica_dup_state>
+duplication_sync_timer::get_dup_states(int app_id, /*out*/ bool *app_found)
+{
```
acelyc111: Add `*app_found = false;` at first?

src/dist/replication/lib/duplication/replica_duplicator_manager.cpp
```diff
@@ -128,5 +128,24 @@ int64_t replica_duplicator_manager::get_pending_mutations_count() const
...
 }
 
+std::vector<replica_duplicator_manager::dup_state>
+replica_duplicator_manager::get_dup_states() const
+{
+    zauto_lock l(_lock);
+
+    std::vector<dup_state> ret;
+    ret.reserve(_duplications.size());
+    for (auto &dup : _duplications) {
```
acelyc111:
```suggestion
    for (const auto &dup : _duplications) {
```

src/dist/replication/lib/duplication/replica_duplicator_manager.cpp
```diff
@@ -128,5 +128,24 @@ int64_t replica_duplicator_manager::get_pending_mutations_count() const
...
+        dup_state state;
+        state.dupid = dup.first;
+        state.duplicating = !dup.second->paused();
+        auto progress = dup.second->progress();
+        state.last_decree = progress.last_decree;
+        state.confirmed_decree = progress.confirmed_decree;
+        ret.push_back(state);
```
acelyc111:
```suggestion
        ret.emplace_back(state);
```

src/dist/replication/lib/replica_http_service.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include <fmt/format.h>
+#include "replica_http_service.h"
+#include "duplication/duplication_sync_timer.h"
+
+namespace dsn {
+namespace replication {
+
+void replica_http_service::query_duplication_handler(const http_request &req, http_response &resp)
+{
+    if (!_stub->_duplication_sync_timer) {
```
acelyc111: In the future refactor, we should add a pure virtual validate function in base class `http_service`, and it will be called before this 'handler' been called.

src/dist/replication/lib/replica_http_service.h
```diff
@@ -0,0 +1,34 @@
...
+    {
+        register_handler("duplication",
+                         std::bind(&replica_http_service::query_duplication_handler,
+                                   this,
+                                   std::placeholders::_1,
+                                   std::placeholders::_2),
+                         "ip:port/replica/duplication?appid=<appid>");
+    }
+
+    std::string path() const override { return "replica"; }
```
acelyc111: And do not forget this refactor point I've ever suggested.

acelyc111:
What do the partitions not present mean?

neverchanje:
> What do the partitions not present mean?

This API exposes only the primaries on this replica-server. Because secondaries do not participate in duplication, only the primaries are printed.

src/dist/replication/lib/duplication/test/replica_http_service_test.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+#include "dist/replication/lib/replica_http_service.h"
+#include "duplication_test_base.h"
+
+namespace dsn {
+namespace replication {
+
+class replica_http_service_test : public duplication_test_base
+{
+public:
```
acelyc111:
```suggestion
```

PR #418, opened 2020-03-17
feat: append mlog in fixed-size blocks using log_appender

neverchanje:
The current `mutation_log` implementation is unfriendly with such case:

There're some writes being flushed into the disk, in the meantime, new incoming writes are pending in the `log_block` waiting to be flushed.

As the pending `log_block` grows, it could be considerably large (like tens of MB) which is unfriendly for both replica recovery and duplication. I've met a bug while testing duplication that reading a large log block may cause tcmalloc to allocate an inaccessible memory.

This PR limits the log block to 1MB using `log_appender`, reduces the possibility of large block related bug.

`log_appender` is a buffer for multiple log blocks. This PR doesn't change the behavior that **all the pending writes** will be flushed to disk (`mutation_log::commit_pending_mutations`), **no matter how large the size is**. What changed is only the size of the log-block, so that reading the log block can be more smooth, without allocating large memory buffer.

While the buffered writes went flushed (which blocks another flush), the incoming writes will be buffered in a new `log_appender`. The old `log_appender` will be removed.

### DIFF
--- a/src/dist/replication/lib/log_block.cpp
+++ b/src/dist/replication/lib/log_block.cpp
@@ -27,8 +27,6 @@ void log_appender::append_mutation(const mutation_ptr &mu, const aio_task_ptr &c
         _callbacks.push_back(cb);
     }
     log_block *blk = &_blocks.back();
-    mu->data.header.log_offset = start_offset() + size();
-    mu->write_to([blk](const blob &bb) { blk->add(bb); });
     if (blk->size() > DEFAULT_MAX_BLOCK_BYTES) {
         _full_blocks_size += blk->size();
         _full_blocks_blob_cnt += blk->data().size();
@@ -36,6 +34,8 @@ void log_appender::append_mutation(const mutation_ptr &mu, const aio_task_ptr &c
         _blocks.emplace_back(new_block_start_offset);
         blk = &_blocks.back();
     }
+    mu->data.header.log_offset = blk->start_offset() + blk->size();
+    mu->write_to([blk](const blob &bb) { blk->add(bb); });
 }
 
 } // namespace replication
--- a/src/dist/replication/lib/log_block.cpp
+++ b/src/dist/replication/lib/log_block.cpp
@@ -20,14 +20,22 @@ void log_block::init()
     add(temp_writer.get_buffer());
 }
 
-void log_block::append_mutation(const mutation_ptr &mu, const aio_task_ptr &cb)
+void log_appender::append_mutation(const mutation_ptr &mu, const aio_task_ptr &cb)
 {
     _mutations.push_back(mu);
     if (cb) {
         _callbacks.push_back(cb);
     }
-    mu->data.header.log_offset = _start_offset + size();
-    mu->write_to([this](const blob &bb) { add(bb); });
+    log_block *blk = &_blocks.back();
+    mu->data.header.log_offset = blk->start_offset() + size();
+    mu->write_to([blk](const blob &bb) { blk->add(bb); });
+    if (blk->size() > DEFAULT_MAX_BLOCK_BYTES) {
+        _full_blocks_size += blk->size();
+        _full_blocks_blob_cnt += blk->data().size();
+        int64_t new_block_start_offset = blk->start_offset() + blk->size();
+        _blocks.emplace_back(new_block_start_offset);
+        blk = &_blocks.back();
+    }
 }
 
 } // namespace replication
--- a/src/dist/replication/lib/log_block.cpp
+++ b/src/dist/replication/lib/log_block.cpp
@@ -27,7 +27,7 @@ void log_appender::append_mutation(const mutation_ptr &mu, const aio_task_ptr &c
         _callbacks.push_back(cb);
     }
     log_block *blk = &_blocks.back();
-    mu->data.header.log_offset = blk->start_offset() + size();
+    mu->data.header.log_offset = start_offset() + size();
     mu->write_to([blk](const blob &bb) { blk->add(bb); });
     if (blk->size() > DEFAULT_MAX_BLOCK_BYTES) {
         _full_blocks_size += blk->size();
--- a/src/dist/replication/lib/log_block.h
+++ b/src/dist/replication/lib/log_block.h
@@ -26,8 +26,6 @@ class log_block
 {
     std::vector<blob> _data; // the first blob is log_block_header
     size_t _size{0};         // total data size of all blobs
-    std::vector<mutation_ptr> _mutations;
-    std::vector<aio_task_ptr> _callbacks;
     int64_t _start_offset{0};
 
 public:
@@ -60,12 +58,6 @@ public:
         _data.push_back(bb);
     }
 
-    void append_mutation(const mutation_ptr &mu, const aio_task_ptr &cb);
-
-    const std::vector<mutation_ptr> &mutations() const { return _mutations; }
-
-    const std::vector<aio_task_ptr> &callbacks() const { return _callbacks; }
-
     // return total data size in the block
     size_t size() const { return _size; }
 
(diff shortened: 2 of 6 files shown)


acelyc111:
Add some description.

src/dist/replication/lib/log_block.cpp
```diff
@@ -30,5 +30,45 @@ void log_block::append_mutation(const mutation_ptr &mu, const aio_task_ptr &cb)
...
+{
+    if (!_blk) { // no data appended
+        return {};
+    }
+    if (_full_blocks.empty()) { // no full block
+        return _blk->mutations();
+    }
+    std::vector<mutation_ptr> ret;
+    for (const auto &block : _full_blocks) {
+        std::copy(block->mutations().begin(), block->mutations().end(), std::back_inserter(ret));
```
acelyc111: Why not use `ret.insert(ret.end(), block->mutations().begin(), block->mutations().end())` ?

src/dist/replication/lib/log_block.h
```diff
@@ -76,5 +76,50 @@ class log_block
...
+        if (!_full_blocks.empty()) {
+            return (*_full_blocks.begin())->size();
+        }
+        return _blk->size();
+    }
+
+protected:
+    static constexpr size_t MAX_BLOCK_BYTES = 1 * 1024 * 1024; // 1MB
+
+    std::deque<std::shared_ptr<log_block>> _full_blocks;
```
acelyc111: Add some necessary comments.

src/dist/replication/lib/log_block.h
```diff
@@ -76,5 +76,50 @@ class log_block
     void init();
 };
 
+class log_appender
```
acelyc111: Does this class need thread-safe?

neverchanje: No. I add a comment here.

src/dist/replication/lib/duplication/test/load_from_private_log_test.cpp
```diff
@@ -200,9 +192,19 @@ class load_from_private_log_test : public duplication_test_base
         std::map<gpid, decree> replay_condition;
         replay_condition[id] = 0; // duplicating
         mutation_log::replay_callback cb = [](int, mutation_ptr &) { return true; };
-        mutation_log_ptr mlog = new mutation_log_private(
-            _replica->dir(), private_log_size_mb, id, _replica.get(), 1024, 512, 10000);
-        EXPECT_EQ(mlog->open(cb, nullptr, replay_condition), ERR_OK);
+        mutation_log_ptr mlog;
+
+        int try_cnt = 0;
+        while (try_cnt < 5) {
```
hycdong: Why add retry logic here?

src/dist/replication/lib/mutation_log.cpp
```diff
@@ -110,14 +110,17 @@ void mutation_log_shared::write_pending_mutations(bool release_lock_required)
...
+    auto pr = mark_new_offset(_pending_write->first_block_size(), false);
     dcheck_eq(pr.second, _pending_write->start_offset());
 
     _is_writing.store(true, std::memory_order_release);
 
     // move or reset pending variables
-    std::shared_ptr<log_block> blk = std::move(_pending_write);
+    std::shared_ptr<log_block> blk = _pending_write->take_log_block();
     int64_t start_offset = blk->start_offset();
+    if (_pending_write->empty()) {
```
hycdong: Could you please explain when will condition `_pending_write->empty()` happen?

src/dist/replication/lib/mutation_log.h
```diff
@@ -468,7 +468,7 @@ class mutation_log_shared : public mutation_log
     // bufferring - only one concurrent write is allowed
     mutable zlock _slock;
     std::atomic_bool _is_writing;
-    std::shared_ptr<log_block> _pending_write;
+    std::unique_ptr<log_appender> _pending_write;
```
hycdong: Maybe `_pending_writes` or `_pending_blocks` is more suitable variety name？I just suggest this variety is plural form.

src/dist/replication/lib/log_block.h
```diff
@@ -76,5 +76,62 @@ class log_block
...
+
+    // Consumes one log block from buffer.
+    // This function may not take all the pending writes when there are multiple blocks.
+    std::shared_ptr<log_block> take_log_block();
+
+    size_t size() const { return _full_blocks_size + (_blk ? _blk->size() : 0); }
+    size_t blob_count() const { return _full_blocks_blob_cnt + (_blk ? _blk->data().size() : 0); }
+    bool empty() const { return !_blk; }
+
+    std::vector<mutation_ptr> mutations() const;
```
hycdong: When we will get all mutations? I only see it is used in unit test.

src/dist/replication/lib/log_block.h
```diff
@@ -76,5 +76,62 @@ class log_block
...
+    size_t first_block_size() const
+    {
+        if (!_full_blocks.empty()) {
+            return (*_full_blocks.begin())->size();
+        }
+        return _blk->size();
+    }
+
+protected:
+    static constexpr size_t MAX_BLOCK_BYTES = 1 * 1024 * 1024; // 1MB
```
hycdong: I consider this pull request aims to restrict the size of `log_block` to avoid flushing a very large block, the default size is 1MB, could you please explain the reason why you set this value? 
Beside, will this pull request lead to any side effect? This pr will effect write operations, I just be careful about it.

src/dist/replication/lib/duplication/test/load_from_private_log_test.cpp
neverchanje: Because Travis sometimes fails to read from disk, which makes the test failed.

src/dist/replication/lib/mutation_log.h
neverchanje: Yes. But I want to keep changes as few as possible. On the other hand, `_pending_writes` is still understandable in the context. Since the blocks actually contain many "writes".

src/dist/replication/lib/log_block.h
neverchanje: They're called during log learning, you can find the usages in `mutation_log_private::get_learn_state_in_memory`, and `get_in_memory_mutations`,

Replica learning requires both the checkpoint, logs on the disk, and buffered writes in memory, which is `mutations()` used for.

src/dist/replication/lib/log_block.cpp
```diff
@@ -30,5 +30,45 @@ void log_block::append_mutation(const mutation_ptr &mu, const aio_task_ptr &cb)
...
+        return {};
+    }
+    if (_full_blocks.empty()) { // no full block
+        return _blk->mutations();
+    }
+    std::vector<mutation_ptr> ret;
+    for (const auto &block : _full_blocks) {
+        ret.insert(ret.end(), block->mutations().begin(), block->mutations().end());
+    }
+    std::copy(_blk->mutations().begin(), _blk->mutations().end(), std::back_inserter(ret));
```
acelyc111, line 69: You can use ret.insert() too

src/dist/replication/lib/log_block.cpp
```diff
@@ -30,5 +30,45 @@ void log_block::append_mutation(const mutation_ptr &mu, const aio_task_ptr &cb)
...
+
+std::vector<mutation_ptr> log_appender::mutations() const
+{
+    if (!_blk) { // no data appended
+        return {};
+    }
+    if (_full_blocks.empty()) { // no full block
+        return _blk->mutations();
+    }
+    std::vector<mutation_ptr> ret;
```
acelyc111, line 65: Reserve ret's size at first.

PR #419, opened 2020-03-18
feat: add perf-counter for backup request

levy5307:
Add a perf-counter to statistics the qps of backup request for each app.

Related issue: https://github.com/XiaoMi/pegasus/issues/251

### DIFF
--- a/src/dist/replication/lib/replica.cpp
+++ b/src/dist/replication/lib/replica.cpp
@@ -82,8 +82,8 @@ replica::replica(
     // init table level latency perf counters
     init_table_level_latency_counters();
 
-    counter_str = fmt::format("table_level_backup_request_qps@{}", _app_info.app_name);
-    _counter_table_level_backup_request_qps.init_app_counter(
+    counter_str = fmt::format("backup_request_qps@{}", _app_info.app_name);
+    _counter_backup_request_qps.init_app_counter(
         "eon.replica", counter_str.c_str(), COUNTER_TYPE_RATE, counter_str.c_str());
 
     if (need_restore) {
@@ -171,7 +171,7 @@ void replica::on_client_read(dsn::message_ex *request)
             return;
         }
     } else {
-        _counter_table_level_backup_request_qps->increment();
+        _counter_backup_request_qps->increment();
     }
 
     uint64_t start_time_ns = dsn_now_ns();
--- a/src/dist/replication/lib/replica.cpp
+++ b/src/dist/replication/lib/replica.cpp
@@ -82,6 +82,10 @@ replica::replica(
     // init table level latency perf counters
     init_table_level_latency_counters();
 
+    counter_str = fmt::format("table_level_backup_request_qps@{}", _app_info.app_name);
+    _counter_table_level_backup_request_qps.init_app_counter(
+        "eon.replica", counter_str.c_str(), COUNTER_TYPE_RATE, counter_str.c_str());
+
     if (need_restore) {
         // add an extra env for restore
         _extra_envs.insert(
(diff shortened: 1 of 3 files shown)


src/dist/replication/lib/replica.cpp
```diff
@@ -166,6 +170,8 @@ void replica::on_client_read(dsn::message_ex *request)
             response_client_read(request, ERR_INVALID_STATE);
             return;
         }
+    } else {
+        _counter_table_level_backup_request_qps->increment();
```
Shuo-Jia: don't (can't) distinguish request types(get, multi_get..)?

levy5307: If we need to distinguish get/multi_get, we should add the perf-counter in pegasus, not rdsn. But in pegasus, we can't get whether the request is backup request or not.
But there is still one way to implement it, just like the way `table level latency` use. But I think it is too complicated.

Shuo-Jia: Well, if can implement, we should consider whether distinguish request types and I think it is necessary.

levy5307: Why it is necessary? I think what we need is the total qps of backup request(which means the added server load), but the request type is not very important.

src/dist/replication/test/replica_test/unit_test/replica_test.cpp
```diff
@@ -59,5 +65,22 @@ TEST_F(replica_test, write_size_limited)
     ASSERT_EQ(get_write_size_exceed_threshold_count(), count);
 }
 
+TEST_F(replica_test, backup_request_qps)
+{
+    // create backup request
+    auto backup_request = dsn::message_ex::create_request(task_code());
```
neverchanje:
```suggestion
    message_ptr backup_request = dsn::message_ex::create_request(task_code());
```

You don't need `cleanup`.

neverchanje:
Backup request QPS should provide partition-level counters. The collector then aggregates the QPS of each partition to a table-level counter. Partition-level QPS can help to find the hotspot partition.

neverchanje:
This new counter is still not enough. If the purpose is to prevent `backup_request_delay_ms` configured too low to actually reduce the tail latency, we need to observe the "additional load" brought by backup request. 
One solution is to add a **table-level counter** (on Collector) to aggregate the backup request QPS of all servers. This QPS can be compared with the read-QPS on primary, to calculate the "backup request ratio". If the ratio is larger than 50%, we can suggest the user reduce the delay time.
For now, the collector collects metrics only on primaries. We still need some works on the collector.

src/dist/replication/lib/replica.cpp
```diff
@@ -82,6 +82,10 @@ replica::replica(
     // init table level latency perf counters
     init_table_level_latency_counters();
 
+    counter_str = fmt::format("table_level_backup_request_qps@{}", _app_info.app_name);
```
neverchanje: remove "table_level"

levy5307:
> This new counter is still not enough. If the purpose is to prevent `backup_request_delay_ms` configured too low to actually reduce the tail latency, we need to observe the "additional load" at table-level.
> 
> So on collector, we can aggregate the backup request QPS of all servers. This QPS can be compared with the read-QPS on primary, to calculate the "backup request ratio". If the ratio is larger than 50%, we can suggest the user reduce the delay time.
> 
> For now, the collector collects metrics only on primaries. We still need some works on the collector.

Can falcon do these work?

neverchanje:
@levy5307 No such method currently.

PR #420, opened 2020-03-19
refactor: delete unused log in replica::on_config_sync

levy5307:
Delete unused log in replica::on_config_sync, which is too messy.


src/dist/replication/lib/replica_config.cpp
```diff
@@ -970,8 +970,6 @@ bool replica::update_local_configuration_with_no_ballot_change(partition_status:
 // ThreadPool: THREAD_POOL_REPLICATION
 void replica::on_config_sync(const app_info &info, const partition_configuration &config)
 {
-    ddebug("%s: configuration sync", name());
```
acelyc111: Use `dinfo_replica` instead would be better.
In the future, we may extend our log util to support "log everey n seconds" feature , like `KLOG_EVERY_N_SECS ` in kudu https://github.com/apache/kudu/blob/1650e2f39449a1b89d4c3abba1106b5ad6bf824a/src/kudu/util/logging.h#L164

src/dist/replication/lib/replica_config.cpp
```diff
@@ -970,8 +970,7 @@ bool replica::update_local_configuration_with_no_ballot_change(partition_status:
...
-
+    dinfo_replica("{}: configuration sync", name());
```
acelyc111:
```suggestion
    dinfo_replica("configuration sync");
```

PR #421, opened 2020-03-20
feat: add a new app_env to limit scan time

hycdong:
As pegasus issue [Limit long time scan operator](https://github.com/XiaoMi/pegasus/issues/486) shows, we should set a time threshold for rocksdb iterate operation to avoid long time scan. To make it more flexible, we provide a new app_env to config this option.


src/dist/replication/common/replication_common.cpp
```diff
@@ -618,6 +618,7 @@ const std::string
     replica_envs::ROCKSDB_CHECKPOINT_RESERVE_MIN_COUNT("rocksdb.checkpoint.reserve_min_count");
 const std::string replica_envs::ROCKSDB_CHECKPOINT_RESERVE_TIME_SECONDS(
     "rocksdb.checkpoint.reserve_time_seconds");
+const std::string replica_envs::ROCKSDB_ITERATION_THRESHOLD("replica.rocksdb_iteration_threshold");
```
neverchanje: threshold of what? size or time?

hycdong: time threshold, I will rename it.

PR #427, opened 2020-03-24
refactor: move log_block class from mutation_log.h to separated file

neverchanje:
mutation_log.h is a long file with many class definitions. This PR simply moves some classes, including `log_block_header` and `log_block`, into a separate file, namely 'log_block.h'. 

This PR only makes trivial changes.

### DIFF
--- a/src/dist/replication/lib/log_block.cpp
+++ b/src/dist/replication/lib/log_block.cpp
@@ -12,10 +12,6 @@ log_block::log_block() { init(); }
 void log_block::init()
 {
     log_block_header hdr;
-    hdr.magic = 0xdeadbeef;
-    hdr.length = 0;
-    hdr.body_crc = 0;
-    hdr.local_offset = 0;
 
     binary_writer temp_writer;
     temp_writer.write_pod(hdr);
--- a/src/dist/replication/lib/log_block.cpp
+++ b/src/dist/replication/lib/log_block.cpp
@@ -1,0 +1,27 @@
+
+// Copyright (c) 2017-present, Xiaomi, Inc.  All rights reserved.
+// This source code is licensed under the Apache License Version 2.0, which
+// can be found in the LICENSE file in the root directory of this source tree.
+
+#include "log_block.h"
+
+namespace dsn {
+namespace replication {
+
+log_block::log_block() { init(); }
+
+void log_block::init()
+{
+    log_block_header hdr;
+    hdr.magic = 0xdeadbeef;
+    hdr.length = 0;
+    hdr.body_crc = 0;
+    hdr.local_offset = 0;
+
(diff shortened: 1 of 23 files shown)


src/dist/replication/lib/log_block.h
```diff
@@ -0,0 +1,63 @@
...
+namespace dsn {
+namespace replication {
+
+// each block in log file has a log_block_header
+struct log_block_header
+{
+    int32_t magic;    // 0xdeadbeef
+    int32_t length;   // block data length (not including log_block_header)
+    int32_t body_crc; // block data crc (not including log_block_header)
+    uint32_t
```
levy5307: remote the line break

src/dist/replication/lib/mutation.cpp
```diff
@@ -164,7 +164,7 @@ void mutation::add_client_request(task_code code, dsn::message_ex *request)
     dassert(client_requests.size() == data.updates.size(), "size must be equal");
 }
 
-void mutation::write_to(std::function<void(const blob &)> inserter) const
+void mutation::write_to(const std::function<void(const blob &)> &inserter) const
```
levy5307: Why are you delete the `const` and `&` ?

src/dist/replication/lib/mutation_log.cpp
```diff
@@ -797,7 +797,7 @@ error_code mutation_log::create_new_log_file()
         header_len = logf->write_file_header(temp_writer, _shared_log_info_map);
     }
 
-    log_block *blk = logf->prepare_log_block();
```
levy5307: There will produce memory leak if blk != nullptr.

src/dist/replication/lib/log_block.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+namespace dsn {
+namespace replication {
+
+log_block::log_block() { init(); }
+
+void log_block::init()
+{
+    log_block_header hdr;
+    hdr.magic = 0xdeadbeef;
```
levy5307: I think you should move these code to the constructor of log_block_header

src/dist/replication/lib/log_block.cpp
```diff
@@ -0,0 +1,22 @@
...
+// can be found in the LICENSE file in the root directory of this source tree.
+
+#include "log_block.h"
+
+namespace dsn {
+namespace replication {
+
+log_block::log_block() { init(); }
+
+void log_block::init()
```
hycdong: why add a function called `init` here, rather than write those code in construction function?

neverchanje: I will write another constructor (in next PR) that's gonna reuse this function. So I intentionally leave it here.

PR #429, opened 2020-03-25
feat(dup): support multiple fail modes for duplication

neverchanje:
This PR introduces a new concept, namely "fail_mode" for duplication, in case of permanent failure occurred, fail-modes will allow flexible and extensible failure handling strategies.

replication.thrift

```thrift
// How duplication reacts on permanent failure.
enum duplication_fail_mode
{
    // The default mode. If some permanent failure occurred that makes duplication
    // blocked, it will retry forever until external interference.
    FAIL_SLOW = 0,

    // Skip the writes that failed to duplicate, which means minor data loss on the remote cluster.
    // This will certainly achieve better stability of the system.
    FAIL_SKIP,

    // Stop immediately after it ensures itself unable to duplicate.
    // WARN: this mode kills the server process, replicas on the server will all be effected.
    FAIL_FAST
}
```

We currently support 3 fail-modes. 

- FAIL_SLOW: the default mode that uses monitoring to report failure, requires human interference, like downgrade the failing primary, to solve the problem.

- FAIL_SKIP: skip the failing file (when the file is corrupted), or failing RPC (the remote cluster is possibly down). This mode is useful for those who care high availability more than data safety.

- FAIL_FAST: Certainly, fail-fast (suicide and coredump) is always the simplest error handling, but not ops-friendly, its blast-radius will cover many the irrelevant replicas on the server.

Users are able to dynamically change the fail mode by 'modify_dup', using shell command  'update_fail_mode'.

### DIFF
--- a/include/dsn/dist/replication/replication_types.h
+++ b/include/dsn/dist/replication/replication_types.h
@@ -167,7 +167,8 @@ struct duplication_fail_mode
     enum type
     {
         FAIL_SLOW = 0,
-        FAIL_SKIP = 1
+        FAIL_SKIP = 1,
+        FAIL_FAST = 2
     };
 };
 
--- a/include/dsn/dist/replication/replication_types.h
+++ b/include/dsn/dist/replication/replication_types.h
@@ -162,6 +162,17 @@ struct duplication_status
 
 extern const std::map<int, const char *> _duplication_status_VALUES_TO_NAMES;
 
+struct duplication_fail_mode
+{
+    enum type
+    {
+        FAIL_SLOW = 0,
+        FAIL_SKIP = 1
+    };
+};
+
+extern const std::map<int, const char *> _duplication_fail_mode_VALUES_TO_NAMES;
+
 class mutation_header;
 
 class mutation_update;
@@ -5017,10 +5028,14 @@ inline std::ostream &operator<<(std::ostream &out, const duplication_add_respons
 
 typedef struct _duplication_modify_request__isset
 {
-    _duplication_modify_request__isset() : app_name(false), dupid(false), status(false) {}
+    _duplication_modify_request__isset()
+        : app_name(false), dupid(false), status(false), fail_mode(false)
+    {
+    }
     bool app_name : 1;
     bool dupid : 1;
     bool status : 1;
+    bool fail_mode : 1;
 } _duplication_modify_request__isset;
 
 class duplication_modify_request
@@ -5030,12 +5045,19 @@ public:
     duplication_modify_request(duplication_modify_request &&);
     duplication_modify_request &operator=(const duplication_modify_request &);
     duplication_modify_request &operator=(duplication_modify_request &&);
-    duplication_modify_request() : app_name(), dupid(0), status((duplication_status::type)0) {}
+    duplication_modify_request()
+        : app_name(),
+          dupid(0),
+          status((duplication_status::type)0),
+          fail_mode((duplication_fail_mode::type)0)
+    {
+    }
 
     virtual ~duplication_modify_request() throw();
     std::string app_name;
     int32_t dupid;
     duplication_status::type status;
+    duplication_fail_mode::type fail_mode;
 
     _duplication_modify_request__isset __isset;
 
@@ -5044,6 +5066,8 @@ public:
     void __set_dupid(const int32_t val);
 
     void __set_status(const duplication_status::type val);
+
+    void __set_fail_mode(const duplication_fail_mode::type val);
 
     bool operator==(const duplication_modify_request &rhs) const
     {
@@ -5054,6 +5078,10 @@ public:
         if (__isset.status != rhs.__isset.status)
             return false;
         else if (__isset.status && !(status == rhs.status))
+            return false;
+        if (__isset.fail_mode != rhs.__isset.fail_mode)
+            return false;
+        else if (__isset.fail_mode && !(fail_mode == rhs.fail_mode))
             return false;
         return true;
     }
@@ -5135,7 +5163,7 @@ typedef struct _duplication_entry__isset
           remote(false),
           create_ts(false),
           progress(false),
-          not_confirmed(false)
+          fail_mode(false)
     {
     }
     bool dupid : 1;
@@ -5143,7 +5171,7 @@ typedef struct _duplication_entry__isset
     bool remote : 1;
     bool create_ts : 1;
     bool progress : 1;
-    bool not_confirmed : 1;
+    bool fail_mode : 1;
 } _duplication_entry__isset;
 
 class duplication_entry
@@ -5153,7 +5181,14 @@ public:
     duplication_entry(duplication_entry &&);
     duplication_entry &operator=(const duplication_entry &);
     duplication_entry &operator=(duplication_entry &&);
-    duplication_entry() : dupid(0), status((duplication_status::type)0), remote(), create_ts(0) {}
+    duplication_entry()
+        : dupid(0),
+          status((duplication_status::type)0),
+          remote(),
+          create_ts(0),
+          fail_mode((duplication_fail_mode::type)0)
+    {
+    }
 
     virtual ~duplication_entry() throw();
     int32_t dupid;
@@ -5161,7 +5196,7 @@ public:
     std::string remote;
     int64_t create_ts;
     std::map<int32_t, int64_t> progress;
-    std::map<int32_t, int64_t> not_confirmed;
+    duplication_fail_mode::type fail_mode;
 
     _duplication_entry__isset __isset;
 
@@ -5175,7 +5210,7 @@ public:
 
     void __set_progress(const std::map<int32_t, int64_t> &val);
 
-    void __set_not_confirmed(const std::map<int32_t, int64_t> &val);
+    void __set_fail_mode(const duplication_fail_mode::type val);
 
     bool operator==(const duplication_entry &rhs) const
     {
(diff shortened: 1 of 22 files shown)


include/dsn/dist/replication/replication_types.h
```diff
@@ -162,6 +162,18 @@ struct duplication_status
 
 extern const std::map<int, const char *> _duplication_status_VALUES_TO_NAMES;
 
+struct duplication_fail_mode
+{
+    enum type
+    {
+        FAIL_SLOW = 0,
+        FAIL_SKIP = 1,
```
levy5307: Add these words here:

FAIL_SLOW: the default mode that uses monitoring to report failure, requires human interference, like downgrade the failing primary, to solve the problem.

FAIL_SKIP: skip the failing file (when the file is corrupted), or failing RPC (the remote cluster is possibly down). This mode is useful for those who care high availability more than data safety.

FAIL_FAST: Certainly, fail-fast (suicide and coredump) is always the simplest error handling, but not ops-friendly, its blast-radius will cover many the irrelevant replicas on the server.

include/dsn/dist/replication/replication_types.h
```diff
@@ -5153,15 +5182,22 @@ class duplication_entry
     duplication_entry(duplication_entry &&);
     duplication_entry &operator=(const duplication_entry &);
     duplication_entry &operator=(duplication_entry &&);
-    duplication_entry() : dupid(0), status((duplication_status::type)0), remote(), create_ts(0) {}
+    duplication_entry()
+        : dupid(0),
+          status((duplication_status::type)0),
+          remote(),
+          create_ts(0),
+          fail_mode((duplication_fail_mode::type)0)
```
levy5307: I think use FAIL_SLOW is better than 0 here

src/dist/replication/lib/duplication/load_from_private_log.cpp
```diff
@@ -131,6 +151,25 @@ void load_from_private_log::replay_log_block()
...
+            _err_file_repeats_num++;
+            if (will_fail_skip()) {
+                // skip this file
+                derror_replica("failed loading for {} times, abandon file {} and try next",
+                               _err_file_repeats_num,
+                               _current->path());
+                _err_file_repeats_num = 0;
+
+                auto prev_offset = _current_global_end_offset;
+                if (switch_to_next_log_file()) {
```
levy5307: what is going to happen when swtich fail?

src/dist/replication/lib/duplication/test/load_from_private_log_test.cpp
```diff
@@ -28,6 +28,24 @@ class load_from_private_log_test : public duplication_test_base
...
+            mutation_log_ptr mlog = create_private_log();
+            for (int i = 1; i <= 10; i++) {
+                std::string msg = "hello!";
+                mutation_ptr mu = create_test_mutation(10 * f + i, msg);
+                mlog->append(mu, LPC_AIO_IMMEDIATE_CALLBACK, nullptr, nullptr, 0);
+            }
+            mlog->tracker()->wait_outstanding_tasks();
+            mlog->close();
+        }
+        return 30;
```
levy5307: return files_num * 10

src/dist/replication/lib/duplication/test/load_from_private_log_test.cpp
```diff
@@ -28,6 +28,24 @@ class load_from_private_log_test : public duplication_test_base
         duplicator = create_test_duplicator();
     }
 
+    // return number of entries written
+    int generate_multiple_log_files(int files_num = 3)
+    {
+        // decree ranges from [1, files_num*10)
```
levy5307: the annotation is wrong when files_num <= 0.
And I suggest use `uint32_t files_num` is better.

src/dist/replication/lib/duplication/test/load_from_private_log_test.cpp
```diff
@@ -28,6 +28,24 @@ class load_from_private_log_test : public duplication_test_base
         duplicator = create_test_duplicator();
     }
 
+    // return number of entries written
+    int generate_multiple_log_files(int files_num = 3)
```
levy5307: I think use `uint32_t files_num` is better.

src/dist/replication/lib/duplication/load_from_private_log.cpp
neverchanje: Yes. Switch fail means no next file to duplicate. So I make it retry and wait the next new file to be generated.

include/dsn/dist/replication/replication_types.h
neverchanje: replication_types.h is auto-generated by thrift compiler. We do not change it.

src/dist/replication/test/meta_test/unit_test/meta_duplication_service_test.cpp
```diff
@@ -678,8 +694,56 @@ TEST_F(meta_duplication_service_test, query_duplication_handler)
     ASSERT_EQ(fake_resp.body,
               std::string() + R"({"1":{"create_ts":")" + ts_buf + R"(","dupid":)" +
                   std::to_string(dup->id) +
+                  R"(,"fail_mode":"FAIL_SLOW")"
                   R"(,"remote":"slave-cluster","status":"DS_START"},"appid":2})");
 }
 
+TEST_F(meta_duplication_service_test, fail_mode)
+{
+    std::string test_app = "test-app";
```
levy5307: I rember we don't support the charactor of `-`

neverchanje: '-' is not restrictively forbidden in Pegasus. So creating such a table in unit test is OK.

src/dist/replication/lib/duplication/load_from_private_log.cpp
```diff
@@ -131,6 +151,25 @@ void load_from_private_log::replay_log_block()
                 _current->path(),
                 _start_offset);
             _counter_dup_load_file_failed_count->increment();
+            _err_file_repeats_num++;
+            if (will_fail_skip()) {
```
acelyc111: Add dsn_unlikely for these two judgements?

src/dist/replication/lib/duplication/load_from_private_log.cpp
```diff
@@ -131,6 +151,25 @@ void load_from_private_log::replay_log_block()
...
+                derror_replica("failed loading for {} times, abandon file {} and try next",
+                               _err_file_repeats_num,
+                               _current->path());
+                _err_file_repeats_num = 0;
+
+                auto prev_offset = _current_global_end_offset;
+                if (switch_to_next_log_file()) {
+                    // successfully skip to next file
+                    auto skipped_bytes = _current_global_end_offset - prev_offset;
+                    _counter_dup_load_skipped_bytes_count->add(skipped_bytes);
```
acelyc111: If there is no next file, the couter will not change?

src/dist/replication/lib/duplication/load_from_private_log.cpp
```diff
@@ -131,6 +151,25 @@ void load_from_private_log::replay_log_block()
...
+                    repeat(_repeat_delay);
+                    return;
+                }
+            } else if (will_fail_fast()) {
```
acelyc111: Could you add some suggestion in this log, then the cluster administrator know how to deal with this issue directly.

src/dist/replication/common/duplication_common.cpp
```diff
@@ -43,6 +43,15 @@ namespace replication {
     return it->second;
 }
 
+/*extern*/ const char *duplication_fail_mode_to_string(duplication_fail_mode::type fmode)
+{
+    auto it = _duplication_fail_mode_VALUES_TO_NAMES.find(fmode);
```
acelyc111: duplication_fail_mode::type is just a struct enum, you can set any value you like. If you set a invalid value, server will coredump here?

src/dist/replication/lib/duplication/load_from_private_log.cpp
neverchanje: Yes, because it will retry the current file. No skipping is done.

src/dist/replication/common/duplication_common.cpp
neverchanje: Yes, if so, something must be wrong. Fail early is necessary.

src/dist/replication/lib/duplication/load_from_private_log.cpp
neverchanje: Done. But I still can't provide much help info. In this case, the admin should check the coredump for the cause for loading failure.

src/dist/replication/ddl_lib/replication_ddl_client.cpp
```diff
@@ -1364,6 +1364,16 @@ error_with<duplication_modify_response> replication_ddl_client::change_dup_statu
     return call_rpc_sync(duplication_modify_rpc(std::move(req), RPC_CM_MODIFY_DUPLICATION));
 }
 
+error_with<duplication_modify_response> replication_ddl_client::update_dup_fail_mode(
+    std::string app_name, int dupid, duplication_fail_mode::type fmode)
+{
+    auto req = make_unique<duplication_modify_request>();
```
acelyc111: You can validate `fmode` here to avoid all replica servers coredump, some necessary protection is reasonable.

neverchanje: It's protected on shell side. Something like:

```
auto fmode = fail_mode_str == "slow" ? FAIL_SLOW : FAIL_SKIP;
```

But sure, I can add another validation here.

PR #430, opened 2020-03-25
refactor(backup): delay the removal of checkpoint files produced by cold backup

levy5307:
The checkpoint files were removed immedately when the cold backup is complete. It is harmful to position problems. So this pr delay the remove of checkpoint files produced by cold backup.


src/dist/replication/lib/replica_backup.cpp
```diff
@@ -421,6 +418,15 @@ static bool backup_parse_dir_name(const char *name,
     }
 }
 
+void replica::background_clear_backup_checkpoint(const std::string &policy_name)
+{
+    tasking::enqueue(LPC_BACKGROUND_COLD_BACKUP,
+                     &_tracker,
+                     [this, policy_name]() { clear_backup_checkpoint(policy_name); },
+                     get_gpid().thread_hash(),
+                     std::chrono::minutes(options()->cold_backup_checkpoint_reserve_minutes));
```
hycdong: We don't restrict backup interval time, which is defined by user who created backup policy, it will be safety to check `cold_backup_checkpoint_reserve_minutes < policy.interval_minute`. 
`clear_backup_checkpoint` will list all checkpoints created by the specific policy, if policy interval_minute is shorter than cold_backup_checkpoint_reserve_minutes, it is possible to remove checkpoints which may be useful during backup.

levy5307: when cold backup is completed, primary replica will send backup_request,  which backup_id = 0, to secondaries. So when the secondaries receive this backup_request, they can't get nothing about this cold backup policy. What they can do are simpely remove the backup dir.
Maybe we can refactor these code firstly.

neverchanje: Good point. I suggest adding a debug log here for this edge case.

src/dist/replication/meta_server/meta_backup_service.cpp
```diff
@@ -1204,6 +1204,7 @@ void backup_service::start()
     start_create_policy_meta_root(after_create_policy_meta_root);
 }
 
+// TODO(zhaoliwei) refactor
```
neverchanje: refactor what?

src/dist/replication/meta_server/meta_backup_service.cpp
```diff
@@ -1212,6 +1213,18 @@ void backup_service::add_backup_policy(dsn::message_ex *msg)
     ::dsn::unmarshall(msg, request);
     std::set<int32_t> app_ids;
     std::map<int32_t, std::string> app_names;
+
+    // The backup interval must be greater than checkpoint reserve time.
+    // Or the next cold backup checkpoint may be cleared by the clear operation.
+    if (request.backup_interval_seconds <=
+        _meta_svc->get_options().cold_backup_checkpoint_reserve_minutes * 60) {
+        response.err = ERR_INVALID_PARAMETERS;
+        response.hint_message = "backup interval must be greater than checkpoint reserve time";
```
neverchanje:
```suggestion
        response.hint_message = fmt::format("backup interval must be greater than cold_backup_checkpoint_reserve_minutes={}", _meta_svc->get_options().cold_backup_checkpoint_reserve_minutes);
```

src/dist/replication/meta_server/meta_backup_service.cpp
hycdong: I guess that liwei will refactor function `add_backup_policy`, its param is `message_ex *`

levy5307: I want to format this function, because it is too long and a little messy

PR #432, opened 2020-03-31
refactor(backup): make backup clear decoupled from on_cold_backup

levy5307:
For now, we use the same rpc(RPC_COLD_BACKUP) between clear backup and start backup, which use backup_id = 0 to represent clear backup. But it will lead to messy code.


src/dist/replication/lib/backup/replica_backup_manager.cpp
```diff
@@ -0,0 +1,129 @@
...
+// get all backup checkpoint dirs which belong to the policy
+static bool get_policy_checkpoint_dirs(const std::string &dir,
+                                       const std::string &policy,
+                                       /*out*/ std::vector<std::string> &chkpt_dirs)
+{
+    chkpt_dirs.clear();
+    // list sub dirs
+    std::vector<std::string> sub_dirs;
+    if (!utils::filesystem::get_subdirectories(dir, sub_dirs, false)) {
+        derror("list sub dirs of dir %s failed", dir.c_str());
```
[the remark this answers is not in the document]
levy5307: We can't use derror_replica here, because `get_policy_checkpoint_dirs` is not a member of `replica_backup_manager`. We can't get replica_name here.

src/dist/replication/lib/replica_stub.h
```diff
@@ -100,6 +100,7 @@ class replica_stub : public serverlet<replica_stub>, public ref_counter
     void on_query_app_info(const query_app_info_request &req,
                            /*out*/ query_app_info_response &resp);
     void on_cold_backup(const backup_request &request, /*out*/ backup_response &response);
+    void on_cold_backup_clear(const backup_clear_request &request);
```
hycdong, line 103: I suggest that rename this function to `on_clear_cold_backup`(and same function in `backup_manager.h`), and rename request type as `clear_backup_request`.

src/dist/replication/lib/backup/replica_backup_manager.cpp
```diff
@@ -0,0 +1,130 @@
...
+        }
+    }
+}
+
+void replica_backup_manager::send_clear_request_to_secondaries(const gpid &pid,
+                                                               const std::string &policy_name)
+{
+    backup_clear_request request;
+    request.__set_pid(pid);
+    request.__set_policy_name(policy_name);
```
hycdong, line 121: You can use `request.pid = pid`, filed `pid` in `backup_clear_request` is not optional.

src/dist/replication/lib/backup/replica_backup_manager.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+        _replica->_cold_backup_contexts.erase(request.policy_name);
+    }
+
+    background_clear_backup_checkpoint(request.policy_name);
+}
+
+void replica_backup_manager::background_clear_backup_checkpoint(const std::string &policy_name)
+{
+    ddebug_replica("schedule to clear all checkpoint dirs of policy({}) in {} minutes",
```
hycdong, line 78: after {} minutes

src/dist/replication/lib/backup/replica_backup_manager.cpp
```diff
@@ -0,0 +1,130 @@
...
+        std::string dirname = utils::filesystem::get_file_name(d);
+        if (is_policy_checkpoint(dirname, policy)) {
+            chkpt_dirs.push_back(std::move(dirname));
+        }
+    }
+    return true;
+}
+
+void replica_backup_manager::on_cold_backup_clear(const backup_clear_request &request)
+{
```
hycdong, line 48: This function is only called by secondary replica, I suggest that add comment to show that, and add status check in this function. 
Besides, I notice that secondary replica don't clear checkpoints, only reset context. I remember that secondary replica may generate checkpoints in some cases, at lease I can find out this logic in former code.

src/dist/replication/lib/backup/replica_backup_manager.cpp
levy5307: I think __set_pid is better, because it is more complicable

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -2026,6 +2039,8 @@ void replica_stub::open_service()
     register_rpc_handler(RPC_QUERY_DISK_INFO, "query_disk_info", &replica_stub::on_query_disk_info);
     register_rpc_handler(RPC_QUERY_APP_INFO, "query_app_info", &replica_stub::on_query_app_info);
     register_rpc_handler(RPC_COLD_BACKUP, "ColdBackup", &replica_stub::on_cold_backup);
```
acelyc111, line 2041:
```suggestion
    register_rpc_handler(RPC_COLD_BACKUP, "cold_backup", &replica_stub::on_cold_backup);
```

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -2026,6 +2039,8 @@ void replica_stub::open_service()
...
+    register_rpc_handler(
+        RPC_CLEAR_COLD_BACKUP, "ClearColdBackup", &replica_stub::on_clear_cold_backup);
```
acelyc111, line 2043:
```suggestion
        RPC_CLEAR_COLD_BACKUP, "clear_cold_backup", &replica_stub::on_clear_cold_backup);
```

Issue #426, opened 2020-03-24
There's another way to get heap profile


vagetablechicken:
https://github.com/XiaoMi/rdsn/blob/41670570b6be142d430ac9ae1c1507fba75c94e7/src/dist/http/pprof_http_service.cpp#L327-L335

We use GetHeapSample() which means we must run servers with sampling turned on(set TCMALLOC_SAMPLE_PARAMETER). If we want to get heap profile, we must restart the server. It's completely unnecessary.

https://gperftools.github.io/gperftools/pprof_remote_servers.html says, 
`The other way is to call HeapProfileStart(filename) (from heap-profiler.h), continue to do work, and then, some number of seconds later, call GetHeapProfile() (followed by HeapProfilerStop()). The server can send the output of GetHeapProfile back as the HTTP response to pprof. (Note you must free() this data after using it.) This is similar to how profile requests are handled, below. This technique does not require the application to run with sampling turned on.`

We can consult the implement in Doris for more details.

@hycdong @levy5307

PR #433, opened 2020-04-03
feat: update the way to get heap profile

hycdong:
## What is this pull request solved
As issue #426 and [pprof_remote_server_doc](https://gperftools.github.io/gperftools/pprof_remote_servers.html) shows, there are two ways to get heap profile. 

We used to use `GetHeapSample`, this interface need server started with param `TCMALLOC_SAMPLE_PARAMETER > 0`, usually is configured as zero in production clusters. If we want to get heap profile of the instance, we need to restart server.

Besides, I read source code of pprof, and find out that the heap got from `GetHeapSample` is not completed.  If we set `TCMALLOC_SAMPLE_PARAMETER > 0`, pprof will sample some objects while it is allocating, and create stack tracker structure for them. If `GetHeapSample` is called, pprof will create a heap profile with current sampled objects. As a result, the heap is not completed, besides, it will costs more to handle those stack tracker.

This pull request update the way to get heap profile, the way [doris](https://github.com/apache/incubator-doris) used:

> Call HeapProfileStart(filename) (from heap-profiler.h), continue to do work, and then, some number of seconds later, call GetHeapProfile() (followed by HeapProfilerStop()). The server can send the output of GetHeapProfile back as the HTTP response to pprof.

The difference between those two functions just like a gperftools issue  https://github.com/gperftools/gperftools/issues/783 says:

> heap sampling will tell you roughly which objects are in use by program at the time when sample is taken. And heap profiler will tell you which pieces of code allocated (and possibly freed) how much memory.

Besides, **some pprof action should not be executing concurrently**, this pull request add restriction about it. Just like doris does ([pprof_actions.cpp](https://github.com/apache/incubator-doris/blob/master/be/src/http/action/pprof_actions.cpp)), we should not get heap profile, cpu profile and heap growth at the same time. 
There are different methods to restrict it, we can use lock or atomic flag.
I tried to add lock on the top of the handler, and send get heap profile request concurrently, and one of them will get profile succeed, another will be failed because of timeout. However, the pegasus server will continue to get profile after the first request release lock, server will get profile twice. Fetching heap profile has the possibility to affect performance, it is not the best solution.
As a result, I use atomic flag, and if server is executing pprof action, it will print log and return error immediately. The error message is like below:
```
> go tool pprof -symbolize=none --seconds=5 http://<ip>:<port>/pprof/heap
Fetching profile over HTTP from http:/<ip>:<port>/pprof/heap?seconds=5
Please wait... (5s)
http://<ip>:<port>/pprof/heap: server response: 500 Internal Server Error
failed to fetch any source profiles
```

## Command Sample
`pprof -symbolize=none --seconds=10 http://<host>:<port>/pprof/heap`
Fetch heap profile of replica server will heavily affect read write performance, please set `seconds` as short as possible.
Set symbolize option as none, and set PPROF_BINARY_PATH for pegasus binary dir, and analysis profile locally, just like: `pprof -symbolize=local <profile_path>` or `pprof -http=localhost:3000 -symbolize=local <profile_path>`

## Performance Comparison
normal case | read write rate | running time(s) | TPS (ops/s) | read latency avg | read latency P99 | write latency avg |  write latency P99
-- | -- | -- | -- | -- | -- | -- | -- 
(1) load: 3 clients*10 threads | 0:1 | 194 | 10299 | - | - | 965 | 4152
(2) run: 3 clients*15 threads | 1:3 | 214 | 14032 | 218 | 714 | 1344 | 6499
(3) run: 3 clients*30 threads | 30:1 | 117 | 86461 | 331 | 878 | 730 | 1545
(4) run: 3 clients*30 threads | 1:0 | 194 | 108425 | 273 | 866 | - | -

**Heap profile for 300 seconds while staring ycsb**
profile replica | read write rate | running time(s) | TPS (ops/s) | read latency avg | read latency P99 | write latency avg |  write latency P99
-- | -- | -- | -- | -- | -- | -- | -- 
(1) load: 3 clients*10 threads | 0:1 | 400 | 4997 | - | - | 1994 | 17887
(2) run: 3 clients*15 threads | 1:3 | 441 | 6798 | 252 | 731 | 2814 | 32228
(3) run: 3 clients*30 threads | 30:1 | 344 | 29043 | 971 | 5945 | 2721 | 12727
(4) run: 3 clients*30 threads | 1:0 | 318 | 66001 | 451 | 3438 | - | -

profile meta | read write rate | running time(s) | TPS (ops/s) | read latency avg | read latency P99 | write latency avg |  write latency P99
-- | -- | -- | -- | -- | -- | -- | -- 
(1) load: 3 clients*10 threads | 0:1 | 193 | 10350 | - | - | 960 | 4021
(2) run: 3 clients*15 threads | 1:3 | 221 | 13591 | 215 | 633 | 1390 | 7392
(3) run: 3 clients*30 threads | 30:1 | 121 | 82837 | 345 | 802 | 729 | 1491
(4) run: 3 clients*30 threads | 1:0 | 201 | 97591 | 304 | 877 | - | -

As the result shows, getting replica server heap profile will strongly affect read write throughput and latency. If the target machine belongs to a cluster who is latency-sensitive, you **SHOULD NOT** get heap profile. If the cluster is less latency-sensitive, you can get profile for several seconds, which should also be as short as possible, better not longer than 1 minutes. 
Getting meta server heap profile has a little impact on throughput and latency. We still recommend that you should set time interval shorter than 1 minute.

## Check List
- Need to update document


src/dist/http/pprof_http_service.cpp
```diff
@@ -320,25 +321,32 @@ void pprof_http_service::symbol_handler(const http_request &req, http_response &
...
-static bool is_heap_profile_enabled() { return ::getenv(TCMALLOC_SAMPLE_PARAMETER) != nullptr; }
-
-static bool get_heap_profile(std::string &result)
+void pprof_http_service::heap_handler(const http_request &req, http_response &resp)
 {
-    if (!is_heap_profile_enabled()) {
-        result = "no TCMALLOC_SAMPLE_PARAMETER in env";
-        return false;
+    const std::string SECOND = "seconds";
+    const int defaultSecond = 10;
```
acelyc111, line 327: Rename to `kDefaultSecond`

src/dist/http/pprof_http_service.cpp
```diff
@@ -320,25 +321,32 @@ void pprof_http_service::symbol_handler(const http_request &req, http_response &
...
+
+    // get seconds from query params, default value is `defaultSecond`
+    int seconds = defaultSecond;
+    const auto iter = req.query_args.find(SECOND);
+    if (iter != req.query_args.end()) {
+        const auto seconds_str = iter->second;
+        if (!seconds_str.empty()) {
+            seconds = std::atoi(seconds_str.c_str());
```
acelyc111, line 335: Use dsn convert utils instead of `atoi`

src/dist/http/pprof_http_service.cpp
```diff
@@ -320,25 +321,32 @@ void pprof_http_service::symbol_handler(const http_request &req, http_response &
...
-
-static bool is_heap_profile_enabled() { return ::getenv(TCMALLOC_SAMPLE_PARAMETER) != nullptr; }
-
-static bool get_heap_profile(std::string &result)
+void pprof_http_service::heap_handler(const http_request &req, http_response &resp)
 {
-    if (!is_heap_profile_enabled()) {
-        result = "no TCMALLOC_SAMPLE_PARAMETER in env";
-        return false;
+    const std::string SECOND = "seconds";
```
acelyc111, line 326: We should prevent this function concurrent invoking.

src/dist/http/pprof_http_service.cpp
```diff
@@ -320,25 +321,32 @@ void pprof_http_service::symbol_handler(const http_request &req, http_response &
...
+    profile_name_prefix << "heap_profile." << getpid() << "." << dsn_now_ns();
+
+    HeapProfilerStart(profile_name_prefix.str().c_str());
+    sleep(seconds);
+    const char *profile = GetHeapProfile();
+    HeapProfilerStop();
 
-    get_heap_profile(resp.body);
+    resp.status_code = http_status_code::ok;
+    resp.body = profile;
```
acelyc111, line 348: Has the problem that response too large and http server can not deal with been solved?

hycdong: I have read source code, and find that GetHeapProfile will return buffers at most 1MB. 
```
// We use buffers of this size in DoGetHeapProfile.
static const int kProfileBufferSize = 1 << 20;
```
During my test, I have observed that some middle profile files whose size will not exceed 1M, it proves the file size limit.
Besides, I have read google pprof source code roughly, find that pprof toolset will collect several heap profile files, and combine them into one file, then serialize, compress the file. I suppose that pprof toolset did something to get whole profile file.

src/dist/http/pprof_http_service.cpp
hycdong: Yes, you are right~ We should prevent this function not executing concurrently, and we may should also prevent pprof function executing concurrently. For example, we should not try to get heap profile and get growth at the same time. I will add a TODO comment here, and research about it.

src/dist/http/pprof_http_service.cpp
```diff
@@ -319,26 +321,39 @@ void pprof_http_service::symbol_handler(const http_request &req, http_response &
...
 // == ip:port/pprof/heap == //
 //                          //
+void pprof_http_service::heap_handler(const http_request &req, http_response &resp)
+{
+    if (_in_pprof_action.load()) {
+        dwarn_f("node is already exectuting pprof action, please wait and retry");
+        resp.status_code = http_status_code::internal_server_error;
+        return;
+    }
+    _in_pprof_action.store(true);
```
acelyc111, line 331: It's better to use CAS operation here.

src/dist/http/pprof_http_service.cpp
```diff
@@ -439,6 +463,13 @@ static bool get_cpu_profile(std::string &result, useconds_t seconds)
 
 void pprof_http_service::profile_handler(const http_request &req, http_response &resp)
 {
+    bool in_pprof = false;
+    if (!_in_pprof_action.compare_exchange_strong(in_pprof, true)) {
+        dwarn_f("node is already exectuting pprof action, please wait and retry");
+        resp.status_code = http_status_code::internal_server_error;
```
neverchanje, line 469:
```suggestion
        resp.body = "node is already exectuting pprof action, please wait and retry";
        resp.status_code = http_status_code::internal_server_error;
```

src/dist/http/pprof_http_service.cpp
```diff
@@ -405,9 +420,18 @@ void pprof_http_service::cmdline_handler(const http_request &req, http_response
 
 void pprof_http_service::growth_handler(const http_request &req, http_response &resp)
 {
+    bool in_pprof = false;
+    if (!_in_pprof_action.compare_exchange_strong(in_pprof, true)) {
+        dwarn_f("node is already exectuting pprof action, please wait and retry");
+        resp.status_code = http_status_code::internal_server_error;
```
neverchanje, line 426:
```suggestion
        resp.body = "node is already exectuting pprof action, please wait and retry";
        resp.status_code = http_status_code::internal_server_error;
```

src/dist/http/pprof_http_service.cpp
```diff
@@ -319,26 +321,39 @@ void pprof_http_service::symbol_handler(const http_request &req, http_response &
 //                          //
 // == ip:port/pprof/heap == //
 //                          //
+void pprof_http_service::heap_handler(const http_request &req, http_response &resp)
+{
+    bool in_pprof = false;
+    if (!_in_pprof_action.compare_exchange_strong(in_pprof, true)) {
+        dwarn_f("node is already exectuting pprof action, please wait and retry");
+        resp.status_code = http_status_code::internal_server_error;
```
neverchanje, line 329:
```suggestion
        resp.body = "node is already exectuting pprof action, please wait and retry";
        resp.status_code = http_status_code::internal_server_error;
```

src/dist/http/pprof_http_service.cpp
hycdong: I consider that Filling error msg into response body is to tell client the reason why this request failed. However, I tried to this, and found out the pprof only print http status_code while meet error, like below:
```
> go tool pprof -symbolize=none --seconds=5 http://<ip>:<port>/pprof/heap
Fetching profile over HTTP from http:/<ip>:<port>/pprof/heap?seconds=5
Please wait... (5s)
http://<ip>:<port>/pprof/heap: server response: 500 Internal Server Error
failed to fetch any source profiles
```
I think we should print error log in server side, http status code is enough for client.

PR #343, opened 2019-11-28
feat: optimize tcmalloc release memory

hycdong:
## What this pr solved
Pull request #278 add a new feature that releasing memory periodically by tcmalloc `ReleaseFreeMemory`. `ReleaseFreeMemory` will release all tcmalloc reserved and not-used memory back to operating system. Releasing memory will lock `page heap`, the structure used to allocate memory from operating system and release memory back to operating system, shared by all threads. As a result, releasing all free memory will lock `page heap` for a long time and block all threads, more seriously, it will affect cluster performance. More tcmalloc information can be find in [tcmalloc website](https://gperftools.github.io/gperftools/tcmalloc.html).

So we optimize the releasing memory method, which is similar to kudu's way [(kudu gc_memory).](https://github.com/apache/kudu/blob/3d57198663f3ed6980f4d92720963ddd0f3a1471/src/kudu/util/process_memory.cc#L140) 
We use tcmalloc statistics below:
```
'tcmalloc.pageheap_free_bytes' - tcmalloc reserved but not-used memory
'generic.current_allocated_bytes' - application whole allocated memory
```
We set a maximum percentage `p`, we assume that tcmalloc reserved but not-used memory should not be greater `whole_used_memory * p`, if exceed this threshold, we will release the exceeding memory by `ReleaseToSystem`. To avoid locking `page heap` for a long time, we only release 1MB at a time.

Besides, we make such percentage `mem_release_max_reserved_mem_percentage` configurable dynamically. Cluster administrator can show and change this value by `remote_command`.
```
>>> remote_command -t replica-server replica.mem-release-max-reserved-percentage 5
COMMAND: replica.mem-release-max-reserved-percentage 5

CALL [replica-server] [ip1:port1] succeed: OK
CALL [replica-server] [ip2:port2] succeed: OK
CALL [replica-server] [ip3:port3] succeed: OK

Succeed count: 3
Failed count: 0

```

```
>>> remote_command -t replica-server replica.mem-release-max-reserved-percentage
COMMAND: replica.mem-release-max-reserved-percentage

CALL [replica-server] [[ip1:port1] succeed: mem-release-max-reserved-percentage=5
CALL [replica-server] [[ip2:port2] succeed: mem-release-max-reserved-percentage=5
CALL [replica-server] [[ip3:port3] succeed: mem-release-max-reserved-percentage=5

Succeed count: 3
Failed count: 0

```
## Config updates
Add two new config
```
mem_release_check_interval_ms = 3600000;
[...]

PR #435, opened 2020-04-08
feat: tcmalloc memory release improvements

hycdong:
## What is this pull request solved
In pull request #343 , we optimize tcmalloc release memory. Based on the former implementation, this pull request add more improvements about it:
1. Add a new perf-counter to record tcmalloc release memory
2. Add a new remote command to control release memory or not
3. Fix percentage range check from (0, 100) to (0, 100]

## New perf-counter
`replica*eon.replica_stub*tcmalloc.release.memory.size`

This perf-counter has a **problem** that it will only be updated every interval time. For example, if tcmalloc memory release time interval is one hour, one server releases 1G memory at 10:00, the value won't change until 11:00. As a result, the monitor graph will be like a trapezoid.
If you have a better way to solve this problem, please comment this pull request or discuss with me offline.

## New remote command
command help: `replica.release-tcmalloc-memory <true|false>`

Usage sample:

```
// query current value
>>> remote_command -t replica-server replica.release-tcmalloc-memory
COMMAND: replica.release-tcmalloc-memory

CALL [replica-server] [ip1: port1] succeed: true
CALL [replica-server] [ip2: port2] succeed: true
CALL [replica-server] [ip3: port3] succeed: true

Succeed count: 3
Failed count: 0

// not release memory
>>> remote_command -t replica-server replica.release-tcmalloc-memory false
COMMAND: replica.release-tcmalloc-memory false

CALL [replica-server] [ip1: port1] succeed: OK
CALL [replica-server] [ip2: port2] succeed: OK
CALL [replica-server] [ip3: port3] succeed: OK

Succeed count: 3
Failed count: 0

```

### DIFF
--- a/src/dist/replication/lib/replica_stub.h
+++ b/src/dist/replication/lib/replica_stub.h
@@ -328,8 +328,10 @@ private:
     dsn_handle_t _query_compact_command;
     dsn_handle_t _query_app_envs_command;
     dsn_handle_t _useless_dir_reserve_seconds_command;
+#ifdef DSN_ENABLE_GPERF
     dsn_handle_t _release_tcmalloc_memory_command;
     dsn_handle_t _max_reserved_memory_percentage_command;
+#endif
 
     bool _deny_client;
     bool _verbose_client_log;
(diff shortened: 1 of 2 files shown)


src/dist/replication/lib/replica_stub.h
```diff
@@ -328,13 +328,15 @@ class replica_stub : public serverlet<replica_stub>, public ref_counter
     dsn_handle_t _query_compact_command;
     dsn_handle_t _query_app_envs_command;
     dsn_handle_t _useless_dir_reserve_seconds_command;
+    dsn_handle_t _release_tcmalloc_memory_command;
```
acelyc111, line 331: Define these variables in DSN_ENABLE_GPERF macro

PR #440, opened 2020-04-17
feat(hotkey): add ringbuf in dsn/utility

Smityz:


include/dsn/utility/ringbuf.h
```diff
@@ -0,0 +1,104 @@
...
+5 7 3
+6 7 2
+7 7 1
+
+*/
+
+namespace dsn {
+namespace utils {
+
+template <typename Type, int MaxElements>
```
acelyc111, line 44: Does this class implement yourself?

include/dsn/utility/ringbuf.h
```diff
@@ -0,0 +1,104 @@
...
+class ringbuf
```
acelyc111, line 45: Could you add some unit tests for it? And I think some benchmark is needed.

include/dsn/utility/ringbuf.h
```diff
@@ -0,0 +1,104 @@
...
+{
```
acelyc111, line 46: This class is not thread-safe, it's necessary to add some comment.
And you want to use it for read/write operation hotkey detection, right? Is there any problem if you use it in multi-threads scenario?

include/dsn/utility/ringbuf.h
```diff
@@ -0,0 +1,104 @@
...
+        return _buf[index];
+    }
+
+    Type front() const { return _buf[_head]; }
+
+    bool empty() const
+    {
+        if (_size != 0)
+            return false;
+        return true;
```
acelyc111, line 92:
```suggestion
        return _size == 0;
```

include/dsn/utility/ringbuf.h
```diff
@@ -0,0 +1,104 @@
...
+namespace dsn {
+namespace utils {
+
+template <typename Type, int MaxElements>
+class ringbuf
+{
+public:
+    ringbuf() : _size(0), _head(0), _tail(0) {}
+
+    void push(const Type &obj)
```
acelyc111, line 50: Is a version of `void push(Type &&obj)` needed?

include/dsn/utility/ringbuf.h
```diff
@@ -0,0 +1,104 @@
...
+    {
+        _size++;
+        _size = std::min(_size, MaxElements);
+        if (_tail == _head && _size == MaxElements) {
+            _head++;
+            if (_head >= MaxElements) {
+                _head %= MaxElements;
+            }
```
acelyc111, line 58: Could you encapsulate these code to a function?

include/dsn/utility/ringbuf.h
```diff
@@ -0,0 +1,104 @@
...
+class ringbuf
+{
+public:
+    ringbuf() : _size(0), _head(0), _tail(0) {}
+
+    void push(const Type &obj)
+    {
+        _size++;
+        _size = std::min(_size, MaxElements);
+        if (_tail == _head && _size == MaxElements) {
```
acelyc111, line 54: When `_size == MaxElements`, `_tail == _head ` is definitely true, right?

include/dsn/utility/ringbuf.h
```diff
@@ -0,0 +1,104 @@
...
+            }
+        }
+        _buf[_tail] = obj;
+        _tail++;
+        if (_tail >= MaxElements) {
+            _tail %= MaxElements;
+        }
+    }
+
+    void pop()
```
acelyc111, line 67: You should protect when this `ringbuf` is empty.

include/dsn/utility/ringbuf.h
```diff
@@ -0,0 +1,104 @@
...
+        _head++;
+        _size--;
+        if (_head >= MaxElements) {
+            _head %= MaxElements;
+        }
+        return;
+    }
+
+    Type back() const
+    {
```
acelyc111, line 78: What happen when it's empty?

include/dsn/utility/ringbuf.h
```diff
@@ -0,0 +1,104 @@
...
+        int index = _tail - 1;
+        if (index < 0) {
+            index += MaxElements;
+        }
+        return _buf[index];
+    }
+
+    Type front() const { return _buf[_head]; }
```
acelyc111, line 86: What happen when it's empty?

include/dsn/utility/ringbuf.h
Smityz: Yes

include/dsn/utility/ringbuf.h
Smityz: It's just like `std::queue`, we should add a mutex to keep safety in a multi-threads scenario.

include/dsn/utility/ringbuf.h
Smityz: Yes

PR #443, opened 2020-04-22
feat(cold-backup): add rate limit for fds

levy5307:
This pr is intend to add rate limit for fds. 
For put operation, we can't send a file in small batches. Because `putContent` interface of fds will overwrite what was sent before for the same file. So we must send a file all at one batch. It will produce network burst. 
But for get operation this problem does not exist, because we get can file content in batches.


acelyc111:
I think we should separte this PR into at least 2 PRs, one to add a new thirdparty lib, and the other to use this lib for coldbackup.

src/dist/block_service/fds/fds_service.cpp
```diff
@@ -484,82 +496,134 @@ fds_file_object::fds_file_object(fds_service *s,
...
-                                             /*out*/ uint64_t &transfered_bytes)
+error_code fds_file_object::get_file_meta()
 {
-    dsn::error_code err;
+    galaxy::fds::GalaxyFDSClient *c = _service->get_client();
+    try {
+        auto meta = c->getObjectMetadata(_service->get_bucket_name(), _fds_path)->metadata();
+        auto iter = meta.find(fds_service::FILE_LENGTH_CUSTOM_KEY);
+        if (iter != meta.end()) {
+            _size = atoll(iter->second.c_str());
```
neverchanje, line 506:
```suggestion
            buf2int64(iter->second, _size)
```

src/dist/block_service/fds/fds_service.cpp
```diff
@@ -484,82 +496,134 @@ fds_file_object::fds_file_object(fds_service *s,
...
+    uint64_t once_transfered_bytes = 0;
+    while (pos < start + to_transfer_bytes) {
+        uint64_t batch = std::min(BATCH_MAX, start + to_transfer_bytes - pos);
+        // get tokens from token bucket
+        _service->_token_bucket->consumeWithBorrowAndWait(batch * 8);
+
+        err = get_content(pos, batch, os, once_transfered_bytes);
+        transfered_bytes += once_transfered_bytes;
+        if (err != ERR_OK || once_transfered_bytes < batch) {
+            return err;
```
acelyc111, line 562: When `err == ERR_OK && once_transfered_bytes < batch` , `ERR_OK` is returned, is it what you expect? Seems like the last part of file is sent, you could add some comments.

src/dist/block_service/fds/fds_service.cpp
```diff
@@ -484,82 +496,134 @@ fds_file_object::fds_file_object(fds_service *s,
...
-                                             /*out*/ std::ostream &os,
-                                             /*out*/ uint64_t &transfered_bytes)
+error_code fds_file_object::get_file_meta()
 {
-    dsn::error_code err;
+    galaxy::fds::GalaxyFDSClient *c = _service->get_client();
+    try {
+        auto meta = c->getObjectMetadata(_service->get_bucket_name(), _fds_path)->metadata();
+        auto iter = meta.find(fds_service::FILE_LENGTH_CUSTOM_KEY);
+        if (iter != meta.end()) {
```
acelyc111, line 505: Old code has `dassert(iter != meta.end()`

src/dist/block_service/fds/fds_service.cpp
```diff
@@ -95,10 +97,20 @@ DEFINE_THREAD_POOL_CODE(THREAD_POOL_FDS_SERVICE)
 DEFINE_TASK_CODE(LPC_FDS_CALL, TASK_PRIORITY_COMMON, THREAD_POOL_FDS_SERVICE)
 
 const std::string fds_service::FILE_LENGTH_CUSTOM_KEY = "x-xiaomi-meta-content-length";
-const std::string fds_service::FILE_LENGTH_KEY = "content-length";
 const std::string fds_service::FILE_MD5_KEY = "content-md5";
 
-fds_service::fds_service() {}
+fds_service::fds_service()
+{
+    uint32_t rate_limit = (uint32_t)dsn_config_get_value_uint64(
```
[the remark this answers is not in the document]
levy5307: I think there should only one token bucket which controls the network rate limit. 
If we seperate upload and download limit, we should add another token bucket.

src/dist/block_service/fds/fds_service.cpp
```diff
@@ -484,82 +496,134 @@ fds_file_object::fds_file_object(fds_service *s,
...
+        }
+    }
+}
+
+error_code fds_file_object::get_content_in_batches(uint64_t start,
+                                                   int64_t length,
+                                                   /*out*/ std::ostream &os,
+                                                   /*out*/ uint64_t &transfered_bytes)
+{
+    const uint64_t BATCH_MAX = 1e6;
```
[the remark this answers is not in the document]
levy5307: 1 byte = 8 bits. Because network use Mb/s as it's unit. And the unit of buffer length is byte

src/dist/block_service/fds/fds_service.cpp
```diff
@@ -484,82 +496,134 @@ fds_file_object::fds_file_object(fds_service *s,
...
+    if (-1 == length) {
+        to_transfer_bytes = _size;
+    }
+
+    uint64_t pos = start;
+    uint64_t once_transfered_bytes = 0;
+    while (pos < start + to_transfer_bytes) {
+        uint64_t batch = std::min(BATCH_MAX, start + to_transfer_bytes - pos);
+        // get tokens from token bucket
+        _service->_token_bucket->consumeWithBorrowAndWait(batch * 8);
```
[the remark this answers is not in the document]
levy5307: 1 byte = 8 bits. Because network use Mb/s as it's unit. And the unit of buffer length is byte

src/dist/block_service/fds/fds_service.cpp
```diff
@@ -95,10 +97,20 @@ DEFINE_THREAD_POOL_CODE(THREAD_POOL_FDS_SERVICE)
...
-fds_service::fds_service() {}
+fds_service::fds_service()
+{
+    uint32_t rate_limit = (uint32_t)dsn_config_get_value_uint64(
+        "replication", "fds_limit_rate", 20, "rate limit of fds(Mb/s)");
+    /// For put operation, we can't send a file in batches. Because putContent interface of fds
+    /// will overwrite what was sent before for the same file. So we must send a file as a whole.
+    /// If file size > burst size, the file will be rejected by the token bucket.
+    /// And sst maximum file size is about 64MB, so burst size must be greater than 64MB.
+    uint32_t burst_size = std::max(3 * rate_limit * 1e6, 70e6 * 8);
```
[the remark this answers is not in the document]
levy5307: It is the burst size, which must be great than 64MB

src/dist/block_service/fds/fds_service.cpp
levy5307: Yes, it means we read the end of the file

src/dist/block_service/fds/fds_service.cpp
levy5307: I don't think we should make the server core dump when we encounter some error in fds, so I delete the assert here.

src/dist/block_service/fds/fds_service.cpp
```diff
@@ -484,82 +496,134 @@ fds_file_object::fds_file_object(fds_service *s,
...
+                                        int64_t to_transfer_bytes,
+                                        uint64_t &transfered_bytes)
 {
-    dsn::error_code err = dsn::ERR_OK;
+    error_code err = ERR_OK;
     transfered_bytes = 0;
     galaxy::fds::GalaxyFDSClient *c = _service->get_client();
+
+    // get tokens from token bucket
+    _service->_token_bucket->consumeWithBorrowAndWait(to_transfer_bytes * 8);
```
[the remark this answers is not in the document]
levy5307: What kind of risk? I think it only let the cold back progress of this replica will delay a lot

src/dist/block_service/fds/fds_service.cpp
```diff
@@ -95,10 +97,20 @@ DEFINE_THREAD_POOL_CODE(THREAD_POOL_FDS_SERVICE)
...
 
-fds_service::fds_service() {}
+fds_service::fds_service()
+{
+    uint32_t rate_limit = (uint32_t)dsn_config_get_value_uint64(
+        "replication", "fds_limit_rate", 20, "rate limit of fds(Mb/s)");
+    /// For put operation, we can't send a file in batches. Because putContent interface of fds
+    /// will overwrite what was sent before for the same file. So we must send a file as a whole.
+    /// If file size > burst size, the file will be rejected by the token bucket.
+    /// And sst maximum file size is about 64MB, so burst size must be greater than 64MB.
```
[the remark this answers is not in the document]
levy5307: we can set the burst size when we create a token bucket, see L112

src/dist/block_service/fds/fds_service.cpp
```diff
@@ -484,82 +498,139 @@ fds_file_object::fds_file_object(fds_service *s,
...
+    // get file meta if it is not synced
+    if (!_has_meta_synced) {
+        err = get_file_meta();
+        if (ERR_OK != err) {
+            return err;
+        }
+    }
+
+    // if length = -1, it means we should transfer the whole file
+    uint64_t to_transfer_bytes = length;
```
hycdong, line 554: `uint64_t to_transfer_bytes = length == -1 ? _size : length;` may be better.

src/dist/block_service/fds/fds_service.cpp
```diff
@@ -484,82 +498,139 @@ fds_file_object::fds_file_object(fds_service *s,
...
+        // get tokens from token bucket
+        _service->_token_bucket->consumeWithBorrowAndWait(batch_len);
+
+        err = get_content(pos, batch_len, os, once_transfered_bytes);
+        transfered_bytes += once_transfered_bytes;
+        if (err != ERR_OK || once_transfered_bytes < batch_len) {
+            return err;
+        }
+        pos += batch_len;
+    }
```
hycdong, line 572: Could you please explain when would the condition `pos >= start + to_transfer_bytes` happen to me? I just can not image when would meet such condition. Condition `err == ERR_OK && once_transfered_bytes < batch_size` means read the end of the file, and it return inside while.

src/dist/block_service/fds/fds_service.cpp
```diff
@@ -484,82 +498,139 @@ fds_file_object::fds_file_object(fds_service *s,
 
 fds_file_object::~fds_file_object() {}
 
-dsn::error_code fds_file_object::get_content(uint64_t pos,
-                                             int64_t length,
-                                             /*out*/ std::ostream &os,
-                                             /*out*/ uint64_t &transfered_bytes)
+error_code fds_file_object::get_file_meta()
```
hycdong, line 501: It seems that this function is used to read remote FDS by FDS client to get file's metadata, and `put_conent` and `create_file` also has similar code, I suggest you to refactor them too. If you would like  to refactor it in further pull request, I suggest adding some TODO comments.

src/dist/block_service/fds/fds_service.cpp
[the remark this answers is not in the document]
neverchanje: Yes. There should be two token buckets controlling upload and download separately.
I think they must be treated differently. Upload reads from disk, which consumes less in terms of SSD.

src/dist/block_service/fds/fds_service.cpp
levy5307: `pos == start + to_transfer_bytes` may happen, when `once_transfered_bytes` always equal with `batch_size`, and we read the end of the file.
but `pos > start + to_transfer_bytes` may not happy in normal case. But for the robustness of these code, I can use   `pos >= start + to_transfer_bytes` here.

src/dist/block_service/fds/fds_service.cpp
levy5307: Yes, there are some code here. I will refactor in later pull request. 
I will add some TODO comments.

src/dist/block_service/fds/fds_service.cpp
acelyc111: If not core dump, you should make sure the following logic is valid. Here, when meta doesn't contain `FILE_LENGTH_CUSTOM_KEY ` or `FILE_MD5_KEY `, `_size ` or `_md5sum ` will not be initialized normally.
In most cases, `dassert` (or similar operation) is used to reduce various of defensive program code, and it is helpful to expose problem as early as possible.

src/dist/block_service/fds/fds_service.cpp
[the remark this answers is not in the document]
acelyc111: I meant sst file size is controlled by another config, it maybe 64MB, but also maybe 32MB or 128MB, so the fixed burst size might be less than sst size.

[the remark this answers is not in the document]
levy5307: OK, so I think we should set burst size according to this config

src/dist/block_service/fds/fds_service.cpp
```diff
@@ -484,82 +511,143 @@ fds_file_object::fds_file_object(fds_service *s,
...
+        dassert(valid && _size >= 0, "error to get file size");
+
+        // get md5 key
+        iter = meta.find(fds_service::FILE_MD5_KEY);
+        dassert(iter != meta.end(),
+                "can't find %s in object(%s)'s metadata",
+                fds_service::FILE_MD5_KEY.c_str(),
+                _fds_path.c_str());
+
+        if (iter != meta.end()) {
```
acelyc111, line 536: Now it has been checked by dassert, so you can assign _md5sum directly.

src/dist/block_service/fds/fds_service.cpp
```diff
@@ -484,82 +511,143 @@ fds_file_object::fds_file_object(fds_service *s,
...
+        dassert(iter != meta.end(),
+                "can't find %s in object(%s)'s metadata",
+                fds_service::FILE_LENGTH_CUSTOM_KEY.c_str(),
+                _fds_path.c_str());
+        bool valid = dsn::buf2uint64(iter->second, _size);
+        dassert(valid && _size >= 0, "error to get file size");
+
+        // get md5 key
+        iter = meta.find(fds_service::FILE_MD5_KEY);
+        dassert(iter != meta.end(),
```
acelyc111, line 531: Use dassert_f to simplify code.

src/dist/block_service/fds/fds_service.cpp
```diff
@@ -95,10 +98,33 @@ DEFINE_THREAD_POOL_CODE(THREAD_POOL_FDS_SERVICE)
...
 const std::string fds_service::FILE_LENGTH_CUSTOM_KEY = "x-xiaomi-meta-content-length";
-const std::string fds_service::FILE_LENGTH_KEY = "content-length";
 const std::string fds_service::FILE_MD5_KEY = "content-md5";
 
-fds_service::fds_service() {}
+fds_service::fds_service()
+{
+    const int BYTE_TO_BIT = 8;
+
+    uint64_t sst_file_size = dsn_config_get_value_uint64("pegasus.server",
```
acelyc111, line 107: There is an other config you should notice, that is `write_buffer_size`, it is the memtable size threashold flushed to level 0, which is also contained in backup directory. It's 64MB by default, but in Pegasus, it will be random in 64MB*[0.75, 1.25].
https://github.com/XiaoMi/pegasus/blob/master/src/server/pegasus_server_impl.cpp#L2420

acelyc111: In BULK_LOAD scenario, it's 64MB*4*[0.75, 1.25], although it's rare to backup a table when it's in BULK_LOAD scenario mode, but we could provide a way observe when it happend, for example, add a log, add a metric would be better.

src/dist/block_service/fds/fds_service.cpp
```diff
@@ -95,10 +98,42 @@ DEFINE_THREAD_POOL_CODE(THREAD_POOL_FDS_SERVICE)
...
-const std::string fds_service::FILE_LENGTH_KEY = "content-length";
 const std::string fds_service::FILE_MD5_KEY = "content-md5";
 
-fds_service::fds_service() {}
+fds_service::fds_service()
+{
+    const int BYTE_TO_BIT = 8;
+
+    // get max sst file size in normal scenario
+    uint64_t target_file_size =
```
acelyc111, line 108: These code is some kind of tricky to read configs of 'pegasus.server', and some threshold is not very strict. Add some comments to explain would be better.

src/dist/block_service/fds/fds_service.cpp
```diff
@@ -484,82 +521,146 @@ fds_file_object::fds_file_object(fds_service *s,
...
+        auto meta = c->getObjectMetadata(_service->get_bucket_name(), _fds_path)->metadata();
+
+        // get file length
+        auto iter = meta.find(fds_service::FILE_LENGTH_CUSTOM_KEY);
+        dassert_f(iter != meta.end(),
+                  "can't find {} in object({})'s metadata",
+                  fds_service::FILE_LENGTH_CUSTOM_KEY.c_str(),
+                  _fds_path.c_str());
+        bool valid = dsn::buf2uint64(iter->second, _size);
+        dassert_f(valid && _size >= 0, "error to get file size");
```
acelyc111, line 537: `_size ` is an `uint64_t `, it is always >= 0

src/dist/block_service/fds/fds_service.cpp
```diff
@@ -484,82 +525,146 @@ fds_file_object::fds_file_object(fds_service *s,
...
     transfered_bytes = 0;
     galaxy::fds::GalaxyFDSClient *c = _service->get_client();
+
+    // get tokens from token bucket
+    if (!_service->_write_token_bucket->consumeWithBorrowAndWait(to_transfer_bytes)) {
+        ddebug_f("the transfer count({}) is greater than burst size({}), so it is rejected by "
+                 "token bucket",
+                 to_transfer_bytes,
+                 _service->_write_token_bucket->burst());
+        return ERR_INJECTED;
```
neverchanje, line 665: what does injected mean?

src/dist/block_service/fds/fds_service.cpp
```diff
@@ -95,10 +99,46 @@ DEFINE_THREAD_POOL_CODE(THREAD_POOL_FDS_SERVICE)
...
 const std::string fds_service::FILE_LENGTH_CUSTOM_KEY = "x-xiaomi-meta-content-length";
-const std::string fds_service::FILE_LENGTH_KEY = "content-length";
 const std::string fds_service::FILE_MD5_KEY = "content-md5";
 
-fds_service::fds_service() {}
+fds_service::fds_service()
+{
+    const int BYTE_TO_BIT = 8;
+
+    /// In normarl scenario, the sst file size of level 0 is write_buffer_size * [0.75, 1.25]
```
neverchanje, line 108: normal

src/dist/block_service/fds/fds_service.cpp
```diff
@@ -95,10 +99,46 @@ DEFINE_THREAD_POOL_CODE(THREAD_POOL_FDS_SERVICE)
...
+    uint64_t target_file_size =
+        dsn_config_get_value_uint64("pegasus.server",
+                                    "rocksdb_target_file_size_base",
+                                    64 * 1024 * 1024,
+                                    "rocksdb options.target_file_size_base");
+    uint64_t write_buffer_size = dsn_config_get_value_uint64("pegasus.server",
+                                                             "rocksdb_write_buffer_size",
+                                                             64 * 1024 * 1024,
+                                                             "rocksdb options.write_buffer_size");
+    uint64_t max_sst_file_size = std::max(target_file_size, (uint64_t)1.25 * write_buffer_size);
```
neverchanje, line 122: rDSN & FDS client should not assume anything about Pegasus, as well as the details around rocksdb.

src/dist/block_service/fds/fds_service.cpp
hycdong, line 665: Why return `ERR_INJECTED`?

src/dist/block_service/fds/fds_service.cpp
levy5307: But we should get the file size, or we must set the burst size to a very big number.
Do you have another better way?

src/dist/block_service/fds/fds_service.cpp
```diff
@@ -95,10 +99,46 @@ DEFINE_THREAD_POOL_CODE(THREAD_POOL_FDS_SERVICE)
...
+
+    uint32_t write_rate_limit = (uint32_t)dsn_config_get_value_uint64(
+        "replication", "fds_write_limit_rate", 20, "rate limit of fds(Mb/s)");
```
Shuo-Jia, line 125: consider add `remote_command` for the config?

neverchanje: Runtime update of configurations should be a basic function of our config utility. We don't have to support dynamic updates one by one and case by case.

neverchanje: Maybe we can implement runtime-update of configurations as a basic function of our config utility. We don't have to support dynamic updates one by one and case by case.

src/dist/block_service/fds/fds_service.cpp
acelyc111: Yes, it's bit of tricky here, but in the following plan, we are going to merge rdsn and pegasus projects.

src/dist/block_service/fds/fds_service.cpp
levy5307: > Maybe we can implement runtime-update of configurations as a basic function of our config utility. We don't have to support dynamic updates one by one and case by case.

Yes, I think it is a good idea.

PR #444, opened 2020-04-23
feat(util): add token bucket for rate limit

levy5307:
In this pr, token bucket is added for rate limit, which is migrated from https://github.com/facebook/folly/blob/master/folly/TokenBucket.h

There are some changes here, which are based on folly's version:
- Replace folly::optional with boost::optional. 
   1. Replace folly::optional::value_or() with boost::optional::get_value_or()
   2. Replace folly::optional::value() with boost::optional::get()
- Replace folly::unlikely with dsn_unlikely

### DIFF
--- a/src/core/tests/TokenBucketTest.h
+++ b/src/core/tests/TokenBucketTest.h
@@ -1,0 +1,32 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


src/core/tests/TokenBucketTest.h
```diff
@@ -0,0 +1,32 @@
+/*
```
hycdong, line 1: Does the order of copyright matter? If dose, should we order it by time order or reverse time order? In liwei's pull request, copyright is ordered by time order.

neverchanje:
Please describe what's different between folly's TokenBucket. As far as I can see, `boost::optional` replaced `folly::Optional`. It will be better if you can give a diff for the changes.

src/core/tests/TokenBucketTest.h
neverchanje: Not matter much. We will modify our license after all...

PR #447, opened 2020-04-24
fix(asan): memory leak in local_service.cpp

levy5307:
fix memory leak in local_service.cpp.


src/dist/block_service/local/local_service.cpp
```diff
@@ -412,7 +412,8 @@ dsn::task_ptr local_file_object::read(const read_request &req,
                 }
 
                 dinfo("read file(%s), size = %ld", file_name().c_str(), total_sz);
-                std::shared_ptr<char> buf = std::shared_ptr<char>(new char[total_sz + 1]);
+                std::shared_ptr<char> buf =
```
neverchanje, line 415: Actually we have a better way to implement this:

```
std::string buf;
buf.reserve(total_sz + 1);
...
fin.read(buf, total_sz);
buf[fin.gcount()] = '\0';
resp.buffer = blob::create_from_bytes(std::move(buf));
```

PR #449, opened 2020-04-27
fix: fix memory leak in meta test

levy5307:
fix memory leak in meta test.


src/dist/replication/test/meta_test/unit_test/server_state_test.cpp
```diff
@@ -80,7 +80,8 @@ void meta_service_test_app::app_envs_basic_test()
         request.__set_keys(keys);
         request.__set_values(values);
 
-        dsn::message_ex *binary_req = dsn::message_ex::create_request(RPC_CM_UPDATE_APP_ENV);
+        dsn::ref_ptr<dsn::message_ex> binary_req =
```
neverchanje, line 83: `dsn::message_ptr`

PR #453, opened 2020-04-30
feat(thirdparty): Introduce readerwriterqueue util

Smityz:
We use this lock-free reader-writer queue to capture data in a multi-thread environment.
Ref https://github.com/cameron314/readerwriterqueue


acelyc111:
Could you import it as a thirdparty lib?

thirdparty/build-thirdparty.sh
```diff
@@ -59,6 +59,17 @@ else
     echo "skip build concurrentqueue"
 fi
 
+# build readerwriterqueue
```
levy5307, line 62: lock-free readerwriterqueue

Smityz: Its original name is readerwriterqueue

include/dsn/utility/readerwriterqueue/README.md
```diff
@@ -0,0 +1,122 @@
+# A single-producer, single-consumer lock-free queue for C++
```
acelyc111, line 1: Remove source files if you import it as a thirdparty lib.

PR #461, opened 2020-05-14
feat: add rate limit for learning and support remote-command

Shuo-Jia:
When add or delete replica server node, too large copy rate will have great influence on server, this pr add [rate limit](https://github.com/XiaoMi/rdsn/pull/444)  for copying file when learning, and it supports remote-command to update it at server running state.
## New configuration
```diff
[nfs]
+ max_copy_rate_megabytes = 500
```


src/dist/nfs/nfs_client_impl.cpp
```diff
@@ -217,6 +233,8 @@ void nfs_client_impl::continue_copy()
             zauto_lock l(req->lock);
             const user_request_ptr &ureq = req->file_ctx->user_req;
             if (req->is_valid) {
+                _copy_token_bucket->consumeWithBorrowAndWait(req->size);
```
levy5307, line 236: Will `req->size` greatter than burst size?

src/dist/nfs/nfs_client_impl.cpp
```diff
@@ -502,5 +520,37 @@ void nfs_client_impl::handle_completion(const user_request_ptr &req, error_code
...
+        "nfs.max_copy_rate [num | DEFAULT]",
+        "control the max rate(MB/s) to copy file from remote node",
+        [this](const std::vector<std::string> &args) {
+            std::string result("OK");
+            if (args.empty()) {
+                result = std::to_string(_opts.max_copy_rate);
+            } else {
+                if (args[0] == "DEFAULT") {
+                    uint32_t burst_size = 1.5 * _opts.max_copy_rate * 1e6;
+                    _copy_token_bucket.reset(
```
levy5307, line 537: Whether _copy_token_bucket is thread safe or not？

src/dist/nfs/nfs_client_impl.cpp
```diff
@@ -502,5 +520,37 @@ void nfs_client_impl::handle_completion(const user_request_ptr &req, error_code
...
+        {"nfs.max_copy_rate"},
+        "nfs.max_copy_rate [num | DEFAULT]",
+        "control the max rate(MB/s) to copy file from remote node",
+        [this](const std::vector<std::string> &args) {
+            std::string result("OK");
+            if (args.empty()) {
+                result = std::to_string(_opts.max_copy_rate);
+            } else {
+                if (args[0] == "DEFAULT") {
+                    uint32_t burst_size = 1.5 * _opts.max_copy_rate * 1e6;
```
levy5307, line 536: `(1.5 * _opts.max_copy_rate)  << 20`

src/dist/nfs/nfs_client_impl.h
```diff
@@ -39,15 +39,22 @@
...
-
+#include <dsn/utility/defer.h>
+#include <dsn/utility/TokenBucket.h>
+#include <dsn/utility/flags.h>
 #include "nfs_client.h"
 
+namespace folly {
+using TokenBucket = BasicTokenBucket<std::chrono::steady_clock>;
+}
+
```
acelyc111, line 50:
```suggestion
using TokenBucket = folly::BasicTokenBucket<std::chrono::steady_clock>;
```
And you can move it to dsn::service namespace

src/dist/nfs/nfs_client_impl.cpp
acelyc111: @Shuo-Jia Use BasicTokenBucket::reset(double genRate, double burstSize, double nowInSeconds) would be better than reset _copy_token_bucket pointer.

src/dist/nfs/nfs_client_impl.cpp
Shuo-Jia: https://github.com/XiaoMi/rdsn/blob/704e1f0b00940cb3418583707154e19f9b2be0e6/src/dist/nfs/nfs_client_impl.cpp#L127 

As above, the requst size <= `nfs_copy_block_bytes`(nfs_copy_block_bytes = 4MB), so it has few influnce

levy5307: So I guess req_size is not definitely smaller than burst size. In this case, `consumeWithBorrowAndWait` will return false, So I suggest to add a judgment here.

src/dist/nfs/nfs_client_impl.cpp
```diff
@@ -502,5 +520,35 @@ void nfs_client_impl::handle_completion(const user_request_ptr &req, error_code
...
+{
+    _ctrl_nfs_max_copy_rate = dsn::command_manager::instance().register_app_command(
+        {"nfs.max_copy_rate"},
+        "nfs.max_copy_rate [num | DEFAULT]",
+        "control the max rate(MB/s) to copy file from remote node",
+        [this](const std::vector<std::string> &args) {
+            std::string result("OK");
+            if (args.empty()) {
+                result = std::to_string(_opts.max_copy_rate);
+            } else {
```
acelyc111, line 534: These if..else.. can be improved.

src/dist/nfs/nfs_client_impl.cpp
```diff
@@ -502,5 +520,35 @@ void nfs_client_impl::handle_completion(const user_request_ptr &req, error_code
...
+                if (args[0] == "DEFAULT") {
+                    uint32_t burst_size = 1.5 * (_opts.max_copy_rate << 20);
+                    _copy_token_bucket.get()->reset(_opts.max_copy_rate << 20, burst_size);
```
acelyc111, line 537:
```suggestion
                    _copy_token_bucket->reset(_opts.max_copy_rate << 20, burst_size);
```

src/dist/nfs/nfs_client_impl.cpp
```diff
@@ -502,5 +520,35 @@ void nfs_client_impl::handle_completion(const user_request_ptr &req, error_code
...
+                } else {
+                    int32_t max_copy_rate = 0;
+                    if (!dsn::buf2int32(args[0], max_copy_rate) || max_copy_rate < 0) {
+                        result = std::string("ERR: invalid arguments");
+                    } else {
+                        uint32_t burst_size = 1.5 * (max_copy_rate << 20);
+                        _copy_token_bucket.get()->reset(_opts.max_copy_rate << 20, burst_size);
```
acelyc111, line 544:
```suggestion
                        _copy_token_bucket->reset(_opts.max_copy_rate << 20, burst_size);
```

src/dist/nfs/nfs_client_impl.cpp
```diff
@@ -65,8 +68,20 @@ nfs_client_impl::nfs_client_impl(nfs_opts &opts)
...
+    // size once
+    dassert(max_copy_rate_bytes > _opts.nfs_copy_block_bytes,
+            "max_copy_rate_bytes should be greater than nfs_copy_block_bytes");
+    _copy_token_bucket.reset(new TokenBucket(max_copy_rate_bytes, 1.5 * max_copy_rate_bytes));
+    _opts.max_copy_rate_megabytes = FLAGS_max_copy_rate_megabytes;
+
+    register_cli_commands();
 }
 
+nfs_client_impl::~nfs_client_impl() { _tracker.cancel_outstanding_tasks(); }
```
acelyc111, line 83: Didn't unregister_app_command, have you run unittests in ASAN mode?

src/dist/nfs/nfs_client_impl.cpp
```diff
@@ -34,11 +34,14 @@
  */
 #include <dsn/utility/filesystem.h>
 #include <queue>
+#include <dsn/tool-api/command_manager.h>
 #include "nfs_client_impl.h"
 
 namespace dsn {
 namespace service {
 
+DSN_DEFINE_int32("nfs", max_copy_rate_megabytes, 500, "max rate of copying from remote node(MB/s)");
```
acelyc111, line 43: How about simplify to `max_copy_mbps`?

Shuo-Jia: `mbps` means `million bits per second`, here is `MB`

src/dist/nfs/nfs_client_impl.cpp
Shuo-Jia: `unregister_app_command` will cause `heap use after free` with ASAN.

acelyc111: Have you checked why `heap use after free` reported?

Shuo-Jia: Yeah, it because the `instance` will be re-freed in https://github.com/XiaoMi/rdsn/blob/30362560829fcc819bd37d9b0f437edb6c8fdf76/src/core/core/command_manager.cpp#L191-L196

I haven't find the root cause why `replica_server` and `meta_server` need be free manually, in `dsn core` the remote commands are no need , for examle:
https://github.com/XiaoMi/rdsn/blob/30362560829fcc819bd37d9b0f437edb6c8fdf76/src/core/perf_counter/perf_counters.cpp#L49-L55
The `perf-counters` just register but no `deregister` and depends:
https://github.com/XiaoMi/rdsn/blob/30362560829fcc819bd37d9b0f437edb6c8fdf76/src/core/core/command_manager.cpp#L191-L196
freeing it.

acelyc111: I found the other calls of `dsn::command_manager::instance().deregister_command(handle)`, a `handle = nullptr` is followed, then `delete c` in command_manager's deconstructor is safe. I doubt whether `handle` has been correctly removed from command_manager when call `deregister_command`.

Shuo-Jia: > a `handle = nullptr` is followed, then `delete c` in command_manager's deconstructor is safe

This is in `replica_server` or `meta_server` is safe, but in here not. `deregister_command` execute before `command_manager's deconstructor`, the crash happens at `command_manager's deconstructor` actually.

PR #463, opened 2020-05-15
feat(bulk-load): meta handle bulk_load_response

hycdong:
The whole bulk load process is like:
1. client sends request to meta server to start bulk load
2. meta sends bulk_load_request to primary replicas
3. primary broadcasts group_bulk_load_request, collects bulk load states from secondary, and reports group bulk load states to meta
4. meta receive bulk_load_response, if bulk load not finished, jump to step2, resend bulk_load_request

This pull request is about meta receive bulk_load_response from primary, the step4 above.
- if rpc meet error or resp.err is OBJECT_NOT_FOUND and INVALID_STATE, meta considers that bulk load meets some recoverable errors and tries to rollback to downloading and resend bulk load request
- each node will limit bulk load downloading count, if resp.err is BUSY, it means that the node has already have enough replica executing downloading, there is no need to do any error handling,  just resend bulk load request
- if other errors happens, such as ERR_CORRUPT, meta will consider bulk load failed and resend bulk load request to replica server to cleanup bulk load states and context
- if resp.err is ok and pass ballot check, meta will call different functions according to current app's bulk_load_status. If bulk load not finished, meta will resend bulk_load_request. Here is an exception, when bulk load status is paused, it means bulk load is paused, meta should not send request to replica to collect states, but bulk load is not finished yet.

There are two different interval to resend bulk load request to primary. Default value is 10 seconds, short interval is 5 seconds. Short interval will be used when app status is downloaded and ingesting (partition status may be ingesting). If partition is ingesting, meta server wishes to get states faster.

This pull request adds lots of function declarations, and the unit tests will be implemented in further pull request.

### DIFF
--- a/src/dist/replication/meta_server/meta_bulk_load_service.cpp
+++ b/src/dist/replication/meta_server/meta_bulk_load_service.cpp
@@ -341,7 +341,7 @@ void bulk_load_service::on_partition_bulk_load_reply(error_code err,
     }
 
     if (response.err == ERR_BUSY) {
-        dwarn_f("node({}) has enough replicas downloading, wait to next round to send bulk load "
+        dwarn_f("node({}) has enough replicas downloading, wait for next round to send bulk load "
                 "request for app({}), partition({})",
                 primary_addr.to_string(),
                 app_name,
--- a/src/dist/replication/meta_server/meta_bulk_load_service.cpp
+++ b/src/dist/replication/meta_server/meta_bulk_load_service.cpp
@@ -316,29 +316,41 @@ void bulk_load_service::on_partition_bulk_load_reply(error_code err,
     const std::string &app_name = request.app_name;
     const gpid &pid = request.pid;
     const rpc_address &primary_addr = request.primary_addr;
-    int32_t interval_ms = _meta_svc->get_options().partition_bulk_load_interval_ms;
+    int32_t interval = bulk_load_constant::BULK_LOAD_REQUEST_INTERVAL;
 
     if (err != ERR_OK) {
         derror_f("app({}), partition({}) failed to recevie bulk load response, error = {}",
                  pid.get_app_id(),
                  pid,
                  err.to_string());
-        try_rollback_to_downloading(pid.get_app_id(), app_name);
-    } else if (response.err == ERR_OBJECT_NOT_FOUND || response.err == ERR_INVALID_STATE) {
+        try_rollback_to_downloading(app_name, pid);
+        try_resend_bulk_load_request(app_name, pid, interval);
+        return;
+    }
+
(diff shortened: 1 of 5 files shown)


src/dist/replication/meta_server/meta_bulk_load_service.cpp
```diff
@@ -312,8 +312,184 @@ void bulk_load_service::partition_bulk_load(const std::string &app_name, const g
...
+                 err.to_string());
+        try_rollback_to_downloading(app_name, pid);
+        try_resend_bulk_load_request(app_name, pid, interval);
+        return;
+    }
+
+    if (response.err == ERR_OBJECT_NOT_FOUND || response.err == ERR_INVALID_STATE) {
+        derror_f(
+            "app({}), partition({}) doesn't exist or has invalid state on node({}), error = {}",
+            app_name,
```
acelyc111, line 334: Why use different method to log app with L323?

src/dist/replication/meta_server/meta_bulk_load_service.cpp
```diff
@@ -312,8 +312,184 @@ void bulk_load_service::partition_bulk_load(const std::string &app_name, const g
...
+    const rpc_address &primary_addr = request.primary_addr;
+    int32_t interval = bulk_load_constant::BULK_LOAD_REQUEST_INTERVAL;
+
+    if (err != ERR_OK) {
+        derror_f("app({}), partition({}) failed to recevie bulk load response, error = {}",
+                 pid.get_app_id(),
+                 pid,
+                 err.to_string());
+        try_rollback_to_downloading(app_name, pid);
+        try_resend_bulk_load_request(app_name, pid, interval);
```
acelyc111, line 327: How to deal with a nonrecoverable error, e.g. files in remote storage is conrrupt?

hycdong: On L353 ~ L363, the app and partitions' bulk load status will be set as bulk_load_failed (will be implemented in function`handle_bulk_load_failed`), and resend request to replica server to cleanup bulk load states and context.

src/dist/replication/meta_server/meta_bulk_load_service.cpp
```diff
@@ -312,8 +312,184 @@ void bulk_load_service::partition_bulk_load(const std::string &app_name, const g
...
+            pid,
+            primary_addr.to_string(),
+            response.err.to_string());
+        try_rollback_to_downloading(app_name, pid);
+        try_resend_bulk_load_request(app_name, pid, interval);
+        return;
+    }
+
+    if (response.err == ERR_BUSY) {
+        dwarn_f("node({}) has enough replicas downloading, wait to next round to send bulk load "
```
levy5307, line 344: wait for?

PR #460, opened 2020-05-13
feat(bulk-load): group bulk load request in replica group

hycdong:
This pull request is about primary broadcast group_bulk_load_request to secondaries.
- primary replica receive bulk_load_request (`on_bulk_load` #457)
- primary replica broadcast group_bulk_load_request (`broadcast_group_bulk_load`)
- secondary replica calls `on_group_bulk_load` when receive group_bulk_load_request
  - validate replica status and ballot
  - execute actual bulk load process (do_bulk_load - not implemented in this pr )
  - report bulk load states to primary (report_bulk_load_states_to_primary - not implemented in this pr)
- primary replica calls `on_group_bulk_load_reply` when receive group_bulk_load_response, part-implemented in this pr

Whole bulk load process document can be found in official website: [bulk_load_design](https://pegasus-kv.github.io/2020/02/18/bulk-load-design.html)

PR #457, opened 2020-05-09
feat(bulk-load): meta server send bulk load request

hycdong:
This pull request is about meta server send bulk_load_request to replica server.
- meta server creates bulk_load_rpc and send bulk_load_request in function `partition_bulk_load`
- primary replica calls `on_bulk_load` when receive bulk_load_request
  1. validate replica status and ballot
  2. execute actual bulk load process (`do_bulk_load` - not implemented in this pr )
  3. broadcast bulk load request to secondaries and report bulk load states to meta (`broadcast_group_bulk_load`, `report_bulk_load_states_to_meta` - not implemented in this pr)
- meta server calls `on_partition_bulk_load_reply` when receive bulk_load_response, also not implemented in this pr.
Whole bulk load process document can be found in official website: [bulk_load_design](https://pegasus-kv.github.io/2020/02/18/bulk-load-design.html)

PR #454, opened 2020-05-06
feat(bulk-load): meta server start bulk load

hycdong:
This pull request is about meta server received client request and started bulk load.

Client sends `start_bulk_load_request` to meta server, meta server will try to start bulk load:
- check if parameters  are valid, see more in function `check_bulk_load_request_params`
- create app bulk load directory on remote storage and set app bulk_load_status as downloading, in function `create_app_bulk_load_dir`
- create partition bulk load directory on remote storage and set partition bulk_load_status as downloading, in function `create_partition_bulk_load_dir`
- initialize some varieties useful during bulk load process

All steps above finished, meta server will return ERR_OK to client, and start to do bulk load asynchronously, which will be implemented in further pull request.

Bulk load design document can be found in official website: [bulk_load_design](https://pegasus-kv.github.io/2020/02/18/bulk-load-design.html)

PR #465, opened 2020-05-18
feat(bulk-load): bulk load download part1 - replica start download

hycdong:
The whole bulk load download process is like:
1. meta server set app and all partitions bulk load status is downloading #454 
2. meta server send request to primary to download files #457 
3. **primary download files**
5. **secondary download files**
6. secondary report download progress and status to primary
7. primary report group download progress and status to meta server
8. meta receive download progress
    - if all replicas of the group finish download all files, meta set this partition status as downloaded, and if all partitions are downloaded, app status will be set as downloaded, otherwise, meta will resend request to primary.

**This pull request is part one of replica download files**, step 3 and step 5 above. 
- add an option called `max_concurrent_bulk_load_downloading_count` to restrict the replica count executing bulk load downloading on the node (replica_stub), and this option call be configured by remote command.
- `bulk_load_start_download`: if current downloading count does not exceed restriction, clear context and call `download_sst_files`
    - `primary_context::cleanup_bulk_load_states` will cleanup bulk load states reported from secondaries
    - `replica::clear_bulk_load_states` will cleanup replica's bulk load states and replica's bulk load context
- `download_sst_files`:
    - create local directory to store sst files temporarily
    - download metadata file synchronously (not implemented)
    - parse metadata structure from file (not implemented)
    - download sst files asynchronously (not implemented)
    - if download files succeed, verify sst files' metadata and update download progress (not implemented)
    - if download files failed or not pass verification, store error in `_download_status`, which will be reported to primary and meta server.

Function `download_sst_files` won't have unit test, but functions called by it such as `do_download`, `parse_bulk_load_metadata` have unit tests, will be added in further pull request.


src/dist/replication/common/replication_common.cpp
```diff
@@ -518,6 +520,11 @@ void replication_options::initialize()
                                                           "bulk_load_provider_root",
                                                           "bulk_load_provider_root",
                                                           "bulk load root on remote file provider");
+    max_concurrent_bulk_load_downloading_count =
```
acelyc111, line 523: For new added configs, preffer to use the util introduced by https://github.com/XiaoMi/rdsn/pull/275

src/dist/replication/lib/replica_stub.cpp
```diff
@@ -2201,6 +2206,31 @@ void replica_stub::open_service()
...
         });
 #endif
+    _max_concurrent_bulk_load_downloading_count_command =
+        dsn::command_manager::instance().register_app_command(
+            {"max-concurrent-bulk-load-downloading-count"},
+            "max-concurrent-bulk-load-downloading-count [num | DEFAULT]",
+            "control stub max_concurrent_bulk_load_downloading_count",
+            [this](const std::vector<std::string> &args) {
+                std::string result("OK");
+                if (args.empty()) {
```
acelyc111, line 2216: These if..else.. are too complex, better to refactor like https://github.com/XiaoMi/rdsn/pull/461/files#diff-a6778028d5f3348c9fb2e6cbd55891adR525

src/dist/replication/lib/replica_bulk_load.cpp
```diff
@@ -194,6 +194,160 @@ error_code replica::do_bulk_load(const std::string &app_name,
...
+void replica::update_bulk_load_download_progress(uint64_t file_size, const std::string &file_name)
+{
+    // TODO(heyuchen): TBD
+    // update download progress after downloading sst files succeed
+}
+
+// ThreadPool: THREAD_POOL_REPLICATION, THREAD_POOL_REPLICATION_LONG
+void replica::try_decrease_bulk_load_download_count()
+{
+    if (_stub->_bulk_load_downloading_count.load() > 0) {
```
acelyc111, line 340: There maybe a  multithread problem, suppose _bulk_load_downloading_count == 1, and there are 2 threads consult `if (_stub->_bulk_load_downloading_count.load() > 0)`, both of them is true, them both of them fetch_sub(1), the result is -1.

src/dist/replication/lib/replica_bulk_load.cpp
```diff
@@ -194,6 +194,160 @@ error_code replica::do_bulk_load(const std::string &app_name,
...
+                                       const std::string &cluster_name,
+                                       const std::string &provider_name)
+{
+    FAIL_POINT_INJECT_F("replica_bulk_load_download_sst_files",
+                        [](string_view) -> error_code { return ERR_OK; });
+
+    // create local bulk load dir
+    const std::string local_dir =
+        utils::filesystem::path_combine(_dir, bulk_load_constant::BULK_LOAD_LOCAL_ROOT_DIR);
+    if (!utils::filesystem::directory_exists(_dir)) {
```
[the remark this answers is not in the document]
hycdong: _dir is created when the replica is initialize.
https://github.com/XiaoMi/rdsn/blob/466bca2e83557c1a0eb01c573498279ad84bb9ad/src/dist/replication/lib/replica_init.cpp#L40-L56

src/dist/replication/lib/replica_bulk_load.cpp
```diff
@@ -194,6 +194,160 @@ error_code replica::do_bulk_load(const std::string &app_name,
...
+            LPC_BACKGROUND_BULK_LOAD, &_tracker, [this, remote_dir, local_dir, f_meta, fs]() {
+                error_code ec = ERR_OK;
+                uint64_t f_size = 0;
+                do_download(remote_dir, local_dir, f_meta.name, fs, ec, f_size);
+                if (ec == ERR_OK && !verify_sst_files(f_meta, local_dir)) {
+                    ec = ERR_CORRUPTION;
+                }
+                if (ec != ERR_OK) {
+                    try_decrease_bulk_load_download_count();
+                    _bulk_load_context._download_status.store(ec);
```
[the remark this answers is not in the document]
hycdong: Its default value is ERR_OK, if download succeed, it will be ERR_OK.

src/dist/replication/lib/replica_bulk_load.cpp
hycdong: I think the result of the multi-thread problem can be accepted. If the result is -1, it may lead actual downloading count = restriction count + 1. However, if you think it is necessary to avoid such problem, I can add a lock for it.

hycdong: Use atomic `++` and `--` instead, and add debug log here

PR #466, opened 2020-05-19
refactor: make aio decoupled from dsn_runtime

levy5307:
make aio a separate library, decoupled from dsn_runtime


src/core/aio/sim_aio_provider.h
```diff
@@ -24,22 +24,11 @@
...
- */
-
 #pragma once
 
-#include <dsn/tool_api.h>
-#include "core/tools/common/native_aio_provider.linux.h"
+#include "native_linux_aio_provider.h"
 
 namespace dsn {
-namespace tools {
```
neverchanje, line 42: I think a 'dsn::aio::' namespace is better.

levy5307: Maybe we can do it in later pull request. Because too many files are involved if we change namespace `dsn::` to `dsn::aio::`. And it is not the purpose of this pull request

src/core/aio/native_linux_aio_provider.cpp
```diff
@@ -24,22 +24,26 @@
...
  */
 
-#include "native_aio_provider.linux.h"
+#include "native_linux_aio_provider.h"
 
 #include <fcntl.h>
 #include <cstdlib>
 
 namespace dsn {
-namespace tools {
```
neverchanje, line 33: replace it with aio:: namespace

```suggestion
namespace aio {
```

src/core/aio/disk_engine.h
```diff
@@ -68,14 +67,12 @@ class disk_file
     work_queue<aio_task> _read_queue;
 };
 
-class disk_engine
+class disk_engine : public utils::singleton<disk_engine>
 {
 public:
-    disk_engine(service_node *node);
+    disk_engine();
```
acelyc111, line 73: better to move ctor and dctor to private section, to restrict create and delete it freely.

levy5307: We can't make ctor and dctor to be private member, because we derived from `singleton`, to implement sigleton. And disk_engine is created in `singleton::instance`

PR #467, opened 2020-05-20
refactor(bulk-load): add replica_bulk_loader class

hycdong:
This pull request refactors replica bulk load functions.
- add `replica_bulk_loader` class to control bulk load functions
- remove bulk_load_context class and move those varieties into replica_bulk_loader class
- add `replica_file_provider.cpp` file to implement `do_download` function, this function will also be used during restore. We will also move other file provider functions into this file after we refactor cold backup in the future.

### DIFF
--- a/src/dist/replication/lib/bulk_load/replica_bulk_load.h
+++ b/src/dist/replication/lib/bulk_load/replica_bulk_load.h
@@ -1,0 +1,112 @@
+// Copyright (c) 2017-present, Xiaomi, Inc.  All rights reserved.
+// This source code is licensed under the Apache License Version 2.0, which
+// can be found in the LICENSE file in the root directory of this source tree.
+
+#pragma once
+
+#include <dist/replication/lib/replica.h>
+#include <dist/replication/lib/replica_context.h>
+#include <dist/replication/lib/replica_stub.h>
+
+namespace dsn {
+namespace replication {
+
+class replica_bulk_load : replica_base
+{
+public:
+    explicit replica_bulk_load(replica *r);
+    ~replica_bulk_load();
+
+    void on_bulk_load(const bulk_load_request &request, /*out*/ bulk_load_response &response);
+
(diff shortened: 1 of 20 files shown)


src/dist/replication/lib/replica_file_provider.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+#include <dsn/dist/block_service.h>
+#include <dsn/dist/fmt_logging.h>
+#include <dsn/utility/filesystem.h>
+
+namespace dsn {
+namespace replication {
+
+// ThreadPool: THREAD_POOL_REPLICATION, THREAD_POOL_REPLICATION_LONG
+error_code replica::do_download(const std::string &remote_dir,
```
neverchanje, line 17: Why still make do_download in the replica class. Will it be refactored in next PR?

src/dist/replication/lib/bulk_load/replica_bulk_load.h
```diff
@@ -0,0 +1,112 @@
...
+#pragma once
+
+#include <dist/replication/lib/replica.h>
+#include <dist/replication/lib/replica_context.h>
+#include <dist/replication/lib/replica_stub.h>
+
+namespace dsn {
+namespace replication {
+
+class replica_bulk_load : replica_base
```
neverchanje, line 14: Better to call it replica_bulk_loader.

src/dist/replication/lib/replica_file_provider.cpp
hycdong: `do_download` will also be used in restore, it should not be added into bulk load class, it should be added into a class used to operate remote file provider functions. However, backup and restore is still refactoring. This function is left in replica temporarily.

neverchanje: Ok. Let's see how it's implemented, and consider how to refactor it.

PR #472, opened 2020-05-23
fix(dup): reject add_dup if remote address incorrect

neverchanje:
This fix is to ensure the robustness when the remote cluster address of `add_dup` was misconfigured.


src/dist/replication/test/meta_test/unit_test/config-test.ini
```diff
@@ -111,3 +111,8 @@ args =
 [duplication-group]
 master-cluster = 1
 slave-cluster  = 2
+cluster_without_address = 3
```
acelyc111, line 114: `cluster_without_address ` is only for test, right? Better to rename to  `cluster_without_address_for_test ` or some other like that to clarify it.

PR #477, opened 2020-05-26
fix: fix memory leak in asio_udp_provider

levy5307:
In rdsn, if you use UDP to send beacon between meta server and replica server, the memory usage will continue to grow. It is because the response is not released after it is sent by asio_udp_provider.

For the callers of `dsn_rpc_reply`, they think the response is released by `dsn_rpc_reply` itself by default. So they shouldn't need to release the response. For example: `serverlet<T>::reply`, `rpc_holder::reply`. They just create a response and pass it to `dsn_rpc_reply`.

```cpp
DSN_API void dsn_rpc_reply(dsn::message_ex *response, dsn::error_code err)
{
    auto msg = ((::dsn::message_ex *)response);
    ::dsn::task::get_current_rpc()->reply(msg, err);
}
```

Then `dsn_rpc_reply` simply passes the response to `rpc_engine::reply`, and let it release the response.
Let's have a look at `rpc_engine::reply`:

```cpp
void rpc_engine::reply(message_ex *response, error_code err)
{
    // when a message doesn't need to reply, we don't do the on_rpc_reply hooks to avoid mistakes
    // for example, the profiler may be mistakenly calculated
    auto s = response->io_session.get();
    if (s == nullptr && response->to_address.is_invalid()) {
        dinfo("rpc reply %s is dropped (invalid to-address), trace_id = %016" PRIx64,
              response->header->rpc_name,
              response->header->trace_id);
        response->add_ref();
        response->release_ref();
        return;
    }

    strncpy(response->header->server.error_name,
            err.to_string(),
            sizeof(response->header->server.error_name) - 1);
    response->header->server.error_name[sizeof(response->header->server.error_name) - 1] = '\0';
    response->header->server.error_code.local_code = err;
    response->header->server.error_code.local_hash = message_ex::s_local_hash;

    // response rpc code may be TASK_CODE_INVALID when request rpc code is not exist
    auto sp = response->local_rpc_code == TASK_CODE_INVALID
                  ? nullptr
                  : task_spec::get(response->local_rpc_code);

    bool no_fail = true;
    if (sp) {
        // current task may be nullptr when this method is directly invoked from rpc_engine.
        task *cur_task = task::get_current_task();
        if (cur_task) {
            no_fail = sp->on_rpc_reply.execute(cur_task, response, true);
        }
    }

    // connection oriented network, we have bound session
    // for tcp, use rpc_session::send_message to send response
    if (s != nullptr) {
        // not forwarded, we can use the original rpc session
        if (!response->header->context.u.is_forwarded) {
            if (no_fail) {
                s->send_message(response);
            } else {
                s->net().inject_drop_message(response, true);
            }
        }

        // request is forwarded, we cannot use the original rpc session,
        // so use client session to send response.
        else {
            dbg_dassert(response->to_address.port() > MAX_CLIENT_PORT,
                        "target address must have named port in this case");

            // use the header format recorded in the message
            auto rpc_channel = sp ? sp->rpc_call_channel : RPC_CHANNEL_TCP;
            network *net = _client_nets[response->hdr_format][rpc_channel].get();
            dassert(
                nullptr != net,
                "client network not present for rpc channel '%s' with format '%s' used by rpc %s",
                RPC_CHANNEL_TCP.to_string(),
                response->hdr_format.to_string(),
                response->header->rpc_name);

            if (no_fail) {
                net->send_message(response);
            } else {
                net->inject_drop_message(response, true);
            }
        }
    }

    // not connection oriented network, we always use the named network to send msgs
    else {
        dbg_dassert(response->to_address.port() > MAX_CLIENT_PORT,
                    "target address must have named port in this case");

        auto rpc_channel = sp ? sp->rpc_call_channel : RPC_CHANNEL_TCP;
        network *net = _server_nets[response->header->from_address.port()][rpc_channel].get();

        dassert(nullptr != net,
                "server network not present for rpc channel '%s' on port %u used by rpc %s",
                RPC_CHANNEL_UDP.to_string(),
                response->header->from_address.port(),
                response->header->rpc_name);

        if (no_fail) {
               /// for udp, use asio_udp_provider::send_message to send response
            net->send_message(response);
        } else {
            net->inject_drop_message(response, true);
        }
    }

    if (!no_fail) {
        // because (1) initially, the ref count is zero
        //         (2) upper apps may call add_ref already
        response->add_ref();
        response->release_ref();
    }
}
```

if we use udp, the response is passed to `asio_udp_provider::send_message`, and it is not released by `rpc_engine::reply`. so we should release it in `asio_udp_provider::send_message`, just like what `rpc_session::send_message` does when we use tcp.

But now `asio_udp_provider::send_message` doesn't do any release work, so we should add these work to it.

### DIFF
--- a/src/core/tools/common/asio_net_provider.cpp
+++ b/src/core/tools/common/asio_net_provider.cpp
@@ -208,6 +208,8 @@ void asio_udp_provider::send_message(message_ex *request)
                 // we do not handle failure here, rpc matcher would handle timeouts
             }
         });
+    request->add_ref();
+    request->release_ref();
 }
 
 asio_udp_provider::asio_udp_provider(rpc_engine *srv, network *inner_provider)


acelyc111:
Could you please add some unit tests, then we can detect it on ASAN mode.

acelyc111:
Add some desciption for this PR.

levy5307:
> Could you please add some unit tests, then we can detect it on ASAN mode.

Sure

PR #478, opened 2020-05-27
fix: fix memory leak in unit test netprovider.cpp

levy5307:
Use std::unique_ptr instead of raw pointer.


src/core/tests/netprovider.cpp
```diff
@@ -148,6 +149,7 @@ TEST(tools_common, asio_net_provider)
 
     asio_network_provider *asio_network =
         new asio_network_provider(task::get_current_rpc(), nullptr);
+    auto cleanup = dsn::defer([=]() { delete asio_network; });
```
acelyc111, line 152: Why not use smart pointer simply?

levy5307: good idea

PR #475, opened 2020-05-25
feat(bulk-load): bulk load download part3 - replica parse metadata, verify files and update progress

hycdong:
This pull request is about what will replica do after downloading metadata file and sst files.
- if download metadata succeed, replica will parse metadata from file to local `_metadata` (in function `parse_bulk_load_metadata`)
- if download sst file succeed, replica will compare sst file size and md5 with the value parsed from metadata (in function `verify_file`)
- if pass verification, replica will update download progress (in function `update_bulk_load_download_progress`)
- if current replica download all sst files, will turn bulk load status to downloaded (in function `check_download_finish`)

PR #471, opened 2020-05-21
feat(bulk-load): bulk load download part2 - replica download files

hycdong:
This pull request add `do_download` function and its unit test. This function is used to download files from remote provider.

PR #479, opened 2020-05-28
feat(bulk-load): bulk load download part4 - replica report download status and progress

hycdong:
The whole bulk load download process is like:
1. meta server set app and all partitions bulk load status is downloading #454
2. meta server send request to primary to download files #457
3. primary download files #465 #471 #475 
5. secondary download files #465 #471 #475 
6. **secondary report download progress and status to primary**
7. **primary report group download progress and status to meta server**
8. meta receive download progress
    - if all replicas of the group finish download all files, meta set this partition status as downloaded, and if all partitions are downloaded, app status will be set as downloaded, otherwise, meta will resend request to primary

This pull request is about replica report download status and progress, step 6 and step 7 above.

### DIFF
--- a/src/dist/replication/lib/bulk_load/replica_bulk_loader.cpp
+++ b/src/dist/replication/lib/bulk_load/replica_bulk_loader.cpp
@@ -431,7 +431,7 @@ void replica_bulk_loader::report_bulk_load_states_to_meta(bulk_load_status::type
         return;
     }
 
-    if (report_metadata && _metadata.files.size() > 0) {
+    if (report_metadata && !_metadata.files.empty()) {
         response.__set_metadata(_metadata);
     }
 
@@ -459,25 +459,28 @@ void replica_bulk_loader::report_group_download_progress(/*out*/ bulk_load_respo
         return;
     }
 
-    partition_bulk_load_state p_state;
-    p_state.__set_download_progress(_download_progress.load());
-    p_state.__set_download_status(_download_status.load());
-    response.group_bulk_load_state[_replica->_primary_states.membership.primary] = p_state;
+    partition_bulk_load_state primary_state;
+    primary_state.__set_download_progress(_download_progress.load());
+    primary_state.__set_download_status(_download_status.load());
+    response.group_bulk_load_state[_replica->_primary_states.membership.primary] = primary_state;
     ddebug_replica("primary = {}, download progress = {}%, status = {}",
                    _replica->_primary_states.membership.primary.to_string(),
-                   p_state.download_progress,
-                   p_state.download_status);
+                   primary_state.download_progress,
+                   primary_state.download_status);
 
-    int32_t total_progress = p_state.download_progress;
+    int32_t total_progress = primary_state.download_progress;
     for (const auto &target_address : _replica->_primary_states.membership.secondaries) {
-        const auto &s_state = _replica->_primary_states.secondary_bulk_load_states[target_address];
-        int32_t s_progress = s_state.__isset.download_progress ? s_state.download_progress : 0;
-        error_code s_status = s_state.__isset.download_status ? s_state.download_status : ERR_OK;
+        const auto &secondary_state =
+            _replica->_primary_states.secondary_bulk_load_states[target_address];
+        int32_t s_progress =
+            secondary_state.__isset.download_progress ? secondary_state.download_progress : 0;
+        error_code s_status =
+            secondary_state.__isset.download_status ? secondary_state.download_status : ERR_OK;
         ddebug_replica("secondary = {}, download progress = {}%, status={}",
                        target_address.to_string(),
                        s_progress,
                        s_status);
-        response.group_bulk_load_state[target_address] = s_state;
+        response.group_bulk_load_state[target_address] = secondary_state;
         total_progress += s_progress;
     }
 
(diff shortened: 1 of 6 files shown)


src/dist/replication/lib/bulk_load/replica_bulk_loader.cpp
```diff
@@ -426,14 +426,91 @@ void replica_bulk_loader::report_bulk_load_states_to_meta(bulk_load_status::type
...
+    response.group_bulk_load_state[_replica->_primary_states.membership.primary] = p_state;
+    ddebug_replica("primary = {}, download progress = {}%, status = {}",
+                   _replica->_primary_states.membership.primary.to_string(),
+                   p_state.download_progress,
+                   p_state.download_status);
+
+    int32_t total_progress = p_state.download_progress;
+    for (const auto &target_address : _replica->_primary_states.membership.secondaries) {
+        const auto &s_state = _replica->_primary_states.secondary_bulk_load_states[target_address];
+        int32_t s_progress = s_state.__isset.download_progress ? s_state.download_progress : 0;
```
levy5307, line 474: `int32_t s_progress = s_state.download_progress;`
`download_progress` is 0 by default

hycdong: I think it's okay to write like this, it's more clearer.

src/dist/replication/lib/bulk_load/replica_bulk_loader.cpp
```diff
@@ -426,14 +426,91 @@ void replica_bulk_loader::report_bulk_load_states_to_meta(bulk_load_status::type
                                                           bool report_metadata,
                                                           /*out*/ bulk_load_response &response)
 {
-    // TODO(heyuchen): TBD
+    if (status() != partition_status::PS_PRIMARY) {
+        response.err = ERR_INVALID_STATE;
+        return;
+    }
+
+    if (report_metadata && _metadata.files.size() > 0) {
```
acelyc111, line 434:
```suggestion
    if (report_metadata && !_metadata.files.empty()) {
```

src/dist/replication/lib/bulk_load/replica_bulk_loader.cpp
```diff
@@ -426,14 +426,91 @@ void replica_bulk_loader::report_bulk_load_states_to_meta(bulk_load_status::type
...
+    p_state.__set_download_status(_download_status.load());
+    response.group_bulk_load_state[_replica->_primary_states.membership.primary] = p_state;
+    ddebug_replica("primary = {}, download progress = {}%, status = {}",
+                   _replica->_primary_states.membership.primary.to_string(),
+                   p_state.download_progress,
+                   p_state.download_status);
+
+    int32_t total_progress = p_state.download_progress;
+    for (const auto &target_address : _replica->_primary_states.membership.secondaries) {
+        const auto &s_state = _replica->_primary_states.secondary_bulk_load_states[target_address];
```
acelyc111, line 473: `p_` or `s_` prefix is confused, use a full word is ok, it's not too long.

src/dist/replication/lib/bulk_load/replica_bulk_loader.cpp
```diff
@@ -426,14 +426,91 @@ void replica_bulk_loader::report_bulk_load_states_to_meta(bulk_load_status::type
...
+        ddebug_replica("secondary = {}, download progress = {}%, status={}",
+                       target_address.to_string(),
+                       s_progress,
+                       s_status);
+        response.group_bulk_load_state[target_address] = s_state;
+        total_progress += s_progress;
+    }
+
+    total_progress /= _replica->_primary_states.membership.max_replica_count;
+    ddebug_replica("total download progress = {}%", total_progress);
```
acelyc111, line 485: How about use float type?

src/dist/replication/lib/bulk_load/replica_bulk_loader.cpp
```diff
@@ -426,14 +426,91 @@ void replica_bulk_loader::report_bulk_load_states_to_meta(bulk_load_status::type
...
+        return;
+    }
+
+    partition_bulk_load_state bulk_load_state;
+    auto local_status = _status;
+    switch (remote_status) {
+    case bulk_load_status::BLS_DOWNLOADING:
+    case bulk_load_status::BLS_DOWNLOADED:
+        bulk_load_state.__set_download_progress(_download_progress.load());
+        bulk_load_state.__set_download_status(_download_status.load());
```
acelyc111, line 505: We always report the two type of status, so why separate to two types?

src/dist/replication/lib/bulk_load/replica_bulk_loader.cpp
levy5307: I thought about it and you are right.

src/dist/replication/lib/bulk_load/replica_bulk_loader.cpp
hycdong: `total_progress` will report to meta server in `bulk_load_response`, defined in thrift, it seems that we don't use float in thrift before, and I think it is okay to define it as integer.

src/dist/replication/lib/bulk_load/replica_bulk_loader.cpp
hycdong: Do you mean that why not combine download progress and status into one structure? I used to define a structure called `download_states` before, but I finally separate it, it will make code more complex, besides, `download_progress` and `download_status` is more clearer than `download_states`.

PR #480, opened 2020-05-28
refactor: refactor block service manager

hycdong:
This pull request is about refactor block service manager.

`block_service_manager` is used for both replica server and meta server to operate remote file provider. `replica_stub` and `meta_service` hold `_block_service_manager`. When server would like to operate remote file provider, it would use `get_block_filesystem(provider_name)` to get the specific block_filesystem pointer, then do exact operations such as read metadata , upload or download files.`block_service_manager` is declared in replication namespace, in directory `replication/common`, which is not suitable. As a result, this pull request focus on refactor it, and move it into block_service namespace.

The main updates for this pull request:
- move `block_service_manager` class from `replication/common` to `block_service`
- move `do_download` from `replica.h` to `block_service_manager.h` and rename it to `download_file` and also move unit tests
- move `verify_file` from `replica_bulk_loader.h` to `block_service_manager.h` and also move unit tests
- refactor replica unit test
    - move `block_service_mock` from replica unit test to block_service unit test and remove code specific to backup, and add `backup_block_service_mock`
    - remove some useless code and config


src/dist/replication/test/replica_test/unit_test/backup_block_service_mock.h
```diff
@@ -0,0 +1,55 @@
+#pragma once
```
neverchanje, line 1: add license

```

// Copyright (c) 2017-present, Xiaomi, Inc.  All rights reserved.
// This source code is licensed under the Apache License Version 2.0, which
// can be found in the LICENSE file in the root directory of this source tree.
```

PR #483, opened 2020-05-30
build: add rocksdb as thirdparty

neverchanje:
RocksDB was maintained as a stable submodule with no frequent modification. So we make it a thirdparty now. Then we can build Pegasus faster if the thirdparties all prebuilt.

### DIFF
--- a/thirdparty/download-thirdparty.sh
+++ b/thirdparty/download-thirdparty.sh
@@ -309,12 +309,12 @@ check_and_download "${S2GEOMETRY_PKG}" \
     "${S2GEOMETRY_NAME}"
 
 # rocksdb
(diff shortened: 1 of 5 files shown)


thirdparty/download-thirdparty.sh
```diff
@@ -308,6 +308,15 @@ check_and_download "${S2GEOMETRY_PKG}" \
     "bfa5f1c08f535a72fb2c92ec16332c64" \
     "${S2GEOMETRY_NAME}"
 
+# rocksdb
+# from: https://github.com/google/s2geometry/archive/0239455c1e260d6d2c843649385b4fb9f5b28dba.zip
```
acelyc111, line 312: Need update.

PR #486, opened 2020-06-03
feat(bulk-load): bulk load ingestion part1 - meta start ingestion

hycdong:
The whole bulk load ingestion process is like:
1. **meta server set app and all partitions bulk load status as ingesting**
2. meta send two requests to primary 
    - **send ingestion request to primary**
    - send bulk load request to query ingestion status #457
3. primary handle two requests above:
    - primary consider ingestion request as a client write and replica group execute 2pc process to do actual rocksdb ingestion operation
    - primary handle bulk load request and broadcast it to secondary
4. secondary handle two cases:
    - ingestion files through 2pc
    - report ingestion status to primary 
5. primary report group ingestion status to primary
6. meta handle two response
    - execute error handling from ingestion_response
    - handle group ingestion status and do bulk load status transformation

This pull request is about meta start ingestion, step 1 and ingestion request of step 2.
- update meta bulk load status
    - when bulk load downloading stage finished, each partition bulk load status and app status are downloaded, then update each partition bulk load status from downloaded to ingesting (in `on_partition_bulk_load_reply`)
    - when all partitions updated, update bulk load status from downloaded to ingesting (in `update_partition_status_on_remote_storage_reply`)
    - when app status updated, send ingestion request to each primary (in `update_app_status_on_remote_storage_reply`)
- send ingestion request (in `partition_ingestion`)
    - this function will execute in default thread pool
    - add related thrift structure and task code
    - replica should consider ingestion request as a write request, as a result, meta server should create a request which is like sent from user client. Client requests have a feature that `msg->header.gpid = pid`. 

**_Special explanation and discussion_**:
I don't use `rpc_holder` utility for meta server to send ingestion request to primary, because I need to update gpid filed in message header, and it is a special case for meta to create a client request. If you have any suggestion about it, please comment this pull request or discuss with me offline.

PR #490, opened 2020-06-04
feat(bulk-load): bulk load ingestion part2 - replica handle ingestion request

hycdong:
    - **primary consider ingestion request as a client write and replica group execute 2pc process to do actual rocksdb ingestion operation**
    - primary handle bulk load request and broadcast it to secondary
    - **ingestion files through 2pc**
    - report ingestion status to primary 
5. primary report group ingestion status to primary
6. meta handle two response
    - execute error handling from ingestion_response

This pull request is about replica handle ingestion request, part of step 3 and step 4.
- primary will consider ingestion_request as a special write request, handle it in function `on_client_write`
    - when primary receive bulk load ingestion request, it will reject normal write requests temporarily (by `_is_bulk_load_ingestion`) until all replicas ingestion finished (not in this pull request), because rocksdb will reject write requests during ingest external sst files
 - for normal write requests, it is okay when there are only one secondary alive, and for ingestion request, all primary and secondaries should be alive during 2pc, because replay ingestion mutation can not learn data in sst files. This will be checked in `on_client_write` and `init_prepare`
- update `apply_mutation`: don't execute error handling for ingestion request, ingestion errors will be handled by meta server in function `on_partition_ingestion_reply`
- add `_dir_bulk_load`, `get_ingestion_status`, `set_ingestion_status` declaration for further pull request

**_Special explanation about simple_kv test_**
I define a new thread pool `THREAD_POOL_INGESTION`, it will lead case 402 failed.
This case will inject a disk error to secondary, this pull request won't affect this case, so I update case 402.


src/dist/replication/lib/replica_2pc.cpp
```diff
@@ -130,6 +150,22 @@ void replica::init_prepare(mutation_ptr &mu, bool reconciliation)
...
+    for (auto i = 0; i < request_count; ++i) {
+        const mutation_update &update = mu->data.updates[i];
+        if (update.code == dsn::apps::RPC_RRDB_RRDB_BULK_LOAD) {
+            ddebug_replica("try to prepare bulk load mutation({})", mu->name());
+            if (static_cast<int>(_primary_states.membership.secondaries.size()) + 1 <
+                _primary_states.membership.max_replica_count) {
+                err = ERR_NOT_ENOUGH_MEMBER;
+                break;
+            }
+        }
```
acelyc111, line 163: If one update in updates is not RPC_RRDB_RRDB_BULK_LOAD type, the others should not be RPC_RRDB_RRDB_BULK_LOAD type too? If that, you can break the loop early.

src/dist/replication/lib/replica_2pc.cpp
```diff
@@ -74,6 +74,25 @@ void replica::on_client_write(dsn::message_ex *request, bool ignore_throttling)
         return;
     }
 
+    if (_is_bulk_load_ingestion) {
+        // reject write requests during ingestion
+        response_client_write(request, ERR_BUSY);
```
neverchanje, line 79: Will ERR_BUSY be ambiguous with the rate-limit error of write? How should the client handle this error?

include/dsn/utility/error_code.h
```diff
@@ -119,5 +119,6 @@ DEFINE_ERR_CODE(ERR_APP_DROPPED)
 DEFINE_ERR_CODE(ERR_MOCK_INTERNAL)
 DEFINE_ERR_CODE(ERR_ZOOKEEPER_OPERATION)
 DEFINE_ERR_CODE(ERR_CHILD_REGISTERED)
+DEFINE_ERR_CODE(ERR_INGESTION_FAILED)
```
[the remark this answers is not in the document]
hycdong: ERR_INGESTION_FAILED, will be used in pegasus

include/dsn/dist/replication/replication.codes.h
```diff
@@ -34,6 +34,7 @@ DEFINE_THREAD_POOL_CODE(THREAD_POOL_REPLICATION)
 DEFINE_THREAD_POOL_CODE(THREAD_POOL_LOCAL_APP)
 DEFINE_THREAD_POOL_CODE(THREAD_POOL_REPLICATION_LONG)
 DEFINE_THREAD_POOL_CODE(THREAD_POOL_COMPACT)
+DEFINE_THREAD_POOL_CODE(THREAD_POOL_INGESTION)
```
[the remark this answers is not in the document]
hycdong: Yes, I have considered it, I meant to discuss it in next pull request in pegasus.
When replica is during ingestion, it will reject other write requests. It is better for us to shorten ingestion time. So I add a new thread pool to execute it with 24 worker count (same as replication thread pool). 
Default thread pool are shared with many functions, and some are tasks execute periodically, and worker count is 8. If I reuse it, I will add worker count for this thread pool and test for it.
If adding a new thread pool won't cost much, it is better to add a new one.

src/dist/replication/lib/replica_2pc.cpp
hycdong: It is for compatibility. If I return a new error code like 'ERR_INGESTING', old version client won't recognize it. So I use `ERR_BUSY`, current client version know it.

src/dist/replication/lib/replica_2pc.cpp
hycdong: For RPC_RRDB_RRDB_BULK_LOAD, it can not be batched, so this mutation will only have one update.

acelyc111: I mean if found a non-RPC_RRDB_RRDB_BULK_LOAD 'update' in the loop, we can break the loop to avoid meanless loop (because non-RPC_RRDB_RRDB_BULK_LOAD mutation is the most normal case and it may have multiple 'updates').
 > so this mutation will only have one update.

You can add an `assert` to ensure it.

hycdong: oh, I got your point. I have updated the code. 
If a write operation is defined as batch_not_allowed, such as multi_set, the mutation will only have one client request, and one update in mutation's data field, it is guaranteed by other function. I think it's okay for not adding an assert here.

src/dist/replication/lib/replica_2pc.cpp
levy5307: I think you can use `ERR_IO_PENDING` or someone else. Because `ERR_BUSY` is a common error code, which means rate limit error.

src/dist/replication/lib/replica_2pc.cpp
```diff
@@ -74,6 +74,25 @@ void replica::on_client_write(dsn::message_ex *request, bool ignore_throttling)
...
+        return;
+    }
+
+    if (request->rpc_code() == dsn::apps::RPC_RRDB_RRDB_BULK_LOAD) {
+        ddebug_replica("receive bulk load ingestion request");
+
+        // bulk load ingestion request requires that all secondaries should be alive
+        if (static_cast<int>(_primary_states.membership.secondaries.size()) + 1 <
+            _primary_states.membership.max_replica_count) {
+            response_client_write(request, ERR_NOT_ENOUGH_MEMBER);
```
levy5307, line 89: How about add a log here?

src/dist/replication/lib/replica_2pc.cpp
```diff
@@ -104,6 +123,7 @@ void replica::init_prepare(mutation_ptr &mu, bool reconciliation)
 
     error_code err = ERR_OK;
     uint8_t count = 0;
+    const auto request_count = mu->client_requests.size();
```
levy5307, line 126: I think you can move this line to L153

src/dist/replication/lib/replica_2pc.cpp
hycdong: For bulk load request, response_client_write will response to meta server, meta server will print error log, it is okay that no log here.

src/dist/replication/lib/replica_2pc.cpp
hycdong: There are a goto sentence on L151, varieties should be defined before goto.

src/dist/replication/lib/replica_2pc.cpp
hycdong: I use `ERR_OPERATION_DISABLED` instead, this error code may be more suitable.

include/dsn/dist/replication/replication_app_base.h
```diff
@@ -268,9 +273,10 @@ class replication_app_base : public replica_base
     ::dsn::error_code update_init_info_ballot_and_decree(replica *r);
 
 protected:
-    std::string _dir_data;   // ${replica_dir}/data
-    std::string _dir_learn;  // ${replica_dir}/learn
-    std::string _dir_backup; // ${replica_dir}/backup
+    std::string _dir_data;      // ${replica_dir}/data
+    std::string _dir_learn;     // ${replica_dir}/learn
+    std::string _dir_backup;    // ${replica_dir}/backup
+    std::string _dir_bulk_load; // ${replica_dir}/.bulk_load
```
neverchanje, line 279: It seems we missed this naming problem in our previous review. I recommend not to prefix a dot before the directory. Better to keep the directories naming aligned.

PR #492, opened 2020-06-05
refactor: refactor restore progress

levy5307:
In `block_service_manager.cpp`, there are some functions which can be reused in restore progress. So this pull request use these functions to refactor restore progess.


src/dist/replication/lib/replica_restore.cpp
```diff
@@ -92,223 +93,122 @@ bool replica::read_cold_backup_metadata(const std::string &file,
...
-    std::string local_backup_metada_file =
+
+    // parse cold_backup_meta from metadata file
+    const std::string local_backup_metada_file =
         utils::filesystem::path_combine(local_chkpt_dir, cold_backup_constant::BACKUP_METADATA);
     if (!read_cold_backup_metadata(local_backup_metada_file, backup_metadata)) {
-        derror("%s: recover cold_backup_metadata from file(%s) failed",
-               name(),
-               local_backup_metada_file.c_str());
+        derror_replica("recover cold_backup_metadata from file({}) failed",
```
hycdong, line 181: 'read cold_backup_metadata' may be better

src/dist/replication/lib/replica_restore.cpp
```diff
@@ -424,42 +324,36 @@ dsn::error_code replica::restore_checkpoint()
...
-                err = skip_restore_partition(restore_dir);
-            } else {
-                tell_meta_to_restore_rollback();
-                return ERR_CORRUPTION;
-            }
-        }
-    } else if (err == ERR_OBJECT_NOT_FOUND) { // find valid checkpoint failed
+    }
+
+    if (err == ERR_OBJECT_NOT_FOUND || err == ERR_CORRUPTION) {
```
hycdong, line 350: This condition will handle `find_valid_checkpoint` and `download_checkpoint`'s error `ERR_OBJECT_NOT_FOUND` and `ERR_CORRUPTION`, will those two functions return other error code?

levy5307: yes, if return other error code, the restore_checkpoint will return the error code, and wil retry later. Only these `ERR_OBJECT_NOT_FOUND` and `ERR_CORRUPTION` should two judge whether they allow `skip_bad_partition` or not. if it is not allowed, the restore will rollback.

PR #493, opened 2020-06-05
feat(bulk-load): bulk load ingestion part4 - meta handle ingestion response

hycdong:
    - primary consider ingestion request as a client write and replica group execute 2pc process to do actual rocksdb ingestion operation #490 pegasus pr [#546](https://github.com/XiaoMi/pegasus/pull/546)
    - primary handle bulk load request and broadcast it to secondary
    - ingestion files through 2pc #490 pegasus pr [#546](https://github.com/XiaoMi/pegasus/pull/546)
    - report ingestion status to primary 
5. primary report group ingestion status to primary
6. meta handle two response
    - **handle ingestion_response**

This pull request is about how meta handle ingestion_response, part of step 6, and add some unit tests. Ingestion is an asynchronous operation, its status will be reported by `bulk_load_response` like download status, `on_partition_ingestion_reply` will only handle write 2pc errors and empty write errors.

### DIFF
--- a/src/dist/replication/meta_server/meta_bulk_load_service.cpp
+++ b/src/dist/replication/meta_server/meta_bulk_load_service.cpp
@@ -804,7 +804,6 @@ void bulk_load_service::on_partition_ingestion_reply(error_code err,
                                                      const std::string &app_name,
                                                      const gpid &pid)
 {
-    // TODO(heyuchen): TBD
     // if meet 2pc error, ingesting will rollback to downloading, no need to retry here
     if (err != ERR_OK) {
         derror_f("app({}) partition({}) ingestion files failed, error = {}", app_name, pid, err);
--- a/src/dist/replication/meta_server/meta_bulk_load_service.cpp
+++ b/src/dist/replication/meta_server/meta_bulk_load_service.cpp
@@ -805,6 +805,46 @@ void bulk_load_service::on_partition_ingestion_reply(error_code err,
                                                      const gpid &pid)
 {
     // TODO(heyuchen): TBD
+    // if meet 2pc error, ingesting will rollback to downloading, no need to retry here
+    if (err != ERR_OK) {
+        derror_f("app({}) partition({}) ingestion files failed, error = {}", app_name, pid, err);
+        tasking::enqueue(
+            LPC_META_STATE_NORMAL,
+            _meta_svc->tracker(),
+            std::bind(&bulk_load_service::try_rollback_to_downloading, this, app_name, pid));
+        return;
+    }
+
+    if (resp.err == ERR_TRY_AGAIN && resp.rocksdb_error != 0) {
+        derror_f("app({}) partition({}) ingestion files failed while empty write, rocksdb error = "
+                 "{}, retry it later",
+                 app_name,
+                 pid,
+                 resp.rocksdb_error);
+        tasking::enqueue(LPC_BULK_LOAD_INGESTION,
+                         _meta_svc->tracker(),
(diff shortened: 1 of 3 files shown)


src/dist/replication/meta_server/meta_bulk_load_service.cpp
```diff
@@ -805,6 +805,46 @@ void bulk_load_service::on_partition_ingestion_reply(error_code err,
                                                      const gpid &pid)
 {
     // TODO(heyuchen): TBD
```
neverchanje, line 807: Still TBD?

src/dist/replication/meta_server/meta_bulk_load_service.cpp
```diff
@@ -805,6 +805,46 @@ void bulk_load_service::on_partition_ingestion_reply(error_code err,
...
+        derror_f("app({}) partition({}) ingestion files failed while empty write, rocksdb error = "
+                 "{}, retry it later",
+                 app_name,
+                 pid,
+                 resp.rocksdb_error);
+        tasking::enqueue(LPC_BULK_LOAD_INGESTION,
+                         _meta_svc->tracker(),
+                         std::bind(&bulk_load_service::partition_ingestion, this, app_name, pid),
+                         0,
+                         std::chrono::milliseconds(10));
```
neverchanje, line 828: What 10ms supposed to do? Will it make any sense while rocksdb failing to ingest files?

src/dist/replication/meta_server/meta_bulk_load_service.cpp
```diff
@@ -805,6 +805,46 @@ void bulk_load_service::on_partition_ingestion_reply(error_code err,
...
+        return;
+    }
+
+    // some unexpected errors happened, such as write empty write failed but rocksdb_error is ok
```
neverchanje, line 832: "such as write empty write failed but rocksdb_error is ok"

Why does rocksdb allow ingesting an empty write? I have no questions on the code but just curious about this comment.

hycdong: You can see https://github.com/XiaoMi/pegasus/pull/546 firstly, which is step3, this pull request is step 4.

src/dist/replication/meta_server/meta_bulk_load_service.cpp
hycdong: This condition will be happened when empty write failed before ingestion, we should retry this request. The delay time I set 10ms to let retry quickly, because current replica has already rejected normal write requests.

src/dist/replication/meta_server/meta_bulk_load_service.cpp
```diff
@@ -804,7 +804,46 @@ void bulk_load_service::on_partition_ingestion_reply(error_code err,
...
+    // stop bulk load process with failed
+    if (resp.err != ERR_OK || resp.rocksdb_error != 0) {
+        derror_f("app({}) partition({}) failed to ingestion files, error = {}, rocksdb error = {}",
```
acelyc111, line 834: All primary and secondaries have its own ingestion procedures, right? So add replica address in log would be helpful for trouble shooting.

hycdong: Yes, but just like normal write request, client will only receive update_response from primary, meta server will only receive ingestion_response from primary.

src/dist/replication/meta_server/meta_bulk_load_service.cpp
neverchanje: So why not retry immediately?

PR #495, opened 2020-06-08
feat: add perf counter for group check failure

levy5307:
add perf counter for group check


src/dist/replication/lib/replica_check.cpp
```diff
@@ -207,16 +207,20 @@ void replica::on_group_check_reply(error_code err,
...
 
-    if (err != ERR_OK) {
+    if (err != ERR_OK || resp->err != ERR_OK) {
+        if (ERR_OK == err) {
+            err = resp->err;
+        }
         handle_remote_failure(req->config.status, req->node, err, "group check");
+        // only if status == PS_SECONDARY should we increase group check fail count,
+        // because other status have no significant effect
+        if (partition_status::PS_SECONDARY == req->config.status) {
```
hycdong, line 217: I think it should include other partition status. Group check is for replica's heart beat, this counter is to show how many group check failed, should contain all status. Besides, for prepare request, its counter `_counter_replicas_recent_prepare_fail_count` also contain other status.

levy5307: Yes, prepare does what you said. But I think other status have no signficant effect on the performance of replica server, So I think there is no need to add them in this perf counter.

hycdong: I think `other status have no significant effect on the performance of replica server` is not the reason exclude other status. Group check is like heartbeat in replica group, should contain all status, should not only contain the status who has significant on performance.
Just like counter `_counter_replicas_recent_prepare_fail_count`, it includes secondary, inactive, potential_secondary partition's prepare failed count, because partition with those status will prepare failed, the counter should record all possible status, potential_secondary partition prepare failed has smaller influence on performance, but it also should be recorded.

levy5307: The purpose that we add this perf counter is to find the group check failure which have bad influence on performance. So if we take all status in consider, we can't see the problem visually
.

acelyc111: @levy5307 If you want only count when `status == PS_SECONDARY`, maybe the counter's name `_counter_replicas_recent_group_check_fail_count` is not very suitable.

levy5307: @acelyc111 yes, you are right,  I was going to use `_counter_replicas_recent_group_check_secondary_fail_count`, but it is a little too long. Do you have any good ideas?

neverchanje: @hycdong I agree we need to monitor the group-check failures on both PS and S. But they should be two counters I think. Failures on S and PS have different meanings.

hycdong: @neverchanje Could you please explain why `Failures on S and PS have different meanings`? In my view, group check is for replica group heartbeat, failed on secondary and potential secondary has same meanings.

neverchanje: OK. I was wrong. Group check has the same priority on S and PS. I agree with you to count learner failures.

PR #496, opened 2020-06-10
feat(bulk-load): bulk load ingestion part5 - replica handle bulk load request during ingestion

hycdong:
    - primary consider ingestion request as a client write and replica group execute 2pc process to do actual rocksdb ingestion operation #490 pegasus pr [#546](https://github.com/XiaoMi/pegasus/pull/546)
    - **primary handle bulk load request** and broadcast it to secondary #460 
    - ingestion files through 2pc #490 pegasus pr [#546](https://github.com/XiaoMi/pegasus/pull/546)
    - **report ingestion status to primary**
5. **primary report group ingestion status to meta server**
6. meta handle two response
    - handle ingestion_response #493 

This pull request is about how replica handle bulk_load_request during ingestion.
- **implement function `do_bulk_load`**
     - this function will be called when primary receives `bulk_load_request` and secondary receives `group_bulk_load_request`. It will first check primary status, and compare bulk load status from request and replica local bulk load status (in function `validate_bulk_load_status`, not implemented in this pull request), then call different functions according to bulk load status.
    - when request status is downloading, and local status is invalid, will call `start_download` to start download files from file provider.
    - when request status is ingestion, local status is downloaded, will call `start_ingestion`
    - when request status is ingestion, local status is ingestion, will call `check_ingestion_finish`
- **ingestion is considered as a write request, but it is different from normal write, we should add some special checking for it**
    - **write an empty put after ingestion.** For write request, when primary receives secondaries prepare reply, it will commit this mutation, and secondaries will commit the mutation when they prepare next mutation, so send an empty prepare to guarantee secondary commit ingestion request.
    - **pop all committed mutation when prepare the empty put.** For ingestion mutation, replay private log can not learn data ingested. To make data correct, we will create a checkpoint ( not implemented in this pull request), and should guarantee learner learn data from checkpoint. To satisfy the condition above, we should pop all committed mutation when prepare the empty put, so the prepare list is empty, learn type will always be LT_APP, learn checkpoint firstly. 
        - add `ingestion_is_empty_prepare_sent` in primary_context to write empty put only once.
        - add `pop_all_committed_mutations` parameter for function `init_prepare`, `send_prepare_message`, `prepare`, add `pop_all` for thrift structure `replica_configuration`.
- **report ingestion status**
    - secondary report ingestion status in function `report_bulk_load_states_to_primary`
    - primary report group ingestion status in function `report_group_ingestion_status`, when group ingestion succeed, recover normal write

### DIFF
--- a/src/dist/replication/lib/bulk_load/replica_bulk_loader.cpp
+++ b/src/dist/replication/lib/bulk_load/replica_bulk_loader.cpp
@@ -194,14 +194,52 @@ error_code replica_bulk_loader::do_bulk_load(const std::string &app_name,
                                              const std::string &cluster_name,
                                              const std::string &provider_name)
 {
+    if (status() != partition_status::PS_PRIMARY && status() != partition_status::PS_SECONDARY) {
+        return ERR_INVALID_STATE;
+    }
+
+    bulk_load_status::type local_status = _status;
+    error_code ec = validate_bulk_load_status(meta_status, local_status);
+    if (ec != ERR_OK) {
+        derror_replica("invalid bulk load status, remote = {}, local = {}",
+                       enum_to_string(meta_status),
+                       enum_to_string(local_status));
+        return ec;
+    }
+
+    switch (meta_status) {
+    case bulk_load_status::BLS_DOWNLOADING:
+        // TODO(heyuchen): add restart downloading status check
+        if (local_status == bulk_load_status::BLS_INVALID) {
+            ec = start_download(app_name, cluster_name, provider_name);
+        }
+        break;
+    case bulk_load_status::BLS_INGESTING:
+        if (local_status == bulk_load_status::BLS_DOWNLOADED) {
+            start_ingestion();
+        } else if (local_status == bulk_load_status::BLS_INGESTING &&
+                   status() == partition_status::PS_PRIMARY) {
+            check_ingestion_finish();
+        }
+        break;
+    // TODO(heyuchen): add other bulk load status
+    default:
+        break;
+    }
+    return ec;
+}
+
+error_code replica_bulk_loader::validate_bulk_load_status(bulk_load_status::type meta_status,
+                                                          bulk_load_status::type local_status)
+{
     // TODO(heyuchen): TBD
     return ERR_OK;
 }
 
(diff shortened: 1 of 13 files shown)


src/dist/replication/lib/bulk_load/replica_bulk_loader.h
```diff
@@ -33,12 +33,18 @@ class replica_bulk_loader : replica_base
                             const std::string &cluster_name,
                             const std::string &provider_name);
 
+    // compare meta bulk load status and local bulk load status
+    // \return ERR_INVALID_STATE if it is invalid for replica whose local status is `local_status`
+    // while meta status is `meta_status`
```
acelyc111, line 38: I really don't know what does this sentence mean :(

hycdong: oh, I have updated the comment.

src/dist/replication/lib/bulk_load/replica_bulk_loader.cpp
```diff
@@ -380,6 +418,35 @@ void replica_bulk_loader::check_download_finish()
...
+{
+    _status = bulk_load_status::BLS_INGESTING;
+    // TODO(heyuchen): add perf-counter
+    if (status() == partition_status::PS_PRIMARY) {
+        _replica->_primary_states.ingestion_is_empty_prepare_sent = false;
+    }
+}
+
+// ThreadPool: THREAD_POOL_REPLICATION
+void replica_bulk_loader::check_ingestion_finish()
```
Shuo-Jia, line 432: I understood the method is `check whether the ingetion is finished`, but the logic seem mean `gurantee it can be finished`?

hycdong: Well, it is depends on when ingestion is finished, in my view, this function means check whether primary replica's ingestion_status is ingestion_succeed, besides, if it has not sent empty prepare to secondaries, sends empty prepare to make secondaries commit this mutation, I think it is to guarantee data integrity. Actually, when primary sends empty prepare to secondaries, ingestion is not finished. All replicas' ingestion_status are ingestion_succeed means ingestion finished.

PR #497, opened 2020-06-10
fix: fix RPC_CHECK_STATUS incorrect usage in meta_service 

levy5307:
In meta_service, there are some rpc interfaces which are using `rpc_holder`. And they use `RPC_CHECK_STATUS` to do some check work. But it will cause the response to be sent repeatedly, for example:
1. A response will replied in `RPC_CHECK_STATUS`, and another response will be sent in the dstor of rpc_holder. So one response is sent repeatedly here.

2. If this node is not a master, a forward message will send in this macro, And this situation is same as above.

The implementation of `RPC_CHECK_STATUS`: 
```
#define RPC_CHECK_STATUS(dsn_msg, response_struct)                                                 \
    dinfo("rpc %s called", __FUNCTION__);                                                          \
    int result = check_leader(dsn_msg, nullptr);                                                   \
    if (result == 0)                                                                               \
        return;                                                                                    \
    if (result == -1 || !_started) {                                                               \
        if (result == -1)                                                                          \
            response_struct.err = ERR_FORWARD_TO_OTHERS;                                           \
        else if (_recovering)                                                                      \
            response_struct.err = ERR_UNDER_RECOVERY;                                              \
        else                                                                                       \
            response_struct.err = ERR_SERVICE_NOT_ACTIVE;                                          \
        ddebug("reject request with %s", response_struct.err.to_string());                         \
        reply(dsn_msg, response_struct);                                                           \
        return;                                                                                    \
    }
```

### DIFF
--- a/include/dsn/cpp/rpc_holder.h
+++ b/include/dsn/cpp/rpc_holder.h
@@ -223,19 +223,30 @@ public:
     using mail_box_u_ptr = std::unique_ptr<mail_box_t>;
     static void enable_mocking()
     {
-        dassert(_mail_box == nullptr, "remember to call clear_mocking_env after testing");
+        dassert(_mail_box == nullptr && _forward_mail_box == nullptr,
+                "remember to call clear_mocking_env after testing");
         _mail_box = make_unique<mail_box_t>();
+        _forward_mail_box = make_unique<mail_box_t>();
     }
 
     // Only use this function when testing.
     // Remember to call it after test finishes, or it may effect the results of other tests.
     // This function is not thread-safe.
-    static void clear_mocking_env() { _mail_box.reset(nullptr); }
+    static void clear_mocking_env() {
+        _mail_box.reset(nullptr);
+        _forward_mail_box.reset(nullptr);
+    }
 
     static mail_box_t &mail_box()
     {
         dassert(_mail_box != nullptr, "call this function only when you are in mock mode");
         return *_mail_box.get();
+    }
+
+    static mail_box_t &forward_mail_box()
+    {
+        dassert(_forward_mail_box != nullptr, "call this function only when you are in mock mode");
+        return *_forward_mail_box.get();
     }
 
     friend bool operator<(const rpc_holder &lhs, const rpc_holder &rhs) { return lhs._i < rhs._i; }
@@ -287,6 +298,14 @@ private:
         void forward(rpc_address addr)
         {
             auto_reply = false;
+            if (dsn_unlikely(_forward_mail_box != nullptr)) {
+                rpc_holder<TRequest, TResponse> rpc(std::move(thrift_request),
+                                                    dsn_request->rpc_code());
+                rpc.response() = std::move(thrift_response);
+                _forward_mail_box->emplace_back(std::move(rpc));
+                return;
+            }
+
             dsn_rpc_forward(dsn_request, addr);
         }
 
@@ -308,6 +327,7 @@ private:
     std::shared_ptr<internal> _i;
 
     static mail_box_u_ptr _mail_box;
+    static mail_box_u_ptr _forward_mail_box;
 };
 
 // ======== type traits ========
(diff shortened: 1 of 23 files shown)


include/dsn/cpp/rpc_holder.h
```diff
@@ -282,6 +284,12 @@ class rpc_holder
             dsn_rpc_reply(dsn_response);
         }
 
+        void forward(rpc_address addr)
```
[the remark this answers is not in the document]
levy5307: good idea

include/dsn/cpp/rpc_holder.h
```diff
@@ -282,6 +296,20 @@ class rpc_holder
...
+        {
+            auto_reply = false;
+            if (dsn_unlikely(_forward_mail_box != nullptr)) {
+                rpc_holder<TRequest, TResponse> rpc(std::move(thrift_request),
+                                                    dsn_request->rpc_code());
+                rpc.response() = std::move(thrift_response);
+                _forward_mail_box->emplace_back(std::move(rpc));
```
neverchanje, line 306:
```suggestion
                _forward_mail_box->emplace_back(*this);
```

src/core/tests/rpc_holder_test.cpp
```diff
@@ -122,3 +122,26 @@ TEST(rpc_holder, mock_rpc_reply)
...
+    {
+        auto &mail_box = t_rpc::mail_box();
+        auto &forward_mail_box = t_rpc::forward_mail_box();
+        rpc_address forward_addr("127.0.0.1", 10086);
+
+        for (int i = 0; i < 10; i++) {
+            configuration_query_by_index_request request;
+            auto msg = from_thrift_request_to_received_message(
+                request, RPC_CM_QUERY_PARTITION_CONFIG_BY_INDEX);
+            auto rpc = t_rpc::auto_reply(msg);
```
neverchanje, line 138:
```suggestion
```
You can not reply as the RPC caller. This case is not suitable.

src/dist/replication/meta_server/meta_service.cpp
```diff
@@ -93,6 +93,52 @@ bool meta_service::check_freeze() const
...
+            if (forward_address != nullptr)
+                forward_address->set_invalid();
+            return -1;
+        }
+    }
+    return 1;
+}
+
+template <typename TRpcHolder>
+bool meta_service::check_status(TRpcHolder rpc, rpc_address *forward_address)
```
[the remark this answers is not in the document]
levy5307: I don't think `try_forward_if_not_leader` is a good name. Because the main purpose for this function is to do some check work.

src/dist/replication/meta_server/meta_service.h
```diff
@@ -218,6 +218,10 @@ class meta_service : public serverlet<meta_service>
     //  -1. meta isn't leader, and rpc-msg can't forward to others
     // if return -1 and `forward_address' != nullptr, then return leader by `forward_address'.
     int check_leader(dsn::message_ex *req, dsn::rpc_address *forward_address);
+    template <typename TRpcHolder>
+    int check_leader(TRpcHolder rpc, /*out*/ rpc_address *forward_address);
+    template <typename TRpcHolder>
+    bool check_status(TRpcHolder rpc, /*out*/ rpc_address *forward_address = nullptr);
```
neverchanje, line 224: I checked all the callers of `check_status`. None is using the parameter `forward_address `. It's useless I think.
```suggestion
    bool check_status(TRpcHolder rpc);
```

levy5307: I will use it in the later pull request, which needs the forward_address.

src/dist/replication/meta_server/meta_server_failure_detector.cpp
```diff
@@ -83,6 +85,23 @@ void meta_server_failure_detector::on_worker_connected(rpc_address node)
 
 bool meta_server_failure_detector::get_leader(rpc_address *leader)
 {
+    FAIL_POINT_INJECT_F("meta_server_failure_detector_get_leader", [leader](dsn::string_view str) {
+        auto pos = str.find("#");
```
hycdong, line 89: Could you please add comments to explain `leader` format? because leader format for unit test `meta_service_test.cpp` is different from structure `rpc_address`, I just can't understand it before I read unit tests.

PR #501, opened 2020-06-15
refactor: use check_status to reimplement some rpc interfaces of meta_service

levy5307:
use check_status to reimplement some rpc interfaces of meta_service: including `on_query_configuration_by_index` and `on_start_recovery`

### DIFF
--- a/src/dist/replication/meta_server/meta_service.cpp
+++ b/src/dist/replication/meta_server/meta_service.cpp
@@ -94,11 +94,13 @@ bool meta_service::check_freeze() const
 }
 
 template <typename TRpcHolder>
-int meta_service::check_leader(TRpcHolder rpc)
+int meta_service::check_leader(TRpcHolder rpc, rpc_address *forward_address)
 {
     dsn::rpc_address leader;
     if (!_failure_detector->get_leader(&leader)) {
         if (!rpc.dsn_request()->header->context.u.is_forward_supported) {
+            if (forward_address != nullptr)
+                *forward_address = leader;
             return -1;
         }
 
@@ -107,6 +109,8 @@ int meta_service::check_leader(TRpcHolder rpc)
             rpc.forward(leader);
             return 0;
         } else {
+            if (forward_address != nullptr)
+                forward_address->set_invalid();
             return -1;
         }
     }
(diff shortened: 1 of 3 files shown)


src/dist/replication/meta_server/meta_service.cpp
```diff
@@ -603,30 +603,13 @@ void meta_service::on_query_cluster_info(configuration_cluster_info_rpc rpc)
...
-                partition_configuration config;
-                config.primary = forward_address;
-                response.partitions.push_back(std::move(config));
-            }
-        } else if (_recovering) {
-            response.err = ERR_UNDER_RECOVERY;
-        } else {
-            response.err = ERR_SERVICE_NOT_ACTIVE;
+    if (!check_status(rpc, &forward_address)) {
+        if (!forward_address.is_invalid()) {
```
neverchanje, line 608: What if `check_leader` returns 0?  `check_status` will then return false, with no change to `forward_address`. 
Since that `forward_address` is invalid by default, it will come into this logic block.

```cpp
void meta_service::on_query_configuration_by_index(configuration_query_by_index_rpc rpc)
{
    configuration_query_by_index_response &response = rpc.response();
    rpc_address forward_address;
    if (!check_status(rpc, &forward_address)) { // returns false when check_leader returns 0
        if (!forward_address.is_invalid()) { // invalid by default
            partition_configuration config; // formerly it returns directly, without coming into this logic
            config.primary = forward_address;
            response.partitions.push_back(std::move(config));
        }
        return;
    }
```

In this situation, we formerly return directly (because the RPC is already forwarded). But now you make it responds to the client with the leader address.

```cpp
void meta_service::on_query_configuration_by_index(configuration_query_by_index_rpc rpc)
{
    configuration_query_by_index_response &response = rpc.response();
    rpc_address forward_address;
    int result = check_leader(rpc.dsn_request(), &forward_address);
    if (result == 0) { //  check_leader returns 0.
        rpc.disable_auto_reply();
        return; // it returns directly
    }
```

Is this change reasonable, or an unconsidered case?

src/dist/replication/meta_server/meta_service.cpp
```diff
@@ -732,28 +715,15 @@ void meta_service::on_start_recovery(configuration_recovery_rpc rpc)
...
-                   dsn_primary_address().to_string());
-            response.err = ERR_SERVICE_ALREADY_RUNNING;
-        } else {
-            _state->on_start_recovery(rpc.request(), response);
-            if (response.err == dsn::ERR_OK) {
-                _recovering = false;
-                start_service();
-            }
-        }
+    zauto_write_lock l(_meta_lock);
```
neverchanje, line 722:
```suggestion
    zauto_write_lock l(_meta_lock);
    if (_started.load()) {
        ddebug("service(%s) is already started, ignore the recovery request",
               dsn_primary_address().to_string());
        response.err = ERR_SERVICE_ALREADY_RUNNING;
    } else {
```

Why you remove this part?

levy5307: yes, you are right. We can't refactor `on_start_recovery` by using `check_status`

PR #500, opened 2020-06-15
feat(bulk-load): bulk load ingestion part6 - meta handle bulk_load_response during ingestion

hycdong:
The whole bulk load ingestion process is like:
1. meta server set app and all partitions bulk load status as ingesting #486 
2. meta send two requests to primary 
    - send ingestion request to primary #486 
    - send bulk load request to query ingestion status #457
3. primary handle two requests above:
    - primary consider ingestion request as a client write and replica group execute 2pc process to do actual rocksdb ingestion operation #490 pegasus pr [#546](https://github.com/XiaoMi/pegasus/pull/546)
    - primary handle bulk load request and broadcast it to secondary #496 #460
4. secondary handle two cases:
    - ingestion files through 2pc #490 pegasus pr [#546](https://github.com/XiaoMi/pegasus/pull/546)
    - report ingestion status to primary #496 
5. primary report group ingestion status to meta server #496 
6. meta handle two response
    - handle ingestion_response #493 
    - **handle group ingestion status and do bulk load status transformation**

This pull request is about meta handle ingestion status reported by replica server, part of step 6 above.
- meta handle ingestion status function `handle_app_ingestion`
    - check if `is_group_ingestion_finished` and replica's `ingest_status` are set, if not set, meta should not update partition ingestion_status in this round
    - if any replica's ingestion status is `IS_FAILED`, we consider it as an unrecoverable error, bulk load will set as failed
    - if all replicas of the group ingest succeed, meta will set this partition status from ingesting to succeed
- bulk load status transformation
    - if all partitions status turn from ingesting to succeed, meta will set app status from ingestion to succeed in `update_partition_status_on_remote_storage_reply`

PR #502, opened 2020-06-16
feat(bulk-load): bulk load succeed part1 - replica handle bulk load succeed

hycdong:
The whole bulk load succeed process is like:
1. meta server set app and all partitions bulk load status is succeed #500 
2. meta server send bulk_load_request to primary #457
3. **primary handle bulk_load_request**
5. **secondary handle bulk_load_request**
6. **secondary report cleanup flag to primary**
7. **primary report group cleanup flag  to meta server**
8. meta handle bulk load succeed

This pull request is about replica handle bulk load succeed, step 3,5,6,7 above.
- **handle bulk_load_request in function `do_bulk_load`**
    - when request status is succeed, local status is ingesting, will call `handle_bulk_load_succeed`
    - when request status is succeed, local status is succeed, will call `handle_bulk_load_finish`
    - both primary and secondary will execute this function
- **handle_bulk_load_succeed**
    - create checkpoint here, this newest checkpoint will include data ingested from files
    - reset ingestion status and turn replica status to succeed
- **handle_bulk_load_finish**
    - remove local bulk load directory who stores downloaded files (function `remove_local_bulk_load_dir`)
    - reset other local bulk load states and context (function `clear_bulk_load_states` and another function in further pull request)
    - this function will also be called when bulk load failed and canceling bulk load, so I call it `handle_bulk_load_finish`
- **report states**
    - secondary will report whether itself bulk load context cleaned (function `report_bulk_load_states_to_primary`)
    - primary will report group flag to meta (function `report_group_cleanup_flag`)

### DIFF
--- a/src/dist/replication/lib/bulk_load/replica_bulk_loader.cpp
+++ b/src/dist/replication/lib/bulk_load/replica_bulk_loader.cpp
@@ -222,6 +222,13 @@ error_code replica_bulk_loader::do_bulk_load(const std::string &app_name,
             check_ingestion_finish();
         }
         break;
+    case bulk_load_status::BLS_SUCCEED:
+        if (local_status == bulk_load_status::BLS_INGESTING) {
+            handle_bulk_load_succeed();
+        } else if (local_status == bulk_load_status::BLS_SUCCEED) {
+            handle_bulk_load_finish(meta_status);
+        }
+        break;
     // TODO(heyuchen): add other bulk load status
     default:
         break;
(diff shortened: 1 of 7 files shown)


src/dist/replication/lib/bulk_load/replica_bulk_loader.cpp
```diff
@@ -447,6 +454,59 @@ void replica_bulk_loader::check_ingestion_finish()
...
+            get_gpid().thread_hash());
+    }
+
+    clear_bulk_load_states();
+}
+
+// ThreadPool: THREAD_POOL_REPLICATION
+error_code replica_bulk_loader::remove_local_bulk_load_dir(const std::string &bulk_load_dir)
+{
+    if (!utils::filesystem::directory_exists(bulk_load_dir) ||
```
acelyc111, line 502: Is there any problem if we return `ERR_OK ` when directory is not exist?

src/dist/replication/lib/bulk_load/replica_bulk_loader.cpp
```diff
@@ -577,6 +666,42 @@ void replica_bulk_loader::report_group_ingestion_status(/*out*/ bulk_load_respon
...
+    if (status() != partition_status::PS_PRIMARY) {
+        dwarn_replica("replica status={}, should be {}",
+                      enum_to_string(status()),
+                      enum_to_string(partition_status::PS_PRIMARY));
+        response.err = ERR_INVALID_STATE;
+        return;
+    }
+
+    partition_bulk_load_state primary_state;
+    primary_state.__set_is_cleanuped(is_cleanuped());
```
acelyc111, line 681: cleanuped -> cleaned_up?

src/dist/replication/lib/bulk_load/replica_bulk_loader.cpp
```diff
@@ -456,10 +516,36 @@ void replica_bulk_loader::cleanup_download_task()
...
+}
+
+// ThreadPool: THREAD_POOL_REPLICATION
+bool replica_bulk_loader::is_cleanuped()
+{
+    return _status == bulk_load_status::type::BLS_INVALID && _cur_downloaded_size.load() == 0 &&
+           _download_progress.load() == 0 && _download_status.load() == ERR_OK &&
+           _download_task.size() == 0 && _metadata.files.size() == 0 &&
+           _metadata.file_total_size == 0 && !_replica->_is_bulk_load_ingestion &&
+           _replica->_app->get_ingestion_status() == ingestion_status::IS_INVALID;
```
acelyc111, line 548: Anyway, I think this sentence is too long :)

src/dist/replication/lib/bulk_load/replica_bulk_loader.cpp
hycdong: It is okay

PR #504, opened 2020-06-17
refactor: move rpc module to a separated directory

levy5307:
put rpc in a seperate directory


src/core/rpc/rpc_task.cpp
```diff
@@ -0,0 +1,113 @@
+// Copyright (c) 2017, Xiaomi, Inc.  All rights reserved.
+// This source code is licensed under the Apache License Version 2.0, which
+// can be found in the LICENSE file in the root directory of this source tree.
```
neverchanje, line 3: this file should be MIT licensed. the property belongs to Microsoft.

PR #507, opened 2020-06-18
refactor: use DSN_REGISTER_COMPONENT_PROVIDER to register aio_provider

levy5307:
***1.*** In rdsn, if we want to register a component provider, we should add register_component_provider in providers.common.cpp, which is a file in dsn_runtime. So the module is coupled with dsn_runtime.
In this pull request, we implement DSN_REGISTER_COMPONENT_PROVIDER. If you want to register a component provider for your module, you can add a DSN_REGISTER_COMPONENT_PROVIDER in this module, and let this module decoupled from dsn_runtime.

***2.*** use DSN_REGISTER_COMPONENT_PROVIDER to register aio_provider


src/core/aio/disk_engine.cpp
```diff
@@ -138,12 +142,9 @@ disk_engine::disk_engine()
...
     _node = service_engine::instance().get_all_nodes().begin()->second.get();
 
-    // use native_linux_aio_provider in default
-    if (!strcmp(FLAGS_aio_factory_name, "dsn::tools::sim_aio_provider")) {
-        _provider.reset(new aio::sim_aio_provider(this, nullptr));
-    } else {
-        _provider.reset(new native_linux_aio_provider(this, nullptr));
-    }
+    aio_provider *provider = utils::factory_store<aio_provider>::create(
+        FLAGS_aio_factory_name, dsn::PROVIDER_TYPE_MAIN, this, nullptr);
```
vagetablechicken, line 146: If FLAGS_aio_factory_name is wrong, provider will be nullptr.Is any protection for null _provider?

levy5307: I can add a assert here.

PR #508, opened 2020-06-18
feat(bulk-load): bulk load succeed part2 - meta handle bulk load succeed

hycdong:
The whole bulk load succeed process is like:
1. meta server set app and all partitions bulk load status is succeed #500 
2. meta server send bulk_load_request to primary #457
3. primary handle bulk_load_request #502 
5. secondary handle bulk_load_request #502 
6. secondary report cleanup flag to primary #502 
7. primary report group cleanup flag  to meta server #502 
8. **meta handle bulk load succeed**

This pull request is about meta handle bulk load succeed, step 8 above.

meta handle bulk_load_response in `handle_bulk_load_finish`
- if all replicas in group cleaned up bulk load state, set `_partitions_cleaned_up[pid] = true` and `_apps_in_progress_count[app_id] - 1`
- if all partitions cleaned up, remove bulk load directory(in function `remove_bulk_load_dir_on_remote_storage`), reset local bulk load states(in function `reset_local_bulk_load_states`), set app's is_bulk_loading as false (in function `update_app_not_bulk_loading_on_remote_storage`), whole bulk load process finished

**_Tips:_** 
when app's is_bulk_loading = false, meta will not send bulk load request to replica server any more in function `try_resend_bulk_load_request` #463

### DIFF
--- a/src/dist/replication/meta_server/meta_bulk_load_service.cpp
+++ b/src/dist/replication/meta_server/meta_bulk_load_service.cpp
@@ -575,30 +575,24 @@ void bulk_load_service::handle_bulk_load_finish(const bulk_load_response &respon
     const std::string &app_name = response.app_name;
     const gpid &pid = response.pid;
 
-    if (!response.__isset.is_group_bulk_load_context_cleaned_up) {
-        dassert_f(
-            false,
-            "receive bulk load response from node({}) app({}), partition({}), primary_status({}), "
-            "but is_group_bulk_load_context_cleaned_up is not set",
-            primary_addr.to_string(),
-            app_name,
-            pid,
-            dsn::enum_to_string(response.primary_bulk_load_status));
-        return;
-    }
+    dassert_f(
+        response.__isset.is_group_bulk_load_context_cleaned_up,
+        "receive bulk load response from node({}) app({}), partition({}), primary_status({}), "
+        "but is_group_bulk_load_context_cleaned_up is not set",
+        primary_addr.to_string(),
+        app_name,
+        pid,
+        dsn::enum_to_string(response.primary_bulk_load_status));
 
     for (const auto &kv : response.group_bulk_load_state) {
-        if (!kv.second.__isset.is_cleaned_up) {
-            dassert_f(false,
-                      "receive bulk load response from node({}) app({}), partition({}), "
-                      "primary_status({}), but node({}) is_cleaned_up is not set",
-                      primary_addr.to_string(),
-                      app_name,
-                      pid,
-                      dsn::enum_to_string(response.primary_bulk_load_status),
-                      kv.first.to_string());
-            return;
-        }
+        dassert_f(kv.second.__isset.is_cleaned_up,
+                  "receive bulk load response from node({}) app({}), partition({}), "
+                  "primary_status({}), but node({}) is_cleaned_up is not set",
+                  primary_addr.to_string(),
+                  app_name,
+                  pid,
+                  dsn::enum_to_string(response.primary_bulk_load_status),
+                  kv.first.to_string());
     }
 
     {
--- a/src/dist/replication/meta_server/meta_bulk_load_service.cpp
+++ b/src/dist/replication/meta_server/meta_bulk_load_service.cpp
@@ -627,7 +627,7 @@ void bulk_load_service::handle_bulk_load_finish(const bulk_load_response &respon
     }
 
     if (group_cleaned_up) {
-        int32_t count;
+        int32_t count = 0;
         {
             zauto_write_lock l(_lock);
             count = --_apps_in_progress_count[pid.get_app_id()];
--- a/src/dist/replication/meta_server/meta_bulk_load_service.cpp
+++ b/src/dist/replication/meta_server/meta_bulk_load_service.cpp
@@ -1003,6 +1003,8 @@ inline void erase_map_elem_by_id(int32_t app_id, std::unordered_map<gpid, T> &my
     for (auto iter = mymap.begin(); iter != mymap.end();) {
         if (iter->first.get_app_id() == app_id) {
             mymap.erase(iter++);
+        } else {
+            iter++;
         }
     }
 }
--- a/src/dist/replication/meta_server/meta_bulk_load_service.cpp
+++ b/src/dist/replication/meta_server/meta_bulk_load_service.cpp
@@ -572,8 +572,84 @@ void bulk_load_service::handle_app_ingestion(const bulk_load_response &response,
 void bulk_load_service::handle_bulk_load_finish(const bulk_load_response &response,
                                                 const rpc_address &primary_addr)
 {
-    // TODO(heyuchen): TBD
-    // called by `on_partition_bulk_load_reply` when app status is succeed, failed, canceled
+    const std::string &app_name = response.app_name;
+    const gpid &pid = response.pid;
+
+    if (!response.__isset.is_group_bulk_load_context_cleaned_up) {
+        dwarn_f(
+            "receive bulk load response from node({}) app({}), partition({}), primary_status({}), "
+            "but is_group_bulk_load_context_cleaned_up is not set",
+            primary_addr.to_string(),
+            app_name,
+            pid,
+            dsn::enum_to_string(response.primary_bulk_load_status));
+        return;
+    }
+
+    for (const auto &kv : response.group_bulk_load_state) {
+        if (!kv.second.__isset.is_cleaned_up) {
+            dwarn_f("receive bulk load response from node({}) app({}), partition({}), "
+                    "primary_status({}), but node({}) is_cleaned_up is not set",
+                    primary_addr.to_string(),
+                    app_name,
+                    pid,
+                    dsn::enum_to_string(response.primary_bulk_load_status),
+                    kv.first.to_string());
+            return;
+        }
+    }
+
+    {
+        zauto_read_lock l(_lock);
+        if (_partitions_cleaned_up[pid]) {
+            dwarn_f(
+                "receive bulk load response from node({}) app({}) partition({}), current partition "
+                "has already be cleaned up",
+                primary_addr.to_string(),
+                app_name,
+                pid);
+            return;
+        }
+    }
+
+    bool group_cleaned_up = response.is_group_bulk_load_context_cleaned_up;
+    ddebug_f("receive bulk load response from node({}) app({}) partition({}), primary status = {}, "
+             "is_group_bulk_load_context_cleaned_up = {}",
+             primary_addr.to_string(),
+             app_name,
+             pid,
+             dsn::enum_to_string(response.primary_bulk_load_status),
+             group_cleaned_up);
+    {
+        zauto_write_lock l(_lock);
+        _partitions_cleaned_up[pid] = group_cleaned_up;
+        _partitions_bulk_load_state[pid] = response.group_bulk_load_state;
+    }
+
+    if (group_cleaned_up) {
+        int32_t count;
+        {
+            zauto_write_lock l(_lock);
+            count = --_apps_in_progress_count[pid.get_app_id()];
+        }
+        if (count == 0) {
+            std::shared_ptr<app_state> app;
+            {
+                zauto_read_lock l(app_lock());
+                app = _state->get_app(pid.get_app_id());
+                if (app == nullptr || app->status != app_status::AS_AVAILABLE) {
+                    dwarn_f("app(name={}, id={}) is not existed, remove bulk load dir on remote "
+                            "storage",
+                            app_name,
+                            pid.get_app_id());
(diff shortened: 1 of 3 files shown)


src/dist/replication/meta_server/meta_bulk_load_service.cpp
```diff
@@ -572,8 +572,84 @@ void bulk_load_service::handle_app_ingestion(const bulk_load_response &response,
...
+             dsn::enum_to_string(response.primary_bulk_load_status),
+             group_cleaned_up);
+    {
+        zauto_write_lock l(_lock);
+        _partitions_cleaned_up[pid] = group_cleaned_up;
+        _partitions_bulk_load_state[pid] = response.group_bulk_load_state;
+    }
+
+    if (group_cleaned_up) {
+        int32_t count;
```
levy5307, line 630: `int32_t count = 0;`
I think specify an initial value to a variate is a good practise.

src/dist/replication/meta_server/meta_bulk_load_service.cpp
```diff
@@ -898,6 +974,65 @@ void bulk_load_service::on_partition_ingestion_reply(error_code err,
 }
 
 // ThreadPool: THREAD_POOL_META_STATE
+void bulk_load_service::remove_bulk_load_dir_on_remote_storage(int32_t app_id,
+                                                               const std::string &app_name)
+{
+    std::string bulk_load_path = get_app_bulk_load_path(app_id);
+    _meta_svc->get_meta_storage()->delete_node_recursively(
```
levy5307, line 981: maybe you can extract the same code between `remove_bulk_load_dir_on_remote_storage(int32_t app_id, const std::string &app_name)` and `remove_bulk_load_dir_on_remote_storage(std::shared_ptr<app_state> app, bool set_app_not_bulk_loading)`

src/dist/replication/meta_server/meta_bulk_load_service.cpp
```diff
@@ -898,6 +974,65 @@ void bulk_load_service::on_partition_ingestion_reply(error_code err,
...
+    erase_map_elem_by_id(app_id, _partition_bulk_load_info);
+    erase_map_elem_by_id(app_id, _partitions_total_download_progress);
+    erase_map_elem_by_id(app_id, _partitions_cleaned_up);
+    // TODO(heyuchen): add other varieties
+    _bulk_load_app_id.erase(app_id);
+    ddebug_f("reset local app({}) bulk load context", app_name);
+}
+
+// ThreadPool: THREAD_POOL_META_STATE
+void bulk_load_service::update_app_not_bulk_loading_on_remote_storage(
```
levy5307, line 1021: `update_app_bulk_load_status_on_remote_storage(std::shared_ptr<app_state> app, bool is_bulk_loading)`

hycdong: I used to write this function as your comment, but `is_bulk_loading` will always be false. `is_bulk_loading` will be true only when start bulk load in function `do_start_app_bulk_load`.

src/dist/replication/meta_server/meta_bulk_load_service.cpp
hycdong: I used to try to combine them into one function, which will make code complex, and it's difficult to extract the similar code for the callback needs different parameters.

src/dist/replication/meta_server/meta_bulk_load_service.h
```diff
@@ -43,6 +43,16 @@ struct bulk_load_info
     DEFINE_JSON_SERIALIZATION(app_id, app_name, partition_count)
 };
 
+template <typename T>
+inline void erase_map_elem_by_id(int32_t app_id, std::unordered_map<gpid, T> &mymap)
+{
+    for (auto iter = mymap.begin(); iter != mymap.end();) {
+        if (iter->first.get_app_id() == app_id) {
+            mymap.erase(iter++);
```
acelyc111, line 51: Missing `break`?

src/dist/replication/meta_server/meta_bulk_load_service.cpp
```diff
@@ -572,8 +572,84 @@ void bulk_load_service::handle_app_ingestion(const bulk_load_response &response,
 void bulk_load_service::handle_bulk_load_finish(const bulk_load_response &response,
                                                 const rpc_address &primary_addr)
 {
-    // TODO(heyuchen): TBD
-    // called by `on_partition_bulk_load_reply` when app status is succeed, failed, canceled
+    const std::string &app_name = response.app_name;
+    const gpid &pid = response.pid;
+
+    if (!response.__isset.is_group_bulk_load_context_cleaned_up) {
```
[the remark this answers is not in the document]
hycdong: It is possible, I have met this condition during test, it is okay to keep it warning even if this condition is not happened.

src/dist/replication/meta_server/meta_bulk_load_service.h
hycdong: The structure of mymap is <gpid,T>, this function is to remove all items whose `gpid.app_id == app_id`, so this function doesn't need break if remove one item.

src/dist/replication/meta_server/meta_bulk_load_service.cpp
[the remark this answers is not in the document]
levy5307: Under what conditions does `is_group_bulk_load_context_cleaned_up` may not set?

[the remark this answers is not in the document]
hycdong: In normal case, it won't happen, I meet it during kill test, I cannot remember the exact condition, but I think it's okay to keep this check here.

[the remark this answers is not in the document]
hycdong: When bulk load is canceled when some partitions have already finished, while this condition is not included this pull request.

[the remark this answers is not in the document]
hycdong: I have met this condition during test, I cannot remember the exact condition. I will continually test the code, if I confirm the condition won't happen I will update it, if not it is okay to have a check here.

src/dist/replication/meta_server/meta_bulk_load_service.h
```diff
@@ -43,6 +43,16 @@ struct bulk_load_info
     DEFINE_JSON_SERIALIZATION(app_id, app_name, partition_count)
 };
 
+template <typename T>
+inline void erase_map_elem_by_id(int32_t app_id, std::unordered_map<gpid, T> &mymap)
```
neverchanje, line 47: This is an impl function, put it in your cpp file. meta_bulk_load_service.cpp

src/dist/replication/meta_server/meta_bulk_load_service.cpp
```diff
@@ -572,8 +572,84 @@ void bulk_load_service::handle_app_ingestion(const bulk_load_response &response,
...
+                primary_addr.to_string(),
+                app_name,
+                pid);
+            return;
+        }
+    }
+
+    bool group_cleaned_up = response.is_group_bulk_load_context_cleaned_up;
+    ddebug_f("receive bulk load response from node({}) app({}) partition({}), primary status = {}, "
+             "is_group_bulk_load_context_cleaned_up = {}",
```
neverchanje, line 617: Please tell a meaningful message in the log so that we can read up what "is_group_bulk_load_context_cleaned_up" means.

src/dist/replication/meta_server/meta_bulk_load_service.cpp
```diff
@@ -572,8 +572,84 @@ void bulk_load_service::handle_app_ingestion(const bulk_load_response &response,
...
+                "receive bulk load response from node({}) app({}) partition({}), current partition "
+                "has already been cleaned up",
+                primary_addr.to_string(),
+                app_name,
+                pid);
+            return;
+        }
+    }
+
+    bool group_cleaned_up = response.is_group_bulk_load_context_cleaned_up;
```
neverchanje, line 615:
```suggestion
    // The replica group has all cleared up their bulk-load temporary files and states.
    bool group_cleaned_up = response.is_group_bulk_load_context_cleaned_up;
```

src/dist/replication/meta_server/meta_bulk_load_service.cpp
neverchanje: Why the replica group will succeed without their states cleared?

hycdong:
> Why the replica group will succeed without their states cleared?

when partition finish ingesting, its status will turn ingesting -> succeed, meta will still send `bulk_load_request` to replica, replica will turn ingesting -> succeed and cleanup its bulk load context, replica status should be `invalid` when it clean up bulk load status not succeed.

src/dist/replication/meta_server/meta_bulk_load_service.cpp
hycdong: when partition finish ingesting, its status will turn ingesting -> succeed, meta will still send `bulk_load_request` to replica, replica will turn ingesting -> succeed and cleanup its bulk load context, replica status should be invalid when it clean up bulk load status not succeed.

src/dist/replication/meta_server/meta_bulk_load_service.cpp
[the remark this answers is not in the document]
acelyc111: Whether `is_group_bulk_load_context_cleaned_up ` could be true or false depends on your code, if you're sure you have set it in all cases, I think you can use dassert instead.
Using assert instead of defensive check could be more clear and simple for logic and code. If a false assert happened, there must a bug and we have to fix it, not avoid it.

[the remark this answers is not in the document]
hycdong: I agree with you~

src/dist/replication/meta_server/meta_bulk_load_service.cpp
```diff
@@ -898,6 +971,75 @@ void bulk_load_service::on_partition_ingestion_reply(error_code err,
...
+                update_app_not_bulk_loading_on_remote_storage(std::move(app));
+            }
+        });
+}
+
+template <typename T>
+inline void erase_map_elem_by_id(int32_t app_id, std::unordered_map<gpid, T> &mymap)
+{
+    for (auto iter = mymap.begin(); iter != mymap.end();) {
+        if (iter->first.get_app_id() == app_id) {
```
levy5307, line 1004: If mymap.begin().get_app_id() != app_id, the iter will not increase in this loop

src/dist/replication/meta_server/meta_bulk_load_service.cpp
```diff
@@ -898,6 +971,75 @@ void bulk_load_service::on_partition_ingestion_reply(error_code err,
 }
 
 // ThreadPool: THREAD_POOL_META_STATE
+void bulk_load_service::remove_bulk_load_dir_on_remote_storage(int32_t app_id,
```
levy5307, line 974:
```
void bulk_load_service::remove_bulk_load_dir_on_remote_storage(int32_t app_id, const std::string &app_name, bool set_app_not_bulk_loading = false)
{
    std::string bulk_load_path = get_app_bulk_load_path(app_id);
    _meta_svc->get_meta_storage()->delete_node_recursively(
        std::move(bulk_load_path), [this, app_id, app_name, bulk_load_path]() {
            ddebug_f("remove app({}) bulk load dir {} succeed", app_name, bulk_load_path);
            reset_local_bulk_load_states(app_id, app_name);
            if (set_app_not_bulk_loading) {
                update_app_not_bulk_loading_on_remote_storage(std::move(app));
            }
        });
}
```
So you can use this function in each call

src/dist/replication/meta_server/meta_bulk_load_service.cpp
```diff
@@ -898,6 +971,77 @@ void bulk_load_service::on_partition_ingestion_reply(error_code err,
...
+        std::move(bulk_load_path), [this, app, set_app_not_bulk_loading, bulk_load_path]() {
+            ddebug_f("remove app({}) bulk load dir {} succeed", app->app_name, bulk_load_path);
+            reset_local_bulk_load_states(app->app_id, app->app_name);
+            if (set_app_not_bulk_loading) {
+                update_app_not_bulk_loading_on_remote_storage(std::move(app));
+            }
+        });
+}
+
+template <typename T>
```
levy5307, line 1000:
```
template <typename T>
inline void erase_map_elem_by_id(int32_t app_id, std::unordered_map<gpid, T> &mymap)
{
    for (auto iter = mymap.begin(); iter != mymap.end();) {
        if (iter->first.get_app_id() == app_id) {
            mymap.erase(iter);
        }        
        iter++;    
    }
}
```

PR #513, opened 2020-06-23
fix: dassert if shared log is damaged when replica server init

Shuo-Jia:
### Releated issue:
https://github.com/XiaoMi/pegasus/issues/552

We found the shared log may be incomplete after one replica_server node down and restart,  for avoid some unexpected problem,  this pr `dassert` the case.  it can help us find problem in time when error happens.

### DIFF
--- a/src/dist/replication/lib/replica_stub.cpp
+++ b/src/dist/replication/lib/replica_stub.cpp
@@ -554,7 +554,6 @@ void replica_stub::initialize(const replication_options &opts, bool clear /* = f
         dassert(lerr == ERR_OK, "restart log service must succeed");
     }
 
-    bool is_log_complete = true;
     for (auto it = rps.begin(); it != rps.end(); ++it) {
         auto err = it->second->background_sync_checkpoint();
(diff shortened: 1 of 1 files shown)


src/dist/replication/lib/replica_stub.cpp
```diff
@@ -598,30 +597,16 @@ void replica_stub::initialize(const replication_options &opts, bool clear /* = f
                smax);
 
         if (err == ERR_OK) {
-            if (smax != pmax) {
```
levy5307, line 601: I guess smax is the max decree of share log, and pmax is the max decredd of private log. Is that right?

levy5307:
So what caused this situation? Can we prevent this situation from happening, by changing some code?

src/dist/replication/lib/replica_stub.cpp
Shuo-Jia: Yeah

Shuo-Jia:
> So what caused this situation? Can we prevent this situation from happening, by changing some code?

Now, I have find the casue

PR #515, opened 2020-06-24
refactor: move task to a seperated directory

levy5307:
Code refactor. Here is the future plan: https://github.com/XiaoMi/rdsn/issues/141


acelyc111:
Is there a final target of how modules/directories distribute in the future?

levy5307:
> Is there a final target of how modules/directories distribute in the future? I think we should follow the plan to do refactors.

Yes, maybe you can have a look of https://github.com/XiaoMi/rdsn/issues/141

PR #514, opened 2020-06-24
feat(bulk-load): pause bulk load part1 - add control_bulk_load rpc and start to pause bulk load

hycdong:
This pull request is part one of pause bulk load.
- add control_bulk_load related structures
    - client -> meta server to pause, restart, cancel bulk load process
- add part of implementation of `on_control_bulk_load`
    - when app bulk load is downloading, bulk load can be paused
    - update app bulk load status to pausing, then update all partitions bulk load status to pausing (in `update_app_status_on_remote_storage_reply`)

PR #524, opened 2020-07-02
feat(bulk-load): restart bulk load

hycdong:
This pull request is about restart bulk load by client.
1. client send restart request to meta server #514 
2. meta server handle restart bulk load 
    - restart bulk load is allowed when app bulk load status is paused (in function `on_control_bulk_load`)
    - update app bulk load status from paused to downloading (in function `update_app_status_on_remote_storage_reply`)
    - update each partition status to downloading (in function `update_partition_status_on_remote_storage_reply`), and resend bulk_load_request to replica server
3. replica server handle restart bulk load
    - when replica server bulk load status is paused, request bulk load status is downloading, replica will restart bulk load from downloading (in function `do_bulk_load`), the bulk load process is restarted.

_**Special explanation**_:
1. when bulk load is paused, meta will not send bulk load request to replica server any more (#463). As a result, we should resend bulk load request to restart bulk load process, in function `update_partition_status_on_remote_storage_reply`.


src/dist/replication/lib/bulk_load/replica_bulk_loader.cpp
```diff
@@ -210,7 +210,8 @@ error_code replica_bulk_loader::do_bulk_load(const std::string &app_name,
     switch (meta_status) {
     case bulk_load_status::BLS_DOWNLOADING:
         // TODO(heyuchen): add restart downloading status check
```
acelyc111, line 212: This comment can be removed now?

hycdong: This comment could not remove now, there will be other possible status in further pull request.

src/dist/replication/meta_server/meta_bulk_load_service.cpp
```diff
@@ -1144,7 +1157,24 @@ void bulk_load_service::on_control_bulk_load(control_bulk_load_rpc rpc)
...
     case bulk_load_control_type::BLC_RESTART:
-    // TODO(heyuchen): TBD
+        if (app_status != bulk_load_status::BLS_PAUSED) {
+            derror_f(
+                "restart bulk load for app({}) failed, can not restart bulk load with status({})",
+                app_name,
+                dsn::enum_to_string(app_status));
+            response.err = ERR_INVALID_STATE;
+            response.__set_hint_msg(
+                fmt::format("app({}) status={}", app_name, dsn::enum_to_string(app_status)));
```
neverchanje, line 1167: It's never a good practice to use server-side logging for user-facing errors. The user, usually the system admin, won't dig into the logs until he falls into trouble to understand the error. Why not just set the full description in the `hint_msg`?

```suggestion
            response.err = ERR_INVALID_STATE;
            response.__set_hint_msg(
                fmt::format("restart bulk load for app({}) failed, can not restart bulk load with status({})",
                app_name, dsn::enum_to_string(app_status));
```

src/dist/replication/meta_server/meta_bulk_load_service.cpp
```diff
@@ -1144,7 +1159,26 @@ void bulk_load_service::on_control_bulk_load(control_bulk_load_rpc rpc)
...
+        if (app_status != bulk_load_status::BLS_PAUSED) {
+            derror_f(
+                "restart bulk load for app({}) failed, can not restart bulk load with status({})",
+                app_name,
+                dsn::enum_to_string(app_status));
+            response.err = ERR_INVALID_STATE;
+            response.__set_hint_msg(
+                fmt::format("can not restart bulk load for app({}) with status({})",
+                            app_name,
+                            dsn::enum_to_string(app_status)));
```
neverchanje, line 1171:
```suggestion
            auto hint_msg = fmt::format("can not restart bulk load for app({}) with status({})",
                            app_name,
                            dsn::enum_to_string(app_status)));
            derror(hint_msg.c_str());
            response.err = ERR_INVALID_STATE;
            response.__set_hint_msg(std::move(hint_msg);
```

hycdong: I think it is necessary for server to print error log of control command, I remain printing log here.

PR #525, opened 2020-07-02
build: use regex to find rocksdb depend module path

vagetablechicken:
It will be easier to update thirdparty/rocksdb, no need to change path in CMakeLists.

### DIFF
--- a/bin/dsn.cmake
+++ b/bin/dsn.cmake
@@ -296,7 +296,11 @@ function(dsn_setup_thirdparty_libs)
     set(DEFAULT_THIRDPARTY_LIBS ${THRIFT_LIB} fmt::fmt CACHE STRING "default thirdparty libs" FORCE)
 
     # rocksdb
-    list(APPEND CMAKE_MODULE_PATH "${DSN_PROJECT_DIR}/thirdparty/src/pegasus-rocksdb-6.6.4-compatible/cmake/modules")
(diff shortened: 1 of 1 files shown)


bin/dsn.cmake
```diff
@@ -296,7 +296,11 @@ function(dsn_setup_thirdparty_libs)
     set(DEFAULT_THIRDPARTY_LIBS ${THRIFT_LIB} fmt::fmt CACHE STRING "default thirdparty libs" FORCE)
 
     # rocksdb
-    list(APPEND CMAKE_MODULE_PATH "${DSN_PROJECT_DIR}/thirdparty/src/pegasus-rocksdb-6.6.4-compatible/cmake/modules")
+    file(GLOB ROCKSDB_DEPENDS_MODULE_PATH ${DSN_PROJECT_DIR}/thirdparty/src/*/cmake/modules)
```
neverchanje, line 299:
```suggestion
    file(GLOB ROCKSDB_DEPENDS_MODULE_PATH ${DSN_PROJECT_DIR}/thirdparty/src/pegasus-rocksdb-6.6.4-compatible/cmake/modules)
```

vagetablechicken: use regex to find modules, so we don't need to modify this when we change the rocksdb src name(e.g. dump version or rename)

PR #526, opened 2020-07-02
fix(scripts): change the download source of zookeeper

levy5307:
In rdsn, we download zoo-keeper from `http://git.n.xiaomi.com/pegasus/packages/raw/master/zookeeper-3.4.6.tar.gz`, but it is not a  right working version, which means if you use this zookeeper file to run unit test, a failure will occured.
So in this pull request, I change the download file to `https://github.com/XiaoMi/pegasus-common/releases/download/deps/zookeeper-3.4.6.tar.gz`


acelyc111:
@neverchanje I think we can also upload this package to Aliyun OSS where our thirdparty libs put.

scripts/linux/start_zk.sh
```diff
@@ -27,16 +27,12 @@ cd $INSTALL_DIR
 
 if [ ! -f zookeeper-3.4.6.tar.gz ]; then
     echo "Downloading zookeeper..."
-    download_url="http://git.n.xiaomi.com/pegasus/packages/raw/master/zookeeper-3.4.6.tar.gz"
-    wget -T 5 -t 1 $download_url
-    if [ $? -ne 0 ]; then
         download_url="https://github.com/XiaoMi/pegasus-common/releases/download/deps/zookeeper-3.4.6.tar.gz"
```
acelyc111, line 30: indent

neverchanje:
> @neverchanje I think we can also upload this package to Aliyun OSS where our thirdparty libs put. Sometimes github is very slow.

downloading of zk server is not quite a frequent action, so I  have no urge to this. maybe someone could help write a general zk download script.

PR #527, opened 2020-07-02
refactor: make json_helper decoupled from dsn_runtime

levy5307:


include/dsn/cpp/json_helper.h
```diff
@@ -214,11 +214,11 @@
     }
 
 #define NON_MEMBER_JSON_SERIALIZATION(type, ...)                                                   \
-    inline void json_encode(dsn::json::JsonWriter &output, const type &t)                          \
+    void json_encode(dsn::json::JsonWriter &output, const type &t)                                 \
```
neverchanje, line 217: Why remove inline?

include/dsn/cpp/json_helper.h
```diff
@@ -339,40 +339,21 @@ ENUM_TYPE_SERIALIZATION(dsn::replication::bulk_load_status::type,
...
-    json_encode(out, pid.to_string());
-}
-inline bool json_decode(const dsn::json::JsonObject &in, dsn::gpid &pid)
-{
-    std::string gpid_message;
-    dverify(json_decode(in, gpid_message));
-    return pid.parse_from(gpid_message.c_str());
-}
+void json_encode(JsonWriter &out, const dsn::gpid &pid);
+bool json_decode(const dsn::json::JsonObject &in, dsn::gpid &pid);
```
neverchanje, line 343: Move the declaration to `gpid.h`.

```suggestion
namespace json {
class JsonWriter;
class JsonObject;
} // namespace json
void json_encode(JsonWriter &out, const gpid &pid);
bool json_decode(const JsonObject &in, gpid &pid);
```

Do the same to other types.

include/dsn/cpp/json_helper.h
levy5307: In cpp, inline function must be defined in `.h`, not `.cpp`. And we should move `json_code` function of `gpid` to `gpid.cpp`.

include/dsn/cpp/json_helper.h
levy5307: Why move the declaration to `gpid.h`. I think placed them centrally in one file is ok.

neverchanje: The module itself should be responsible for its related code. There's no modularity if everything is in a central file.

PR #528, opened 2020-07-02
refactor(aio): simplify disk_engine interfaces

neverchanje:
```cpp
class disk_engine : public utils::singleton<disk_engine>
{
public:
    // asynchronous file read/write
    disk_file *open(const char *file_name, int flag, int pmode);
    error_code close(disk_file *fh);
    error_code flush(disk_file *fh);
    void read(aio_task *aio);
```

`disk_engine` was somehow designed as a middle layer between "file_io" API and "aio_provider" API. 

For example, `file::flush` forwards the call to `disk_engine::flush`, then forwards to `aio_provider::flush`.

```cpp
/*extern*/ error_code flush(disk_file *file) { return disk_engine::instance().flush(file); }
```

```cpp
error_code disk_engine::flush(disk_file *fh)
{
    if (nullptr != fh) {
        auto df = (disk_file *)fh;
        return _provider->flush(df->native_handle());
    } else {
        return ERR_INVALID_HANDLE;
    }
}
```

As a middle layer, `disk_engine` adds complexity of the code but doesn't bring abstraction above `aio_provider`. So I plan to reduce the functionalities of this class. I may eventually remove this class.


src/core/aio/disk_engine.h
```diff
@@ -70,15 +70,10 @@ class disk_file
...
-    // asynchronous file read/write
-    disk_file *open(const char *file_name, int flag, int pmode);
-    error_code close(disk_file *fh);
-    error_code flush(disk_file *fh);
-    void read(aio_task *aio);
     void write(aio_task *aio);
 
-    aio_context *prepare_aio_context(aio_task *tsk) { return _provider->prepare_aio_context(tsk); }
     service_node *node() const { return _node; }
+    static aio_provider *provider() { return instance()._provider.get(); }
```
levy5307, line 76: return type can be std::unique_ptr<aio_provider>, and std::unique_ptr is as efficient as raw pointer.

src/core/aio/file_io.cpp
```diff
@@ -52,11 +73,20 @@ namespace file {
...
 
-    disk_engine::instance().read(cb);
+    if (!cb->spec().on_aio_call.execute(task::get_current_task(), cb, true)) {
+        cb->enqueue(ERR_FILE_OPERATION_FAILED, 0);
+        return cb;
+    }
+    auto wk = file->read(cb);
+    if (wk) {
+        disk_engine::provider()->submit_aio_task(wk);
+    }
```
Shuo-Jia, line 89: I don't see the `aio_task_ptr write` was refactored, is it no need?

neverchanje: Because we have `file::write` and `file::write_vector` both using `disk_engine::write`. The write-process is much more complicated than others.  I would try this refactoring later.

PR #530, opened 2020-07-04
refactor: make simple logger decoupled from dsn_runtime

levy5307:
In `simple_logger.cpp`, function `print_header` calls `dsn::task::get_current_node_name()` and `dsn::task::get_current_worker2()>pool_spec().name.c_str()`, which are coupled with dsn_runtime.

### DIFF
--- a/include/dsn/tool-api/logging_provider.h
+++ b/include/dsn/tool-api/logging_provider.h
@@ -39,6 +39,9 @@
 #include <stdarg.h>
 
 namespace dsn {
+namespace tools {
+void set_log_prefixed_message_func(std::function<std::string()> func);
+} // namespace tools
 
 /*!
 @addtogroup tool-api-providers
--- a/include/dsn/tool-api/logging_provider.h
+++ b/include/dsn/tool-api/logging_provider.h
@@ -86,8 +86,6 @@ private:
     static logging_provider *create_default_instance();
 };
 
-namespace tools {
 void set_log_prefixed_message_func(std::function<std::string()> func);
-} // namespace tools
-
+extern std::function<std::string()> log_prefixed_message_func;
 } // namespace dsn
--- a/include/dsn/tool-api/logging_provider.h
+++ b/include/dsn/tool-api/logging_provider.h
@@ -39,10 +39,6 @@
 #include <stdarg.h>
 
 namespace dsn {
-namespace tools {
-void set_log_prefixed_message_func(std::function<std::string()> func);
-} // namespace tools
-
 /*!
 @addtogroup tool-api-providers
 @{
@@ -90,6 +86,8 @@ private:
     static logging_provider *create_default_instance();
 };
 
-/*@}*/
-// ----------------------- inline implementation ---------------------------------------
+namespace tools {
+void set_log_prefixed_message_func(std::function<std::string()> func);
+} // namespace tools
+
 } // namespace dsn
--- a/src/core/core/logging.cpp
+++ b/src/core/core/logging.cpp
@@ -50,7 +50,7 @@ static void log_on_sys_exit(::dsn::sys_exit_type)
 
 void dsn_log_init(const std::string &logging_factory_name,
                   const std::string &dir_log,
-                  std::function<std::string()> current_node_name_func)
+                  std::function<void(FILE *fp, dsn_log_level_t log_level)> print_header_func)
 {
     dsn_log_start_level =
         enum_from_string(FLAGS_logging_start_level, dsn_log_level_t::LOG_LEVEL_INVALID);
@@ -100,8 +100,8 @@ void dsn_log_init(const std::string &logging_factory_name,
 
     {
         using dsn::tools;
-        extern std::function<std::string()> node_name_func;
-        node_name_func = current_node_name_func;
+        extern std::function<std::string()> print_header;
+        node_name_func = print_header_func;
     }
 }
 
--- a/src/core/core/logging.cpp
+++ b/src/core/core/logging.cpp
@@ -41,10 +41,11 @@ DSN_DEFINE_bool("core", logging_flush_on_exit, true, "flush log when exit system
 
 namespace dsn {
 std::function<std::string()> log_prefixed_message_func = []() {
-    static thread_local std::string prefixed_message(' ', 23);
+    static thread_local std::string prefixed_message;
 
     static thread_local std::once_flag flag;
     std::call_once(flag, [&]() {
+        prefixed_message.resize(23);
         int tid = dsn::utils::get_current_tid();
         sprintf(const_cast<char *>(prefixed_message.c_str()), "unknown.io-thrd.%05d: ", tid);
     });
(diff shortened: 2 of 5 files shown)


src/core/core/service_api_c.cpp
```diff
@@ -287,6 +292,51 @@ inline void dsn_global_init()
     dsn::service_engine::instance();
 }
 
+static void print_header(FILE *fp, dsn_log_level_t log_level)
```
neverchanje, line 295:
```suggestion
static std::string dsn_log_prefixed_message() {
  ...
}
```

We should not assume the underlying logger using `fprintf(FILE*)` for printing.

levy5307: return std::string is not a good idea. If the string you get from `dsn_log_prefixed_message` is `%sab%dc`, it would occur a disater when you print it using `fprintf(fp, "%s", string);`

levy5307: fixed

src/core/tools/common/simple_logger.cpp
```diff
@@ -53,49 +53,35 @@ DSN_DEFINE_validator(stderr_start_level, [](const char *level) -> bool {
...
 });
 
+std::function<const std::string()> log_prefixed_message_func = []() {
+    int tid = ::dsn::utils::get_current_tid();
+    char res[25];
+    sprintf(res, "unknow.io-thrd.%05d: ", tid);
+    return std::string(res);
+};
+
+void set_log_prefixed_message_func(std::function<const std::string()> func)
```
neverchanje, line 63:
```suggestion
void set_log_prefixed_message_func(std::function<std::string()> func)
```

Place this function in `logging.cpp`. Because log_prefixed_message_func is not limited to simple_logger, it's shared with other implementations (though we do not have now).

src/core/tools/common/simple_logger.cpp
```diff
@@ -53,49 +53,35 @@ DSN_DEFINE_validator(stderr_start_level, [](const char *level) -> bool {
     return strcmp(level, "LOG_LEVEL_INVALID") != 0;
 });
 
+std::function<const std::string()> log_prefixed_message_func = []() {
```
neverchanje, line 56:
```suggestion
std::function<std::string()> log_prefixed_message_func = []() {
```

src/core/tools/common/simple_logger.h
```diff
@@ -33,6 +33,8 @@
 namespace dsn {
 namespace tools {
 
+void set_log_prefixed_message_func(std::function<const std::string()> func);
```
neverchanje, line 36:
```suggestion
void set_log_prefixed_message_func(std::function<std::string()> func);
```

Place this function in `logging_provider.h`. Because `log_prefixed_message_func` is not limited to `simple_logger`, it's shared with other implementations (though we do not have now).

include/dsn/tool-api/logging_provider.h
```diff
@@ -87,6 +86,8 @@ class logging_provider
     static logging_provider *create_default_instance();
 };
 
-/*@}*/
-// ----------------------- inline implementation ---------------------------------------
+namespace tools {
+void set_log_prefixed_message_func(std::function<std::string()> func);
+} // namespace tools
+
```
neverchanje, line 92:
```suggestion
void set_log_prefixed_message_func(std::function<std::string()> func);
```

No need to keep the meaningless `namespace tools`.

src/core/core/service_api_c.cpp
```diff
@@ -287,6 +291,43 @@ inline void dsn_global_init()
     dsn::service_engine::instance();
 }
 
+static const std::string dsn_log_prefixed_message_func()
```
neverchanje, line 294: Why returning a constant value? Didn't your IDE warn you that is redundant?

src/core/tools/common/simple_logger.cpp
```diff
@@ -53,49 +53,30 @@ DSN_DEFINE_validator(stderr_start_level, [](const char *level) -> bool {
     return strcmp(level, "LOG_LEVEL_INVALID") != 0;
 });
 
+std::function<std::string()> log_prefixed_message_func = []() {
+    int tid = dsn::utils::get_current_tid();
+    char res[25];
+    sprintf(res, "unknow.io-thrd.%05d: ", tid);
```
neverchanje, line 59:
```suggestion
    sprintf(res, "unknown.io-thrd.%05d: ", tid);
```

src/core/core/logging.cpp
```diff
@@ -39,13 +39,25 @@ DSN_DEFINE_string("core",
 
 DSN_DEFINE_bool("core", logging_flush_on_exit, true, "flush log when exit system");
 
+namespace dsn {
+namespace tools {
+void set_log_prefixed_message_func(std::function<std::string()> func)
+{
+    extern std::function<std::string()> log_prefixed_message_func;
```
neverchanje, line 46: It's unconventional to declare an extern var inside a function body.

src/core/core/logging.cpp
```diff
@@ -39,13 +39,29 @@ DSN_DEFINE_string("core",
 
 DSN_DEFINE_bool("core", logging_flush_on_exit, true, "flush log when exit system");
 
+namespace dsn {
+std::function<std::string()> log_prefixed_message_func = []() {
+    int tid = dsn::utils::get_current_tid();
+    char res[25];
+    sprintf(res, "unknown.io-thrd.%05d: ", tid);
+    return std::string(res);
```
acelyc111, line 47: Could you use a static thread_local variable for this prefix? It's not neede to create the string every time.

src/core/core/service_api_c.cpp
```diff
@@ -287,6 +291,43 @@ inline void dsn_global_init()
...
+                    "%6s.%7s%u: ",
+                    dsn::task::get_current_node_name(),
+                    dsn::task::get_current_worker2()->pool_spec().name.c_str(),
+                    dsn::task::get_current_worker2()->index());
+        } else {
+            sprintf(res, "%6s.%7s.%05d: ", dsn::task::get_current_node_name(), "io-thrd", tid);
+        }
+    }
+
+    return res;
```
acelyc111, line 328: The same, use static thread_local variable can save much time for log.

src/core/core/logging.cpp
```diff
@@ -39,13 +39,29 @@ DSN_DEFINE_string("core",
 
 DSN_DEFINE_bool("core", logging_flush_on_exit, true, "flush log when exit system");
 
+namespace dsn {
+std::function<std::string()> log_prefixed_message_func = []() {
+    static thread_local std::string prefixed_message(' ', 23);
+    int tid = dsn::utils::get_current_tid();
+    sprintf(const_cast<char *>(prefixed_message.c_str()), "unknown.io-thrd.%05d: ", tid);
```
acelyc111, line 46: Thoungh prefixed_message is a thread_local type, but these two lines will execute for every calls, you can use a std::call_once here.

levy5307: good idea

PR #533, opened 2020-07-07
refactor: move aio module from src/core/aio to src/aio

levy5307:
Code refactor. Here is the future plan: https://github.com/XiaoMi/rdsn/issues/141


src/aio/tests/command.txt
```diff
@@ -0,0 +1,10 @@
+
```
acelyc111, line 1: What does this file used for?

levy5307: It's used for copy data from `command.txt` to `command.copy.txt` in `aio.dsn_file` test,. And the content in `command.txt` in meaningless.

hycdong: I recommend rename this file name and add comments in unit test to show what is the file used for and its content is meaningless.

levy5307: Yes, you are right. But it is not my main purpose in this pr. Maybe we can do it in later pull request.

PR #534, opened 2020-07-08
feat(bulk-load): replica handle errors during bulk load

hycdong:
As previous pull requests shows, bulk load process is like:
1. meta sends bulk load request to primary replica to let primary replica executing bulk load (#457)
2. primary replica broadcasts group_bulk_load_request to secondaries to let secondary replicas executing bulk load (#460)
3. secondary replica reports its bulk_load_state through group_bulk_load_response (#479)
4. primary replica reports group bulk_load_state through bulk_load_response (#479)

This pull request is about how replica handle errors during bulk load process, including two parts:
1. primary meet errors while receiving `group_bulk_load_response`
    - primary stores secondary bulk load state in primary context, when meet errors, primary should reset this secondary's state (in function `on_group_bulk_load_reply` and `reset_node_bulk_load_states`)
2. replica partition status changes during bulk load
    - if primary changes, primary should clean up its primary bulk load states in primary context (#465 #479 #496)
    - if other status switches, bulk load states should also be cleared (in function `update_local_configuration` and `clear_bulk_load_states_if_needed`)

### DIFF
--- a/src/dist/replication/lib/replica_config.cpp
+++ b/src/dist/replication/lib/replica_config.cpp
@@ -38,6 +38,7 @@
 #include "mutation.h"
 #include "mutation_log.h"
 #include "replica_stub.h"
+#include "bulk_load/replica_bulk_loader.h"
 #include <dsn/dist/fmt_logging.h>
 #include <dsn/dist/replication/replication_app_base.h>
 #include <dsn/utility/fail_point.h>
@@ -772,6 +773,7 @@ bool replica::update_local_configuration(const replica_configuration &config,
             break;
         case partition_status::PS_INACTIVE:
             _primary_states.cleanup(old_ballot != config.ballot);
+            _bulk_loader->clear_bulk_load_states();
             // here we use wheather ballot changes and wheather disconnecting with meta to
             // distinguish different case above mentioned
             if (old_ballot == config.ballot && _stub->is_connected()) {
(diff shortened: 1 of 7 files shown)


src/dist/replication/lib/replica_config.cpp
```diff
@@ -772,6 +773,7 @@ bool replica::update_local_configuration(const replica_configuration &config,
             break;
         case partition_status::PS_INACTIVE:
             _primary_states.cleanup(old_ballot != config.ballot);
+            _bulk_loader->clear_bulk_load_states();
```
neverchanje, line 776: If possible, reduce the `_bulk_loader`-related code in class `replica`, which is bloated as more functions added.

A feasible way is:

```
    _bulk_loader->clear_bulk_load_states_if_needed(); // Add here.

    switch (old_status) {
    case partition_status::PS_PRIMARY:
        cleanup_preparing_mutations(false);
        switch (config.status) {
        case partition_status::PS_PRIMARY:
            replay_prepare_list();
            break;
        case partition_status::PS_INACTIVE:
            _primary_states.cleanup(old_ballot != config.ballot);
```

```cpp
void replica_bulk_loader::clear_bulk_load_states_if_needed(old_status) {
    switch (old_status) {
    case partition_status::PS_PRIMARY:
        switch (config.status) {
        case partition_status::PS_INACTIVE:
        case partition_status::PS_ERROR:
            clear_bulk_load_states();
            break;
    case partition_status::PS_SECONDARY:
            clear_bulk_load_states();
            break;         
    }
}
```

PR #535, opened 2020-07-08
feat(bulk-load): handle recoverable errors during bulk load

hycdong:
When bulk load meet recoverable errors, such as rpc errors, replica status changes, meta server will rollback to downloading, this pull request is about how meta and replica handle this condition.
1. meta server will call function `try_rollback_to_downloading`
    - update app status to downloading and partition status to downloading (#524)
    - this function will be called by `on_partition_bulk_load_reply` and `on_partition_ingestion_reply` when meeting recoverable errors (#463 #493)
    - when bulk load is under normal stage (downloading, downloaded, ingesting, succeed), will rollback to downloading, for other status, bulk load process is supposed to be stopped or is considered as failed, it should not rollback to downloading to retry bulk load
2. replica server handle it through `do_bulk_load`
    - when request status is downloading, local is ingesting or succeed, will restart downloading calling function `start_download`
    - we needn't to handle downloading and downloaded status here. If replica status changes, bulk load status will be cleared and local status is invalid (#534). If replica status doesn't change, replica is still downloading or downloaded, report download progress, needn't restart downloading


src/dist/replication/meta_server/meta_bulk_load_service.cpp
```diff
@@ -705,8 +705,19 @@ void bulk_load_service::handle_app_pausing(const bulk_load_response &response,
...
+    zauto_read_lock l(_lock);
+    const auto app_status = get_app_bulk_load_status_unlocked(pid.get_app_id());
+    if (app_status == bulk_load_status::BLS_DOWNLOADING ||
+        app_status == bulk_load_status::BLS_DOWNLOADED ||
+        app_status == bulk_load_status::BLS_INGESTING ||
+        app_status == bulk_load_status::BLS_SUCCEED) {
+        update_app_status_on_remote_storage_unlocked(pid.get_app_id(),
+                                                     bulk_load_status::type::BLS_DOWNLOADING);
+    } else {
+        ddebug_f("app({}) status={}, no need to rollback to downloading, wait for next round",
```
levy5307, line 717: Why does here should waiting for next round?

hycdong: when bulk load is under normal stage (downloading, downloaded, ingesting, succeed), will rollback to downloading, for other status, bulk load process is supposed to be stopped or is considered as failed, it should not rollback to downloading to retry bulk load.

PR #538, opened 2020-07-08
refactor: move nfs module from src/dist/nfs to src/nfs

levy5307:
Code refactor. Here is the future plan: #141


hycdong:
nfs_types.h and nfs_types.cpp are auto-generated by thrift compiler, you should also change its path in file `compile_thrift.py`, otherwise, we should move files manually when update thrift structure.

levy5307:
> `nfs_types.h` and `nfs_types.cpp` are auto-generated by thrift compiler, you should also change its path in file `compile_thrift.py`, otherwise, we should move files manually when updating thrift structure.

I have changed it. Please have a look

PR #543, opened 2020-07-09
feat(bulk-load): replica validate status

hycdong:
When meta sends bulk_load_request, replica will compare meta bulk load status and local bulk load status, function `validate_bulk_load_status` will do this check, it is call by `do_bulk_load`, if not pass validation, replica will set ERR_INVALID_STATE in bulk_load_response.


src/dist/replication/lib/bulk_load/replica_bulk_loader.cpp
```diff
@@ -274,8 +274,50 @@ error_code replica_bulk_loader::do_bulk_load(const std::string &app_name,
 error_code replica_bulk_loader::validate_bulk_load_status(bulk_load_status::type meta_status,
```
neverchanje, line 274: This function seems pure static. You should declare it static to make it clear.

```suggestion
/*static*/ error_code replica_bulk_loader::validate_bulk_load_status(bulk_load_status::type meta_status,
```

levy5307, line 274: `const bulk_load_status::type meta_status`
`const bulk_load_status::type local_status`

src/dist/replication/lib/bulk_load/test/replica_bulk_loader_test.cpp
```diff
@@ -153,6 +153,11 @@ class replica_bulk_loader_test : public replica_test_base
         _bulk_loader->on_group_bulk_load_reply(rpc_error, _group_req, resp);
     }
 
+    bool validate_status(bulk_load_status::type meta_status, bulk_load_status::type local_status)
+    {
+        return _bulk_loader->validate_bulk_load_status(meta_status, local_status) == ERR_OK;
```
neverchanje, line 158:
```suggestion
        return replica_bulk_loader::validate_bulk_load_status(meta_status, local_status) == ERR_OK;
```

PR #545, opened 2020-07-09
refactor: make throttle controller decoupled from dsn_runtime

levy5307:
class rpc_message, which is a type in dsn_runtime, is used in throttle controller. So throttle controller is coupled with dsn_runtime now.
In this pull request, I want to decouple throttle controller from dsn_runtime,  so we can move throttle controller to utils in later pull requests.


vagetablechicken:
I'm confused with the title 'decouple from dsn_runtime', maybe 'cleanup argument list' is better?

levy5307:
> I'm confused with the title 'decouple from dsn_runtime', maybe 'cleanup argument list' is better?

I don't think so, because my main purpose is to do decouple work.

src/dist/replication/lib/throttling_controller.h
```diff
@@ -83,7 +81,7 @@ class throttling_controller
     // do throttling control, return throttling type.
     // 'delay_ms' is set when the return type is not PASS.
     throttling_type
-    control(const message_ex *request, int32_t request_units, /*out*/ int64_t &delay_ms);
+    control(const int64_t client_timeout, int32_t request_units, /*out*/ int64_t &delay_ms);
```
hycdong, line 84: I suggest using `client_timeout_ms`

levy5307: good idea.

PR #548, opened 2020-07-10
feat(bulk-load): small update for meta bulk load service

hycdong:
This pull request is about some small update for meta bulk load service:
1. update function `initialize_bulk_load_service` and `on_start_bulk_load` in THREAD_POOL_META_SERVER
    - most meta functions are executing in thread pool THREAD_POOL_META_STATE, which is partitioned with only one worker, only some control functions or functions used for starting meta are executing in thread pool THREAD_POOL_META_SERVER
    - function `initialize_bulk_load_service` is called by `start_service`, it will create bulk load root directory on remote storage and try to get possible bulk load info from it synchronously, operating remote storage data should be executing in THREAD_POOL_META_STATE. if initialize_bulk_load_service executes in it, it will lead dead lock, because `create_bulk_load_root_dir` is a synchronous function.
    - function `on_start_bulk_load` will check parameters by reading files on remote file provider(fds), it should not executing in THREAD_POOL_META_STATE
2. fix thread pool in function `partition_ingestion`
    -  function `partition_ingestion` is executing in THREAD_POOL_DEFAULT, while function `handle_bulk_load_failed` and `handle_app_unavailable` should be executed in THREAD_POOL_META_STATE
3. fix bugs in function `on_control_bulk_load`
    - remove useless app lock
    - call function update_app_status_on_remote_storage_unlocked directly instead of adding it into task queue

Special explanation:
In pull request #490 , I update simple_kv test case 402, this pull request I recover it into previous state.


src/dist/replication/meta_server/meta_bulk_load_service.cpp
```diff
@@ -16,21 +16,19 @@ bulk_load_service::bulk_load_service(meta_service *meta_svc, const std::string &
     _state = _meta_svc->get_server_state();
 }
 
-// ThreadPool: THREAD_POOL_META_STATE
+// ThreadPool: THREAD_POOL_META_SERVER
 void bulk_load_service::initialize_bulk_load_service()
 {
     task_tracker tracker;
-    error_code err = ERR_OK;
+    _sync_bulk_load_storage.reset(new mss::meta_storage(_meta_svc->get_remote_storage(), &tracker));
```
neverchanje, line 23: Do you intend to set `_sync_bulk_load_storage` to a temporary tracker?

levy5307, line 23: make_unique

src/dist/replication/meta_server/meta_bulk_load_service.h
```diff
@@ -306,11 +306,14 @@ class bulk_load_service
         return (_bulk_load_app_id.find(app_id) != _bulk_load_app_id.end());
     }
 
+    mss::meta_storage *get_sync_bulk_load_storage() const { return _sync_bulk_load_storage.get(); }
+
```
levy5307, line 310: return `mss::meta_storage &` is a little better

src/dist/replication/meta_server/meta_bulk_load_service.cpp
hycdong: Yes, `_sync_bulk_load_storage` will only be used in functions for sync bulk load info from remote storage, other operations still use `_meta_svc->get_meta_storage()` and track by meta service tracker.

src/dist/replication/meta_server/meta_bulk_load_service.h
hycdong: `get_sync_bulk_load_storage` is a redundant function, I remove it.

src/dist/replication/meta_server/meta_bulk_load_service.cpp
neverchanje: OK. Then I suggest making it a local variable, since as you implied, this variable only use in this function (`initialize_bulk_load_service`).

hycdong: It will be used in functions for sync bulk load info, you can see function `create_bulk_load_root_dir`, it will also be used in `sync_apps_bulk_load_from_remote_storage`. If I use it as a local variable, it should be pass to those functions.

hycdong: I remove `_sync_bulk_load_storage` in this pull request, I will update it in next pull request.

PR #555, opened 2020-07-13
refactor: move replica module from src/dist/replication/lib to src/replica

levy5307:
Code refactor. Here is the future plan: #141


src/replica/storage/simple_kv/tests/case.cpp
```diff
@@ -39,9 +39,15 @@
 
 #include <dsn/tool-api/task.h>
 #include <dsn/tool-api/rpc_message.h>
+<<<<<<< HEAD:src/replica/storage/simple_kv/tests/case.cpp
```
acelyc111, line 42: Conflicts need to be resolved.

src/replica/storage/simple_kv/tests/checker.cpp
```diff
@@ -41,13 +41,23 @@
 #include "checker.h"
 #include "case.h"
 
+<<<<<<< HEAD:src/replica/storage/simple_kv/tests/checker.cpp
```
acelyc111, line 44: Same

src/replica/storage/simple_kv/tests/case.cpp
levy5307: I am sorry I didn't found it yesterday. fixed

PR #553, opened 2020-07-13
feat(bulk-load): sync bulk load from remote storage

hycdong:
This pull request is about sync bulk load info from remote storage while starting meta server.
- add `_sync_bulk_load_storage` to simplify code
- add code for sync bulk load info from remote storage
- add unit test

PR #559, opened 2020-07-14
feat(bulk-load): continue bulk load while meta server start part1

hycdong:
When meta server crash during bulk load, meta server will sync bulk load info from remote storage (#553), then try to continue bulk load process, this pull request is about validation before continue bulk load.


src/meta/meta_bulk_load_service.cpp
```diff
@@ -1349,9 +1349,197 @@ void bulk_load_service::do_sync_partition(const gpid &pid, std::string &partitio
...
+                     different_status_count,
+                     ainfo.partition_count - pinfo_map.size());
+            is_valid = false;
+        }
+        break;
+    case bulk_load_status::BLS_DOWNLOADED:
+    case bulk_load_status::BLS_INGESTING: {
+        // if app status is downloaded, valid partition status is downloaded or ingesting
+        // if app status is ingesting, valid partition status is ingesting or succeed
+        const auto valid_status = (app_status == bulk_load_status::BLS_DOWNLOADED)
```
acelyc111, line 1496: The same status with `app_status ` is valid too, how about rename `valid_status ` to `other_ valid_status `?

src/meta/meta_bulk_load_service.cpp
```diff
@@ -1349,9 +1349,197 @@ void bulk_load_service::do_sync_partition(const gpid &pid, std::string &partitio
...
+                         dsn::enum_to_string(app_status),
+                         dsn::enum_to_string(valid_status));
+                is_valid = false;
+                break;
+            }
+        }
+    } break;
+    case bulk_load_status::BLS_SUCCEED:
+    case bulk_load_status::BLS_PAUSED:
+        // if app status is succeed and paused, partition status should not be different from app
```
acelyc111, line 1516:
```suggestion
        // if app status is succeed or paused, all partitions' status should not be different from app's
```

src/meta/meta_bulk_load_service.cpp
```diff
@@ -1349,9 +1349,197 @@ void bulk_load_service::do_sync_partition(const gpid &pid, std::string &partitio
...
 
+// ThreadPool: THREAD_POOL_META_SERVER
 void bulk_load_service::try_to_continue_bulk_load()
+{
+    FAIL_POINT_INJECT_F("meta_try_to_continue_bulk_load", [](dsn::string_view) {});
+    for (const auto app_id : _bulk_load_app_id) {
+        app_bulk_load_info ainfo = _app_bulk_load_info[app_id];
+        // <partition_index, partition_bulk_load_info>
+        std::unordered_map<int32_t, partition_bulk_load_info> pinfo_map;
+        for (const auto kv : _partition_bulk_load_info) {
```
levy5307, line 1360: `const auto &kv`

src/meta/meta_bulk_load_service.cpp
```diff
@@ -1349,9 +1349,197 @@ void bulk_load_service::do_sync_partition(const gpid &pid, std::string &partitio
     });
 }
 
+// ThreadPool: THREAD_POOL_META_SERVER
 void bulk_load_service::try_to_continue_bulk_load()
+{
+    FAIL_POINT_INJECT_F("meta_try_to_continue_bulk_load", [](dsn::string_view) {});
+    for (const auto app_id : _bulk_load_app_id) {
+        app_bulk_load_info ainfo = _app_bulk_load_info[app_id];
```
levy5307, line 1357: Do we need to judge whether the app_id is contained in _app_bulk_load_info or not?

src/meta/meta_bulk_load_service.cpp
```diff
@@ -1349,9 +1349,197 @@ void bulk_load_service::do_sync_partition(const gpid &pid, std::string &partitio
...
+                 "partition_count = {}, current app_id = {}, partition_count = {}",
+                 ainfo.app_name,
+                 ainfo.app_id,
+                 ainfo.partition_count,
+                 app_id,
+                 partition_count);
+        return false;
+    }
+
+    // partition_bulk_load_info_size should not be greater than partition_count
```
levy5307, line 1438: partition_count should not be greater than partition_count

src/meta/meta_bulk_load_service.cpp
hycdong: `_bulk_load_app_id` and `_app_bulk_load_info` are sync from remote storage at the same time, you can see function `do_sync_app`.

PR #560, opened 2020-07-14
fix: cleaup lock in ut abnormal_api_call

vagetablechicken:
If we don't unlock [lock_id, my_id2], my_id2 will be kept in zookeeper until the session expired. It should be cleaned after test finished.


src/zookeeper/test/CMakeLists.txt
```diff
@@ -23,7 +23,6 @@ set(MY_BINPLACES
     "${CMAKE_CURRENT_SOURCE_DIR}/run.sh"
     "${CMAKE_CURRENT_SOURCE_DIR}/clear.sh"
     "${CMAKE_CURRENT_SOURCE_DIR}/config-test.ini"
-    "${CMAKE_CURRENT_SOURCE_DIR}/gtest.filter"
-)
+        )
```
acelyc111, line 26: indent

vagetablechicken: fix

neverchanje:
I can see this is a tiny fix on the test, to free up the resources after the test. Does it bring any other benefits? @vagetablechicken

vagetablechicken:
> I can see this is a tiny fix on the test, to free up the resources after the test. Does it bring any other benefits? @vagetablechicken

We were confused why ut `simple_lock_unlock` was running separately. We can't find the reason. So I check the code to ensure that we could run these tests together. The only problem of this ut is: if the lock doesn't unlock and you run it again quickly, the first lock op will wait too long.
So it contains a fix and script cleanup.

PR #562, opened 2020-07-15
refactor: use DSN_REGISTER_COMPONENT_PROVIDER to register logger

levy5307:
1. use DSN_REGISTER_COMPONENT_PROVIDER to register simple_logger and screen_logger
2. some small refactor for optimization

### DIFF
--- a/src/core/core/coredump.posix.cpp
+++ b/src/core/core/coredump.posix.cpp
@@ -39,6 +39,7 @@
 #include <dsn/tool_api.h>
 #include <sys/types.h>
 #include <signal.h>
+#include <dsn/c/app_model.h>
 
 namespace dsn {
 namespace utils {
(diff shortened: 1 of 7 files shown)


src/core/core/coredump.posix.cpp
```diff
@@ -39,6 +39,7 @@
 #include <dsn/tool_api.h>
 #include <sys/types.h>
 #include <signal.h>
+#include <dsn/c/app_model.h>
```
hycdong, line 42: Why include this file?

levy5307: When I removed the include file of <dsn/service_api_c.h> in include/dsn/tool-api/logging_provider.h, the complier tells me that it can't find `dsn_exit` in `coredump.posix.cpp`

PR #563, opened 2020-07-16
feat(bulk-load): continue bulk load while meta server start part2

hycdong:
When meta server crash during bulk load, meta server will sync bulk load info from remote storage (#553), then do validation before continue bulk load(#559), this pull request will continue app bulk load process.

PR #564, opened 2020-07-17
feat(bulk-load): check app bulk load states while sync app from remote storage

hycdong:
The previous pull requests(#553 #559 #563) show how to continue bulk load after meta server start or restarts, which is based on the assumption that the app is executing bulk load process before. However, we should check whether the app is executing bulk load or not  firstly. This pull request will do this check(in function `check_app_bulk_load_states`).

Before meta server really starts, meta server will sync app info and all partition info from remote storage, when an app and its all partitions' info synced succeed, call function `check_app_bulk_load_states`. This function will check consistency between app bulk_loading flag and app bulk load directory.
- the app is executing bulk load should have bulk load directory on remote storage and its `is_bulk_loading` should be true. 
- `is_bulk_loading` is true, it is expected that bulk load directory existed on remote storage, if it is not existed, should set `is_bulk_loading` as false. 
- `is_bulk_loading` is false, it is expected that bulk load directory not existed on remote storage, if it exists, should remove useless bulk load directory on remote storage.

Special explanation:
I don't use `meta_storage` utility in this function. For one hand, `check_app_bulk_load_states` should execute in thread pool `THREAD_POOL_META_SERVER`, because meta server is not started. For another hand, this function use `node_exist` interface, this function callback is not an universal case, node not exists is not a wrong case.


src/meta/meta_bulk_load_service.cpp
```diff
@@ -1661,5 +1661,48 @@ void bulk_load_service::create_missing_partition_dir(const std::string &app_name
...
+                tasking::enqueue(LPC_META_CALLBACK,
+                                 nullptr,
+                                 std::bind(&bulk_load_service::check_app_bulk_load_states,
+                                           this,
+                                           app,
+                                           is_app_bulk_loading),
+                                 0,
+                                 std::chrono::seconds(1));
+                return;
+            } else {
```
acelyc111, line 1683: Now that you have returned in the previous condition, you can remove `else`, it will be more friendly to read with less indents.

src/meta/meta_bulk_load_service.cpp
```diff
@@ -1661,5 +1661,48 @@ void bulk_load_service::create_missing_partition_dir(const std::string &app_name
...
+                if (err != ERR_OK && is_app_bulk_loading) {
```
acelyc111, line 1684: So there isn't a special return code to indicate the node is not exist?

src/meta/test/meta_bulk_load_service_test.cpp
```diff
@@ -283,6 +283,11 @@ class bulk_load_service_test : public meta_test_base
         return count;
     }
 
+    bool app_bulk_load_states_reset(int32_t app_id)
```
levy5307, line 286: the name of `app_bulk_load_states_reset`ave an ambiguity, I think `is_app_bulk_load_states_reset` is better

PR #565, opened 2020-07-18
feat(bulk-load): add bulk load ddl client interface

hycdong:
The previous pull requests support functions start bulk load(#454), control bulk load(#514) and query bulk load process(#556), this pull request add related ddl client interface of them.


src/client/replication_ddl_client.cpp
```diff
@@ -1547,5 +1549,36 @@ void replication_ddl_client::query_disk_info(
...
+    auto req = make_unique<start_bulk_load_request>();
+    req->app_name = app_name;
+    req->cluster_name = cluster_name;
+    req->file_provider_type = file_provider_type;
+    return call_rpc_sync(start_bulk_load_rpc(std::move(req), RPC_CM_START_BULK_LOAD));
+}
+
+error_with<control_bulk_load_response>
+replication_ddl_client::control_bulk_load(const std::string &app_name,
+                                          bulk_load_control_type::type control_type)
```
levy5307, line 1566: `const bulk_load_control_type::type control_type`

src/client/replication_ddl_client.cpp
```diff
@@ -1547,5 +1549,36 @@ void replication_ddl_client::query_disk_info(
...
+    req->type = control_type;
+    return call_rpc_sync(control_bulk_load_rpc(std::move(req), RPC_CM_CONTROL_BULK_LOAD));
+}
+
+error_with<query_bulk_load_response>
+replication_ddl_client::query_bulk_load(const std::string &app_name)
+{
+
+    auto req = make_unique<query_bulk_load_request>();
+    req->app_name = app_name;
```
levy5307, line 1579: `__set_app_name`

hycdong: I think `app_name` is a required filed, it is okay to set its value directly.

src/client/replication_ddl_client.cpp
```diff
@@ -1547,5 +1549,36 @@ void replication_ddl_client::query_disk_info(
...
 }
 
+error_with<start_bulk_load_response>
+replication_ddl_client::start_bulk_load(const std::string &app_name,
+                                        const std::string &cluster_name,
+                                        const std::string &file_provider_type)
+{
+    auto req = make_unique<start_bulk_load_request>();
+    req->app_name = app_name;
+    req->cluster_name = cluster_name;
```
acelyc111, line 1559: When connect to a cluster to start bulk load, the cluster name is definite, so why specify the cluster name?

src/client/replication_ddl_client.cpp
```diff
@@ -1547,5 +1549,36 @@ void replication_ddl_client::query_disk_info(
...
+    req->file_provider_type = file_provider_type;
+    return call_rpc_sync(start_bulk_load_rpc(std::move(req), RPC_CM_START_BULK_LOAD));
+}
+
+error_with<control_bulk_load_response>
+replication_ddl_client::control_bulk_load(const std::string &app_name,
```
acelyc111, line 1565: Which operations does 'control' include, dose it include 'start'?

src/client/replication_ddl_client.cpp
hycdong: The cluster name is used to find sst files on file provider(fds), the files are under the path: `<bulk_load_provider_root>/<cluster_name>/<app_name>`

src/client/replication_ddl_client.cpp
hycdong: pause, restart, cancel and force cancel, `start` is not included.

src/client/replication_ddl_client.cpp
levy5307: `__set_app_name` will always right, but it is wrong to use `app_name` directly when someone change it to optional.

hycdong: If you would like to update a required filed into optional, you should also update the related code. In my view, `__set_app_name` is used for optional filed, the programmer who reads the code will figure out which filed is optional or required easily.

PR #566, opened 2020-07-20
refactor: move runtime module from src/core/core to src/runtime

levy5307:
Code refactor. For more details: #141


src/core/CMakeLists.txt
```diff
@@ -1,17 +0,0 @@
-add_subdirectory(core)
```
acelyc111, line 1: Now the core directory has been completely removed?

levy5307: Yes.

PR #569, opened 2020-07-22
perf: implement thread-pool-based asynchrounous IO

Shuo-Jia:


neverchanje:
As https://github.com/XiaoMi/rdsn/pull/633 is an improvement version upon this PR. We no longer need to keep this PR here.

PR #570, opened 2020-07-23
fix(bulk-load): update cleanup download task

hycdong:
This pull request fix a bug while cleanup download task.
It used to use `CLEANUP_TASK_ALWAYS` marco to cleanup download task, it will cancel task without waiting, the task must be not running at this time. However, when pause or cancel bulk load, it is possible that downloading task is still running, this pull request fix this problem.


src/replica/bulk_load/replica_bulk_loader.cpp
```diff
@@ -590,12 +590,15 @@ error_code replica_bulk_loader::remove_local_bulk_load_dir(const std::string &bu
 }
 
 // ThreadPool: THREAD_POOL_REPLICATION
-void replica_bulk_loader::cleanup_download_task()
+bool replica_bulk_loader::cleanup_download_task()
 {
```
levy5307, line 594: why did you and a return type here?

src/replica/bulk_load/replica_bulk_loader.cpp
```diff
@@ -590,12 +590,15 @@ error_code replica_bulk_loader::remove_local_bulk_load_dir(const std::string &bu
 }
 
 // ThreadPool: THREAD_POOL_REPLICATION
-void replica_bulk_loader::cleanup_download_task()
+bool replica_bulk_loader::cleanup_download_task()
```
neverchanje, line 593: 1. The returned value is never used.
2. `cleanup_download_task` is running in REPLICATION thread pool, it can't be blocked, otherwise critical tasks will be blocked too.
3. Do we need to loop over to wait when the returned value of `CLEANUP_TASK` is false?

src/replica/bulk_load/replica_bulk_loader.cpp
hycdong: https://github.com/XiaoMi/rdsn/blob/9d1059eb6597efbfb00b405c867d9b0edd82e0ec/src/replica/replica_context.h#L53-L63

CLEANUP_TASK marco will return value, it should only be used in the function whose return value is bool.

src/replica/bulk_load/replica_bulk_loader.cpp
hycdong: This `CLEANUP_TASK` is widely used in `replica_context.cpp`, context switch `cleanup` functions. You can see `cancel` function in `task.cpp`, if it cost too long to cancel the function, it will assert. It is the answer of Q2.
In this case, `CLEANUP_TASK` will not return false, so the returned value is not used. I just would like use the marco.

neverchanje: If you ensure the returned value is always true, keep that `dassert` there.

I'd more like to see a program dies because of the bad assumption (such as the assert here) rather than the program being slow due to an unknown reason.

hycdong: Well, I think you misunderstand my view. 
CLEANUP_TASK_ALWAYS won't wait while cancel the task, CLEANUP_TASK provide `force` parameter, if `force = true`, cancel task will wait for this task finished, otherwise not. 
when cancel downloading task, I would like to wait for task finished, so I should use CLEANUP_TASK and set `force = true`. You can see the cancel function below:

https://github.com/XiaoMi/rdsn/blob/113e7e601459d6165b448f731a7f7aa8bcf26f2b/src/runtime/task/task.cpp#L310-L369

When `wait_until_finished = true`, and current task is not task who will be canceled, if cancel cost too long, it will assert on L335, otherwise, `finished` will always be true.

```
#define CLEANUP_TASK(task_, force)                                                                 
     {                                                                                              
         task_ptr t = task_;                                                                        
         if (t != nullptr) {                                                                        
             bool finished;                                                                         
             t->cancel(force, &finished);                                                           
             if (!finished && !dsn_task_is_running_inside(task_.get()))                              
                 return false;                                                                      
             task_ = nullptr;                                                                       
         }                                                                                          
     }                     
```
As a result, `finished` will always be true, CLEANUP_TASK will not return false.

neverchanje: >  if cancel cost too long, it will assert on L335, otherwise, finished will always be true.

Yes. If cancellation waits for more than 0xFFFFFFF ms (this number sucks), it asserts. But this means the cancellation will block for a very long time. 0xFFFFFFF ms = 268,435 seconds!

hycdong: It is the worst case, actually it will not cost such a long time.
When a task is created, its states is READY, while it is executed, its state is RUNNING, and its state is FINISH when it is finished. From the code in my previous comment, only the task whose state is RUNNING should can function `wait_on_cancel`. Each downloading task only download one file, run in threadpool `THREAD_POOL_REPLICATION_LONG` whose worker_count is 2. For each partition, there are only at most downloading tasks is running, it won't cost such a long time.
Besides, clean up downloading tasks will only be used when downloading succeed, pause bulk load and cleanup context, tasks are usually be finished in most cases.

hycdong: I have updated the code, cancel tasks without waiting for it.

PR #572, opened 2020-07-29
fix(bulk-load): fix bug while rolling back to downloading

hycdong:
This pull request adds `_apps_rolling_back` to fix bug while rolling back to downloading.
We assume that the app who is executing bulk load has four partitions, and is during downloaded stage, partition[0] status is ingesting, partition[1], partition[2] and partition[3] are still downloaded, are trying to update status to ingesting on remote storage. At this time, something wrong with partition[2], meta server will try to rollback to downloading.
partition[2] will call function `update_partition_status_on_remote_storage` twice, the first one is to update its status to ingesting, the second one is to update its status to downloading. In previous implementation, the second one will be ignored and not retried, partition[2] will be ingesting and not downloading, as a result, bulk load process will be blocked.
To fix the problem above, I add `_apps_rolling_back` structure to avoid repeated rolling back request, retry partition status request.


src/meta/meta_bulk_load_service.cpp
```diff
@@ -822,22 +860,23 @@ void bulk_load_service::update_partition_status_on_remote_storage_reply(
...
-                _partitions_bulk_load_state.erase(pid);
-                _partitions_total_download_progress[pid] = 0;
-                _partitions_cleaned_up[pid] = false;
-                if (--_apps_in_progress_count[pid.get_app_id()] == 0) {
-                    _apps_in_progress_count[pid.get_app_id()] =
-                        _app_bulk_load_info[pid.get_app_id()].partition_count;
-                    ddebug_f("app({}) restart to bulk load", app_name);
-                }
+        case bulk_load_status::BLS_DOWNLOADING: {
+            _partitions_bulk_load_state.erase(pid);
```
levy5307, line 869: Could you explain why did you remove the judgement of `if (old_status != new_status) {` ?

hycdong: In previous implementation, when partition is downloading, it needn't to be updated into downloading again. However, it will raise bug because the downloading progress is not cleaned.

src/meta/meta_bulk_load_service.cpp
```diff
@@ -706,12 +706,22 @@ void bulk_load_service::handle_app_pausing(const bulk_load_response &response,
...
 void bulk_load_service::try_rollback_to_downloading(const std::string &app_name, const gpid &pid)
 {
-    zauto_read_lock l(_lock);
+    zauto_write_lock l(_lock);
+
     const auto app_status = get_app_bulk_load_status_unlocked(pid.get_app_id());
     if (app_status == bulk_load_status::BLS_DOWNLOADING ||
         app_status == bulk_load_status::BLS_DOWNLOADED ||
         app_status == bulk_load_status::BLS_INGESTING ||
         app_status == bulk_load_status::BLS_SUCCEED) {
```
neverchanje, line 715: I think App in the ingesting state could not be rolled back. How can you make rocksdb ingestion rolled back?

hycdong: The app in ingesting state should also rollback, Ingestion requires all replicas have sst files. If a replica server crash, partitions on this server will be unavailable, those partitions should rollback to downloading to ensure all replicas have downloaded files succeed.
Rocksdb ingestion don't support rollback.

neverchanje: Ok. An INGESTING app being rolled-back to DOWNLOADING is reasonsable. But why does a SUCCEED app also need rollback? If I don't misunderstand, SUCCEED means all partitions having the ingested data. The data won't lost even when all the ReplicaServers are down. This can be assured because ingestion is a 2PC.

hycdong: The app whose state is succeed are cleaning its downloaded files, and app `is_bulk_loading` is still equal to true. In current implementation, succeed will also rollback to downloading if meet recoverable errors. It may be optimized in next version.

neverchanje: You can add a comment here later. I think this case is trivial but needs notice.

src/meta/meta_bulk_load_service.cpp
```diff
@@ -769,17 +779,45 @@ void bulk_load_service::update_partition_status_on_remote_storage(const std::str
                                                                   bulk_load_status::type new_status,
                                                                   bool should_send_request)
 {
-    zauto_read_lock l(_lock);
+    zauto_write_lock l(_lock);
     partition_bulk_load_info pinfo = _partition_bulk_load_info[pid];
-
-    if (pinfo.status == new_status) {
+    if (pinfo.status == new_status && new_status != bulk_load_status::BLS_DOWNLOADING) {
```
neverchanje, line 784: This if statement implies DOWNLOADING => DOWNLOADING requires a meta sync. I don't understand the reason of this code. I suggest a comment here.

src/meta/meta_bulk_load_service.cpp
hycdong: You can see bulk load process graph in `meta_bulk_load_service.h`, it shows possible status convert.

src/meta/meta_bulk_load_service.cpp
hycdong: When trying to call `update_partition_status_on_remote_storage` to update partition status to downloading, it will only be happened when rolling back to downloading. Consider the case, app is downloading, partition[0-2] is downloaded, partition[3] is downloading, some recoverable error happens, whole status will rollback to downloading, app status downloading -> downloading, partition status should also turn to downloading,  and clear local status in function `update_partition_status_on_remote_storage_reply`

PR #579, opened 2020-08-03
fix(bulk-load): fix bug while check_all_partitions

hycdong:
Pull request #571 add a check for load balance, apps who are executing bulk load should not be balanced. However, it raises a bug that the apps who are executing bulk load won't add secondary. This pull request fix this bug.

### DIFF
--- a/src/meta/server_state.cpp
+++ b/src/meta/server_state.cpp
@@ -2344,11 +2344,6 @@ bool server_state::check_all_partitions()
                    ::dsn::enum_to_string(app->status));
             continue;
         }
-        if (app->is_bulk_loading) {
-            ddebug_f(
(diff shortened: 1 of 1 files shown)


src/meta/server_state.cpp
```diff
@@ -2344,11 +2344,6 @@ bool server_state::check_all_partitions()
                    ::dsn::enum_to_string(app->status));
             continue;
         }
-        if (app->is_bulk_loading) {
```
levy5307, line 2347: Does it will produce  error when rebalance the bulk loading app?

hycdong: balance operation is in function `greedy_balancer`, you can see it in pull request #571.

PR #580, opened 2020-08-04
fix: remote-command max_copy_rate_megabytes failed when set "DEFAULT"

Shuo-Jia:
When update the `max_copy_rate_megabytes` via remote-command, the DEFAULT value will be changed. This PR add `current_max_copy_rate_megabytes` store the latest value and let the `FLAGS_max_copy_rate_megabytes` store the default value.

### DIFF
--- a/src/nfs/nfs_client_impl.cpp
+++ b/src/nfs/nfs_client_impl.cpp
@@ -39,6 +39,7 @@
 
 namespace dsn {
 namespace service {
+static uint32_t current_max_copy_rate_megabytes = 0;
 
 DSN_DEFINE_uint32("nfs",
                   nfs_copy_block_bytes,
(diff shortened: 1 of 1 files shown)


src/nfs/nfs_client_impl.cpp
```diff
@@ -567,12 +569,13 @@ void nfs_client_impl::register_cli_commands()
                 std::string result("OK");
 
                 if (args.empty()) {
-                    return std::to_string(FLAGS_max_copy_rate_megabytes);
```
levy5307, line 570: Does it make sense that change `FLAGS_max_copy_rate_megabytes` to `current_max_copy_rate_megabytes` in this pull request?

Shuo-Jia: `args.empty()` means query current value but default value

PR #581, opened 2020-08-04
fix(bulk-load): fix bug about concurrent downloading count

hycdong:
This pull request fix the bug about concurrent bulk load downloading replica count.
The bug will be raised when bulk load failed or paused. In previous implementation, when bulk load failed or paused, if original status is downloading, we will decrease this count. However, the partition may be not doing exact downloading. For example, the max count is 5, this stub already have 5 partitions executing downloading, besides there are 8 partitions waiting for downloading, one of them meet unrecoverable error so the whole bulk load process failed, all 13 partition will try to decrease the count, which leads to wrong count -8.
To fix this bug, this pull request adds a `is_downloading` variety, only the partition who is executing downloading will decrease this count. Besides, this pull request adds unit test for this condition.

### DIFF
--- a/src/replica/bulk_load/replica_bulk_loader.cpp
+++ b/src/replica/bulk_load/replica_bulk_loader.cpp
@@ -349,6 +349,7 @@ error_code replica_bulk_loader::start_download(const std::string &app_name,
     _stub->_counter_bulk_load_downloading_count->increment();
 
     // start download
+    _is_downloading = true;
     ddebug_replica("start to download sst files");
     error_code err = download_sst_files(app_name, cluster_name, provider_name);
     if (err != ERR_OK) {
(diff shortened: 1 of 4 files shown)


acelyc111:
Reviewer can't distinguish which code resolve which problem, I think separet this PR to 2 PRs would be better.

hycdong:
> Reviewer can't distinguish which code resolve which problem, I think separet this PR to 2 PRs would be better.

ok, I will remove the code about `clear_bulk_load_states_if_needed` in this pull request

src/replica/bulk_load/replica_bulk_loader.cpp
```diff
@@ -487,7 +488,11 @@ void replica_bulk_loader::update_bulk_load_download_progress(uint64_t file_size,
 // ThreadPool: THREAD_POOL_REPLICATION, THREAD_POOL_REPLICATION_LONG
 void replica_bulk_loader::try_decrease_bulk_load_download_count()
 {
+    if (!_is_downloading.load()) {
```
acelyc111, line 491: Could this function be called concurrently in multi-thread? If it could, there is still a gap that `--_stub->_bulk_load_downloading_count;` will be executed repeatly.

hycdong: This function is run in replication thread pool, will not be called concurrently in different thread.

PR #584, opened 2020-08-05
refactor: make read batch size of fds configurable

levy5307:
The read batch size of fds is limited to 1MB now. But it may limit the read speed of fds. So in this pull request, I add a configration for it.

###New-Config
```diff
[replication]
+ fds_read_batch_size = 100
```

### DIFF
--- a/src/block_service/fds/fds_service.cpp
+++ b/src/block_service/fds/fds_service.cpp
@@ -569,7 +569,8 @@ error_code fds_file_object::get_content_in_batches(uint64_t start,
                                                    /*out*/ uint64_t &transfered_bytes)
 {
     // the max batch size is 1MB
-    const uint64_t BATCH_SIZE = std::min(FLAGS_fds_read_limit_rate, FLAGS_fds_read_batch_size) << 20;
+    const uint64_t BATCH_SIZE = std::min(FLAGS_fds_read_limit_rate, FLAGS_fds_read_batch_size)
+                                << 20;
     error_code err = ERR_OK;
     transfered_bytes = 0;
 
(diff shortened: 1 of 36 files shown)


src/block_service/fds/fds_service.cpp
```diff
@@ -37,9 +37,11 @@ DSN_DEFINE_uint64("pegasus.server",
                   64 << 20,
                   "rocksdb options.write_buffer_size");
 
-DSN_DEFINE_uint32("replication", fds_write_limit_rate, 100, "rate limit of fds(MB/s)");
+DSN_DEFINE_uint32("replication", fds_write_limit_rate, 100, "write rate limit of fds(MB/s)");
 
-DSN_DEFINE_uint32("replication", fds_read_limit_rate, 100, "rate limit of fds(MB/s)");
+DSN_DEFINE_uint32("replication", fds_read_limit_rate, 100, "read rate limit of fds(MB/s)");
+
+DSN_DEFINE_uint32("replication", fds_read_batch_size, 5, "read batch size of fds(MB)");
```
Shuo-Jia, line 44: Have you test the default 5 is OK？

src/block_service/fds/fds_service.cpp
```diff
@@ -37,9 +37,11 @@ DSN_DEFINE_uint64("pegasus.server",
                   64 << 20,
                   "rocksdb options.write_buffer_size");
 
-DSN_DEFINE_uint32("replication", fds_write_limit_rate, 100, "rate limit of fds(MB/s)");
+DSN_DEFINE_uint32("replication", fds_write_limit_rate, 100, "write rate limit of fds(MB/s)");
```
hycdong, line 40: I recommend those three configuration can be updated dynamically

levy5307: Do you mean add a remote command? I think we can add it in next pull request.

hycdong: Yes~ I will expect your further pull request.

src/block_service/fds/fds_service.cpp
levy5307: I set it to 100MB to ensure the speed will be very quick.  Somebody can change it to a small number when he wants to smooth the network speed.

PR #585, opened 2020-08-06
feat(security): init kerberos

levy5307:
If the config of `enable_auth` is true, we should init kerberos.
Here is the detail procedure:
1. set envs which are used in kerberos, like `KRB5CCNAME`, `KRB5_CONFIG` and so on.
2. create a kerberos context.
3. convert a string principal name to a krb5_principal structure. principal and username that logged in as, this determines "who I am"
4. get a handle for a key table. key table is the file to store principals and it's correponding key.
5. acquire credential cache handle and initialize credential cache. The credential cache is used for storing credentials.
6. allocate a new initial credential options structure, the options are used for getting credentials.
7. get and schedule to renew credentials from KDC and store it into _ccache


include/dsn/utility/time_utils.h
```diff
@@ -68,6 +68,13 @@ inline void time_ms_to_date_time(uint64_t ts_ms, char *str, int len)
     strftime(str, len, "%Y-%m-%d %H:%M:%S", get_localtime(ts_ms, &tmp));
 }
```
acelyc111, line 70: Add comment like functions above

src/runtime/CMakeLists.txt
```diff
@@ -30,5 +30,5 @@ add_library(dsn_runtime STATIC
         tracer.cpp
         zlocks.cpp
         )
-target_link_libraries(dsn_runtime dsn_utils)
+target_link_libraries(dsn_runtime dsn_utils sasl2 gssapi_krb5 krb5)
```
acelyc111, line 33: So don't forget to update docs

src/runtime/security/kerberos_utils.cpp
```diff
@@ -0,0 +1,313 @@
+// Copyright (c) 2017, Xiaomi, Inc.  All rights reserved.
```
acelyc111: update license

src/runtime/security/kerberos_utils.h
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) 2017, Xiaomi, Inc.  All rights reserved.
+// This source code is licensed under the Apache License Version 2.0, which
+// can be found in the LICENSE file in the root directory of this source tree.
+
+#pragma once
+
+#include <dsn/utility/errors.h>
+#include <dsn/utility/synchronize.h>
```
acelyc111, line 21: This header is not needed here, if cpp needed it, move it there.

src/runtime/security/kerberos_utils.cpp
```diff
@@ -0,0 +1,313 @@
...
+    /// setup kerberos envs(for more details:
+    /// https://web.mit.edu/kerberos/krb5-1.12/doc/admin/env_variables.html)
+    setenv("KRB5CCNAME", is_server ? "MEMORY:pegasus-server" : "MEMORY:pegasus-client", 1);
+    setenv("KRB5_CONFIG", FLAGS_krb5_config, 1);
+    setenv("KRB5_KTNAME", FLAGS_krb5_keytab, 1);
+    setenv("KRB5RCACHETYPE", "none", 1);
+
+    g_kinit_ctx.reset(new internal::kinit_context);
+    err = g_kinit_ctx->kinit(FLAGS_krb5_keytab, FLAGS_krb5_principal);
+    ddebug_f("something is wrong with kinit, err = {}", err.description());
```
acelyc111, line 321: This log is only needed when init failed.

src/runtime/security/kerberos_utils.cpp
```diff
@@ -0,0 +1,313 @@
...
+    ddebug_f("logged in from keytab as {}, local username {}", _principal_name, _username);
+    return error_s::make(ERR_OK);
+}
+
+#undef KRB5_RETURN_NOT_OK // only used in this internal namespace
+#undef WRAP_KRB5_ERR      // only used in this internal namespace
+} // namespace internal
+
+error_s check_configuration()
+{
```
acelyc111, line 283: Add `dassert(FLAGS_enable_auth)` here to ensure and tell reviewers this functions only executed when FLAGS_enable_auth is true.

src/runtime/security/kerberos_utils.cpp
```diff
@@ -0,0 +1,313 @@
...
+    uint64_t _cred_expire_timestamp;
+
+    std::shared_ptr<boost::asio::deadline_timer> _timer;
+
+private:
+    // get _principal_name and _user_name from _principal
+    error_s get_formatted_identities();
+    // get or renew credentials from KDC and store it to _ccache
+    error_s get_credentials();
+    void schedule_renew_credentials();
```
acelyc111, line 159:
```suggestion
private:
    // get _principal_name and _user_name from _principal
    error_s get_formatted_identities();
    // get or renew credentials from KDC and store it to _ccache
    error_s get_credentials();
    void schedule_renew_credentials();

private:
    // krb5 structure
    krb5_principal _principal;
    // keytab file with absolute path
    krb5_keytab _keytab;
    krb5_ccache _ccache;
    krb5_get_init_creds_opt *_opt;

    // principal and username that logged in as, this determines "who I am"
    std::string _principal_name;
    std::string _username;

    uint64_t _cred_expire_timestamp;

    std::shared_ptr<boost::asio::deadline_timer> _timer;
```

src/runtime/security/kerberos_utils.cpp
```diff
@@ -0,0 +1,313 @@
...
+    //  1. currently the rdsn framework may not started yet.
+    //  2. the rdsn framework is used for codes of a service_app,
+    //     not for codes under service_app
+    if (nullptr == _timer) {
+        _timer.reset(new boost::asio::deadline_timer(tools::shared_io_service::instance().ios));
+    }
+    _timer->expires_from_now(boost::posix_time::seconds(renew_gap));
+    _timer->async_wait([this](const boost::system::error_code &err) {
+        if (!err.failed()) {
+            get_credentials();
```
acelyc111, line 182: Will you try to find credential in krb5_ccache in other PRs?

acelyc111:
Take a look at https://github.com/apache/kudu/blob/master/src/kudu/security/init.cc may help.

src/runtime/security/kerberos_utils.cpp
levy5307: No, the krb5_ccache is used for kerberos internal, we don't need it.

vagetablechicken: It's not internal

src/runtime/CMakeLists.txt
levy5307: Which docs do you mean?

src/runtime/security/kerberos_utils.cpp
levy5307: I have read some docs of kerberos. It says 
```The credential cache file holds Kerberos protocol credentials (for example, tickets, session keys, and other identifying information) in semipermanent storage. The Kerberos protocol reads credentials from the cache as they are required and stores new credentials in the cache as they are obtained. This relieves the application of the responsibility for managing the credentials itself.``` in https://www.ibm.com/support/knowledgecenter/en/SSGSMK_7.1.0/management_sym/sym_kerberos_configuring_credential_cache_file.html

And I have read the corresponding code in kudu, there are no code to try to find credential in krb5_ccache. It just put credentials into krb5_ccache.

Comment (vagetablechicken) [reply src/runtime/security/kerberos_utils.cpp]: so you mean krb5_cc_ funcs are useless? You should read more. It's not for internal. I didn't say we need to use it.

Comment (levy5307) [reply src/runtime/security/kerberos_utils.cpp]: The Kerberos protocol reads credentials from the cache as they are required and stores new credentials in the cache as they are obtained. I didn't say they are useless

Comment (levy5307) [reply src/runtime/security/kerberos_utils.cpp]: So what the credential cache is used for in your opinion? @vagetablechicken

Comment (vagetablechicken) [reply src/runtime/security/kerberos_utils.cpp]: What I said is the basic conceptive problem. krb5_ccache is not for krb internal. The owner can use it to avoid unnecessary connections.
There's no need to optimize prematurely.

File: src/runtime/service_api_c.cpp [hunk 9]
```diff
@@ -523,7 +535,7 @@ bool run(const char *config_file,
     // start the tool
     dsn_all.tool->run();
 
-    if (sleep_after_init) {
+    if (is_server) {
```
hycdong, line 538: Why server should sleep? and when `is_server = true` and when `is_server=false`? besides, do we need `sleep_after_init` any more?

src/runtime/security/kinit_context.cpp
```diff
@@ -0,0 +1,306 @@
...
+DSN_DEFINE_string("security", krb5_principal, "", "kerberos principal");
+
+// Attention: we can't do these check work by `DSN_DEFINE_validator`, because somebody may don't
+// want to use security, so these configuration may not setted. In this situation, these checks
+// will not pass.
+error_s check_configuration()
+{
+    dassert(FLAGS_enable_auth,
+            "There is no need to check configuration if FLAGS_enable_auth is not true");
+
```
hycdong, line 55: The comment shows that the code try to get keytab files from KDC, why should we check file exists on disk? Do we download the files from KDC or just cache those information in memory?

src/runtime/security/kinit_context.cpp
```diff
@@ -0,0 +1,306 @@
...
+    _cred_expire_timestamp = creds.times.endtime;
+    ddebug_f("get credentials of {} from KDC ok, expires at {}",
+             FLAGS_krb5_principal,
+             utils::time_s_to_date_time(_cred_expire_timestamp));
+    return err;
+}
+
+void kinit_context_impl::schedule_renew_credentials()
+{
+    // reserve 300 seconds for renew
```
hycdong, line 246: Why use 300 seconds as the interval for renew?

src/runtime/security/kinit_context.cpp
```diff
@@ -0,0 +1,306 @@
...
+    _user_name.assign((const char *)buf);
+    return error_s::ok();
+}
+
+error_s kinit_context_impl::get_credentials()
+{
+    krb5_creds creds;
+    error_s err = error_s::ok();
+
+    // get initial credentials using a key table
```
hycdong, line 210: Maybe `using a keytab`?

src/runtime/security/kinit_context.cpp
```diff
@@ -0,0 +1,306 @@
...
+    // Notice: the contents of a krb5_creds structure need to be freed by ourselves
```
hycdong, line 211: Dose this comment mean that we should call function `krb5_free_cred_contents` to free krb5_creds structure?

src/runtime/security/kinit_context.cpp
```diff
@@ -0,0 +1,306 @@
...
+            schedule_renew_credentials();
+        } else if (err == boost::system::errc::operation_canceled) {
+            dwarn("the renew credentials timer is cancelled");
+        } else {
+            dassert_f(false, "unhandled error({})", err.message());
+        }
+    });
+}
+
+// switch the code of krb5_xxx function to error_s
```
hycdong, line 272: What is the `krb5_xxx function`? maybe " switch krb5_error_code and message to error_s"?

src/runtime/security/kinit_context.cpp
```diff
@@ -0,0 +1,306 @@
...
+    void schedule_renew_credentials();
+
+    error_s wrap_krb5_err(krb5_error_code krb5_err, const std::string &msg);
+    error_s krb5_call_to_errors(krb5_error_code krb5_code, const std::string &prefix_msg);
+
+private:
+    krb5_context _krb5_context;
+    // krb5 principal
+    krb5_principal _principal;
+    // keytab file with absolute path
```
hycdong, line 99: This comment makes confused.

src/runtime/CMakeLists.txt
hycdong, line 33: Do `sasl2 gssapi_krb5 krb5` all download from thirdparty in this pull request?

neverchanje: http://pegasus.apache.org/overview/compilation & https://github.com/XiaoMi/pegasus/blob/master/docs/installation.md

src/runtime/security/kinit_context.h
```diff
@@ -0,0 +1,42 @@
...
+    // implementation of 'kinit -k -t <keytab_file> <principal>'
+    error_s kinit();
+
+private:
+    friend class dsn::utils::singleton<kinit_context>;
+    kinit_context();
+
+    // compilation firewall: decouple interface from implementation
+    std::unique_ptr<kinit_context_impl> impl;
+};
```
neverchanje, line 40: How about simply using one func?

```suggestion
error_s run_kinit();
```

In somewhere the source file:

```
error_s run_kinit() {
  static kinit_context_impl ctx;
  ctx.kinit();
}
```

src/runtime/security/kinit_context.cpp
levy5307: `keytab` is the abbreviations of `key table`, so I think it's ok to use key table

src/runtime/security/kinit_context.cpp
levy5307: yes

src/runtime/CMakeLists.txt
levy5307: yes

src/runtime/security/kinit_context.cpp
levy5307: No, keytab file is provided in our service's local machine. The code get from KDC is TGT

src/runtime/service_api_c.cpp
levy5307: If it is a server, the startup thread should sleep. In previous implementation, it use `sleep_after_init` instead of `is_server`. So I change it to `is_server` in this pull request

src/runtime/security/kinit_context.cpp
levy5307: Let me have a look. I didn't notice it before :(

src/runtime/security/kinit_context.cpp
levy5307: What makes you confused? 
Sorry I don't know how to explain this to you. Because it's the basic conceptive in kerberos.

src/runtime/security/kinit_context.cpp
```diff
@@ -0,0 +1,322 @@
...
+    krb5_keytab _keytab;
+    // credential cache
+    krb5_ccache _ccache;
```
acelyc111, line 104: Now ccache is not used, could you plz add a TODO?

acelyc111 (approved):

acelyc111 (approved):

src/runtime/security/init.h
```diff
@@ -0,0 +1,27 @@
...
+// under the License.
+
+#pragma once
+
+#include <dsn/utility/errors.h>
+
+namespace dsn {
+namespace security {
+// init security(kerberos and sasl)
+bool init(bool is_server);
```
neverchanje, line 25:
```suggestion
extern bool init_server();

extern bool init_client();
```

levy5307: I think there is no need to split the function into two functions. Because they are almost same

hycdong (dismissed):

PR #587, opened 2020-08-10
fix(bulk-load): fix bug during ingestion

hycdong:
This pull request fixs the bug of bulk load ingestion.
In previous implementation, when replica is executing ingestion request, it will reject normal write request by returning `ERR_OPERATION_DISABLED`. When replica received repeated ingestion request, it will also return `ERR_OPERATION_DISABLED`, meta will rollback to downloading if receiving a response with this error code. However, meta needn't to rollback to downloading, it should just wait for this ingestion. This pull request separates those different case. When replica is executing ingestion request, if it receives normal write request, it will still return `ERR_OPERATION_DISABLED` to user client, if it receives repeated ingestion request, it will return `ERR_NO_NEED_OPERATE` to meta server, meta server will ignore this repeated request. Besides, this pull request also adds bulk load status check while receiving ingestion request, if it is invalid, it will return `ERR_INVALID_STATE` to meta server.

### DIFF
--- a/src/meta/meta_bulk_load_service.cpp
+++ b/src/meta/meta_bulk_load_service.cpp
@@ -1019,6 +1019,14 @@ void bulk_load_service::on_partition_ingestion_reply(error_code err,
                                                      const std::string &app_name,
                                                      const gpid &pid)
 {
+    if (err == ERR_NO_NEED_OPERATE) {
+        dwarn_f(
+            "app({}) partition({}) has already executing ingestion, ignore this repeated request",
+            app_name,
+            pid);
+        return;
+    }
+
     // if meet 2pc error, ingesting will rollback to downloading, no need to retry here
     if (err != ERR_OK) {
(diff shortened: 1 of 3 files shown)


src/meta/meta_bulk_load_service.cpp
```diff
@@ -1019,6 +1019,14 @@ void bulk_load_service::on_partition_ingestion_reply(error_code err,
                                                      const std::string &app_name,
                                                      const gpid &pid)
 {
+    if (err == ERR_NO_NEED_OPERATE) {
```
Shuo-Jia, line 1022: `ERR_NO_NEED_OPERATE` seems to be not clear, how about `ERR_OPERATION_REPEATED`

hycdong: `ERR_NO_NEED_OPERATE` is a existed error code. If I use `ERR_OPERATION_REPEATED`, I should define a new one.

PR #588, opened 2020-08-11
feat(security): handle SASL_LIST_MECHANISMS by server_negotiation

levy5307:
server_negotiation receives SASL_LIST_MECHANISMS request, and reply the mechanism that it supports.

```
 negotiation process:

                                                  client                             server
                                                     |---   SASL_LIST_MECHANISMS    -->|
                                                     |<-- SASL_LIST_MECHANISMS_RESP ---|
                                                     |---  SASL_SELECT_MECHANISMS  --->|
                                                     |<-- SASL_SELECT_MECHANISMS_OK ---|
                                                     |                                 |
                                                     |------    SASL_INITIATE   ------>|
                                                     |                                 |
                                                     |<----     SASL_CHALLENGE    -----|
                                                     |-----      SASL_RESPONSE    ---->|
                                                     |                                 |
                                                     |              .....              |
                                                     |                                 |
                                                     |<----     SASL_CHALLENGE    -----|
                                                     |-----      SASL_RESPONSE    ---->|
                                                     |                                 |        (authentication will succeed
                                                     |                                 |        if all chanllenges passed)
                                                     | <---        SASL_SUCC      -----|
    (client won't response                           |                                 |
     if servers says ok)                             |                                 |
                                                     | ----         RPC_CALL      ---> |
                                                     | <---         RPC_RESP       ----|
```


include/dsn/utility/strings.h
```diff
@@ -49,5 +50,23 @@ char *trim_string(char *s);
 
 // calculate the md5 checksum of buffer
 std::string string_md5(const char *buffer, unsigned int length);
+
+// merge strings with comma delimiter
```
acelyc111, line 54:
```suggestion
// merge strings with specified delimiter
```

src/runtime/rpc/network.cpp
```diff
@@ -442,6 +449,21 @@ void rpc_session::start_negotiation()
...
+    }
+}
+
+void rpc_session::complete_negotiation(bool succ)
+{
+    if (succ) {
+        if (is_client()) {
+            set_connected();
+            on_send_completed();
+        }
```
acelyc111, line 464: wrap as a function `on_success` would be better, it has the same level to `on_failure `

src/runtime/rpc/network.cpp
```diff
@@ -442,6 +449,21 @@ void rpc_session::start_negotiation()
         }
 
         auth_negotiation();
+    } else {
+        // set negotiation success if auth is disabled
+        complete_negotiation(true);
```
acelyc111, line 454: In which case to pass `false` to `complete_negotiation `?

src/runtime/security/negotiation_service.cpp
```diff
@@ -0,0 +1,51 @@
...
+    dassert(!rpc.dsn_request()->io_session->is_client(),
+            "only server session receive negotiation request");
+
+    // return SASL_AUTH_DISABLE if auth is not enable
+    if (!security::FLAGS_enable_auth) {
+        rpc.response().status = negotiation_status::type::SASL_AUTH_DISABLE;
+        return;
+    }
+
+    server_negotiation *s_negotiation =
```
acelyc111, line 45: `s_` prefix is often use for static variables, it may cause some misunderstanding. Use `srv` or some other

src/runtime/security/negotiation_service.cpp
```diff
@@ -0,0 +1,51 @@
...
+        RPC_NEGOTIATION, "Negotiation", &negotiation_service::on_negotiation_request);
+}
+
+void negotiation_service::on_negotiation_request(negotiation_rpc rpc)
+{
+    dassert(!rpc.dsn_request()->io_session->is_client(),
+            "only server session receive negotiation request");
+
+    // return SASL_AUTH_DISABLE if auth is not enable
+    if (!security::FLAGS_enable_auth) {
```
acelyc111, line 40: negotiation_service is opened only if FLAGS_enable_auth is true, so it will never happen?

src/runtime/rpc/network.cpp
levy5307: When negotiation failed, false will passed to `complete_negotiation`

src/runtime/security/negotiation_service.cpp
levy5307: If `FLAGS_enable_auth` is false,  it means server is not enable auth. So there is no need to do negotiation and return to client with `SASL_AUTH_DISABLE`. When client receives `SASL_AUTH_DISABLE`, it will set negotiation success

include/dsn/utility/strings.h
```diff
@@ -49,5 +50,23 @@ char *trim_string(char *s);
 
 // calculate the md5 checksum of buffer
 std::string string_md5(const char *buffer, unsigned int length);
+
+// merge strings with specified delimiter
+template <typename ForwardIterator>
+std::string merge(ForwardIterator begin, ForwardIterator end, const std::string &delimiter)
```
neverchanje, line 56: I think [boost::join](https://www.boost.org/doc/libs/1_65_1/boost/algorithm/string/join.hpp) can replace this function.

src/runtime/security/negotiation.h
```diff
@@ -24,7 +24,6 @@ namespace dsn {
 class rpc_session;
 
 namespace security {
-
```
neverchanje, line 27: why remove this empty line?

src/runtime/security/client_negotiation.cpp
```diff
@@ -43,9 +49,14 @@ void client_negotiation::list_mechanisms()
...
-    dsn::marshall(req, request);
-    _session->send_message(req);
+    message_ptr msg = message_ex::create_request(RPC_NEGOTIATION);
+    dsn::marshall(msg.get(), request);
+
+    rpc_response_task_ptr t = rpc::create_rpc_response_task(
+        msg, nullptr, [this](error_code err, negotiation_response response) {
+            handle_response(err, std::move(response));
+        });
+    dsn_rpc_call(_session->remote_address(), t);
```
neverchanje, line 59:
```
void client_negotiation::send(std::unique_ptr<negotiation_request> req) {
  negotiation_rpc rpc(std::move(req), RPC_NEGOTIATION);
  rpc.call(_session->remote_address(), nullptr, [rpc](){handle_response(err, rpc.response())});
}
```

```
void client_negotiation::list_mechanisms()
{
    auto request = dsn::make_unique<negotiation_request>();
    _status = request->status = negotiation_status::type::SASL_LIST_MECHANISMS;
    send(std::move(request));
}
```

src/runtime/rpc/network.cpp
acelyc111: Only `true` is passed in this PR, will `false` be passed by another PR?

src/runtime/security/negotiation_service.cpp
acelyc111: In src/runtime/service_api_c.cpp:L556, `negotiation_service::open_service ` is called only when `FLAGS_enable_auth ` is true, so if program execute here, `FLAGS_enable_auth ` must be true, not needed to judge it.

src/runtime/rpc/network.cpp
levy5307: yes, you are right

src/runtime/security/negotiation_service.cpp
levy5307: I changed the code in `service_api_c.cpp` like this:
```
service_app::service_app(const dsn::service_app_info *info) : _info(info), _started(false)
{
    security::negotiation_service::instance().open_service();
}
```
Remove the judgement of `if (FLAGS_enable_auth)`

src/runtime/rpc/network.cpp
```diff
@@ -442,6 +457,18 @@ void rpc_session::start_negotiation()
...
```
acelyc111, line 462: For boolean parameter, better to add some comments on it, alternatively, use enum instead of bool, https://abseil.io/tips/94

levy5307: I have add some comments in front of the interface `complete_negotiation` in network.h

src/runtime/rpc/network.cpp
```diff
@@ -442,6 +457,18 @@ void rpc_session::start_negotiation()
...
+    }
+}
+
+void rpc_session::complete_negotiation(bool succ)
```
neverchanje, line 466: I think `on_failure` or `on_success` is easy enough to use, why add a new function complete_negotiation?

src/runtime/security/server_negotiation.h
```diff
@@ -21,13 +21,19 @@
 
 namespace dsn {
 namespace security {
+extern const std::set<std::string> supported_mechanisms;
 
 class server_negotiation : public negotiation
```
neverchanje, line 26: Why is `rpc_session` holding a "server_negotiation" with no data? What gonna be inside `server_negotiation`?

levy5307: `server_negotiation` have some members. I will add them in later pull requests.

neverchanje: Can you list them? Just check if the overall design is good.

levy5307: `std::string _selected_mechanism` is the mechanism selected by negotiation
`std::unique_ptr<sasl_conn_t, sasl_deleter> _sasl_conn` is the sasl connection context

neverchanje: Ok. You gave a quite bad name. So the class `negotiation` is actually "sasl_context". I think the latter is more descriptive. I don't have to check the code to know the overall design.

“server_negatiation” is "server_sasl_context".

levy5307: No, it's not only `sasl_context`. The class `negotiation` should also do negotiation work to select mechansim.

src/runtime/security/negotiation_service.h
```diff
@@ -0,0 +1,40 @@
...
+class negotiation_service : public serverlet<negotiation_service>,
+                            public utils::singleton<negotiation_service>
+{
+public:
+    void open_service();
+
+private:
+    negotiation_service();
+    void on_negotiation_request(negotiation_rpc rpc);
+    friend class utils::singleton<negotiation_service>;
```
neverchanje, line 36:
```suggestion
```
Is this line redundant?

src/runtime/security/negotiation_service.cpp
```diff
@@ -0,0 +1,51 @@
...
+            "only server session receive negotiation request");
+
+    // reply SASL_AUTH_DISABLE if auth is not enable
+    if (!security::FLAGS_enable_auth) {
+        rpc.response().status = negotiation_status::type::SASL_AUTH_DISABLE;
+        return;
+    }
+
+    server_negotiation *srv_negotiation =
+        dynamic_cast<server_negotiation *>(rpc.dsn_request()->io_session->get_negotiation());
```
neverchanje, line 46:
```suggestion
    server_negotiation *srv_negotiation =
        dynamic_cast<server_negotiation *>(rpc.dsn_request()->io_session->get_negotiation());
```

[dynamic_cast](https://en.cppreference.com/w/cpp/language/dynamic_cast) introduces RTTI,  which unaffordable to the critical path(rpc negotiation). Use static_cast instead.

```
auto srv_negotiation = static_cast<server_negotiation *>(rpc.dsn_request()->io_session->get_negotiation());
```

src/runtime/security/server_negotiation.cpp
```diff
@@ -33,5 +36,38 @@ void server_negotiation::start()
...
+        on_list_mechanisms(rpc);
+        return;
+    }
+}
+
+void server_negotiation::on_list_mechanisms(negotiation_rpc rpc)
+{
+    if (rpc.request().status == negotiation_status::type::SASL_LIST_MECHANISMS) {
+        std::string mech_list = boost::join(supported_mechanisms, ",");
+        ddebug_f("{}: reply server mechs({})", _name, mech_list);
```
neverchanje, line 51:
```suggestion
        std::string mech_list = boost::join(supported_mechanisms, ",");
        ddebug_f("{}: reply server mechs({})", _name, mech_list);
```

This log will be as many as client connections. I don't think it's necessary.

src/runtime/security/negotiation_service.h
levy5307: no, because the cstor of `negotiation_service` is private. We should make `class utils::singleton` to be a friend class to access cstor of `negotiation_service`

src/runtime/security/negotiation_service.cpp
levy5307: But the RTTI is used for judging whether the cast is safe or not in `dynamic_cast`.  In 《effective C++》, `dynamic_cast` is recommended

neverchanje: It depends on whether you need runtime type check in production, and whether you can afford the penalty of RTTI. If yourself can ensure the safety, use `static_cast` is enough and efficient.

https://github.com/apache/kudu/blob/62123b097abf14d5a71fb3d26c0909ce2bfde9f4/src/kudu/gutil/casts.h#L69

kudu has a utility called `down_cast` which is `dynamic_cast` when it's in debug mode, and turns to `static_cast` when it's compiled in release mode.

levy5307: I think the implementation of kudu about this is a best practise.

acelyc111 (approved):

neverchanje (approved):

PR #589, opened 2020-08-12
feat(security): make network support security negotiation

levy5307:
1. Add judgement for whether it is auth success when receves message.  Reject the msg if auth is not success and the msg is not negotiation.

2. Add judgement for whether it is negotiation message when send message.Only negotiation message is sent if the connecting state is not connected.


src/runtime/rpc/network.cpp
```diff
@@ -391,13 +415,36 @@ bool rpc_session::on_disconnected(bool is_write)
...
+{
+    if (security::FLAGS_enable_auth && !_negotiation->negotiation_succeed()) {
+        dwarn_f("reject message({}) from {}, session {} client",
+                msg->rpc_code().to_string(),
+                _remote_addr.to_string(),
+                is_client() ? "is" : "isn't");
+        // reply response with ERR_UNAUTHENTICATED if msg is request
+        if (msg->header->context.u.is_request) {
+            _net.engine()->reply(msg->create_response(), ERR_UNAUTHENTICATED);
+        }
```
acelyc111, line 428: This is only a 'check' type function, I think reply message in `on_recv_message ` would be better.

include/dsn/tool-api/network.h
```diff
@@ -271,6 +271,10 @@ class rpc_session : public ref_counter
     ::dsn::utils::ex_lock_nr _lock; // [
     volatile session_state _connect_state;
 
+    // when a session isn't connected, all messages are queued in _pending_connected.
+    // after connected, all of them are moved to "_messages"
+    std::vector<message_ex *> _pending_connected;
```
acelyc111, line 276: Why introduce it? Suppose the negotiation procedure last a very long time, `_pending_connected` may consume a large amount of memory.

src/runtime/rpc/network.cpp
```diff
@@ -272,18 +287,27 @@ void rpc_session::send_message(message_ex *msg)
...
+
+            if (!_is_sending_next) {
+                _is_sending_next = true;
+                sig = _message_sent + 1;
+                unlink_message_for_send();
+            } else { // is under send msg, just return
+                return;
+            }
         } else {
+            dassert(_connect_state != SS_CONNECTED, "invalid session state");
```
acelyc111, line 306: Use `dcheck_ne` (check not equal) would be better, then when check failed, you will know the exact value of  _connect_state which may help you to debug.
However, now `dcheck_ne` is lacked, you can add it by another PR.

include/dsn/tool-api/network.h
levy5307: So what's your opinion? throw all of these requests which are generated during negotiation?

src/runtime/rpc/network.cpp
levy5307: I will remove `dassert(_connect_state != SS_CONNECTED, "invalid session state");`, because `_connect_state != SS_CONNECTED` is always true in the else code branch

include/dsn/tool-api/network.h
acelyc111: I think drop them is OK, message in the queue may also timeouted when negotiation finished.
However, I want to hear somebody other's voice.

levy5307: @neverchanje @hycdong

neverchanje: I agree with @acelyc111. It's an early-optimization.

Actually the client could also have implemented a pending queue. The requests may be queued on the client-side until the session authenticated. Thus the server-queue during authentication could always be empty and useless.

levy5307: I agree with you. But if we remove the `pending_connected`, so the cpp sdk will not support the pending queue

neverchanje: There's always one pending queue on the client-side once it supports asynchronous operations.

PR #591, opened 2020-08-13
build: use nodejs/http_parser as a external dep

neverchanje:


thirdparty/CMakeLists.txt
```diff
@@ -295,3 +295,13 @@ ExternalProject_Add(rocksdb
...
         -DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
         )
+
+ExternalProject_Add(http-parser
+        URL ${OSS_URL_PREFIX}/http-parser-2.9.4.zip
+        https://github.com/nodejs/http-parser/archive/v2.9.4.zip
+        URL_MD5 f6900b9209d3d6b80c70e050ac33b834
+        CONFIGURE_COMMAND mkdir -p ${TP_OUTPUT}/include/nodejs
+        BUILD_COMMAND ""
+        INSTALL_COMMAND cp -R http_parser.h ${TP_OUTPUT}/include/nodejs/http_parser.h
```
acelyc111, line 305: This thirdparty is only copied to output directory but compiled in rdsn project?

neverchanje: Yes. We don't compile http_parser as "libhttp_parser.a", otherwise we must link this library to every place.

acelyc111: Only http related modules need link this library, I think keep logic clear and unified would be better.

neverchanje: > "Only http related modules need link this library"

Yes, which means meta-server, replica-server, and all related modules need to link this library.
If we don't make it an external dependency, I will paste the http-parser v2.9.4 sources to replace the current files.

PR #595, opened 2020-08-19
feat(security): receive mechanisms which are supported by server_negotiation

levy5307:
In pr #588 ,  server_negotiation receives SASL_LIST_MECHANISMS request, and reply the mechanism that it supports. This pr is implemented to handle the  `SASL_LIST_MECHANISMS_RESP` to get the mechanism supported by `server_negotiation`.  And then select a mechanism which is supported by both `client_negotiation` and `server_negotiation`.

```
 negotiation process:

                                                  client                             server
                                                     |---    SASL_LIST_MECHANISMS     -->|
                                                     |<--  SASL_LIST_MECHANISMS_RESP  ---|
                                                     |---   SASL_SELECT_MECHANISMS   --->|
                                                     |<-- SASL_SELECT_MECHANISMS_RESP ---|
                                                     |                                   |
                                                     |------     SASL_INITIATE    ------>|
                                                     |                                   |
                                                     |<----      SASL_CHALLENGE     -----|
                                                     |-----       SASL_RESPONSE     ---->|
                                                     |                                   |
                                                     |               .....               |
                                                     |                                   |
                                                     |<----      SASL_CHALLENGE     -----|
                                                     |-----       SASL_RESPONSE     ---->|
                                                     |                                   |        (authentication will succeed
                                                     |                                   |        if all chanllenges passed)
                                                     | <---         SASL_SUCC       -----|
    (client won't response                           |                                   |
     if servers says ok)                             |                                   |
                                                     | ----          RPC_CALL       ---> |
                                                     | <---          RPC_RESP        ----|
```


neverchanje (approved):

src/runtime/security/client_negotiation.cpp
```diff
@@ -36,18 +40,87 @@ void client_negotiation::start()
...
+        break;
+    default:
+        fail_negotiation();
+    }
+}
+
+void client_negotiation::recv_mechanisms(const negotiation_response &resp)
+{
+    if (resp.status != negotiation_status::type::SASL_LIST_MECHANISMS_RESP) {
+        ddebug_f("{}: get message({}) while expect({})",
```
acelyc111, line 83: `dwarn_f ` would be better because it lead failure.

src/runtime/security/client_negotiation.cpp
```diff
@@ -36,18 +40,87 @@ void client_negotiation::start()
...
+                 _name,
+                 enum_to_string(resp.status),
+                 enum_to_string(negotiation_status::type::SASL_LIST_MECHANISMS_RESP));
+        fail_negotiation();
+        return;
+    }
+
+    std::string match_mechanism = "";
```
acelyc111, line 91:
```suggestion
    std::string match_mechanism;
```

src/runtime/security/client_negotiation.cpp
```diff
@@ -36,18 +40,87 @@ void client_negotiation::start()
...
+
+    for (const std::string &server_support_mechanism : server_support_mechanisms) {
+        if (supported_mechanisms.find(server_support_mechanism) != supported_mechanisms.end()) {
+            match_mechanism = server_support_mechanism;
+            break;
+        }
+    }
+
+    if (match_mechanism.empty()) {
+        ddebug_f("server only support mechanisms of ({}), can't find expected ({})",
```
acelyc111, line 104: `dwarn_f ` too.

src/runtime/security/client_negotiation.cpp
levy5307: I think define a variable with initial value is ok

acelyc111: Default value of std::string is "", not need to assign it explicitly

neverchanje (approved):

acelyc111 (approved):

PR #597, opened 2020-08-20
refactor: move rpc handler on_query_configuration_by_index from meta_service to server_state 

levy5307:
In previous implementation, all rpc_handlers in meta server is registered in `meta_service`. And then the rpc request is forwarded to `server_state`、`bulk_load_service` and so on, to handle this request.
In this pull request, the rpc handler `on_query_configuration_by_index` is put to `server_state`, no need to forward by `meta_service`

### DIFF
--- a/src/meta/meta_service.h
+++ b/src/meta/meta_service.h
@@ -243,7 +243,6 @@ private:
     friend class test::test_checker;
     friend class meta_service_test_app;
     friend class bulk_load_service_test;
-    friend class server_state;
 
     replication_options _opts;
     meta_options _meta_opts;
--- a/src/meta/meta_service.h
+++ b/src/meta/meta_service.h
@@ -47,12 +47,12 @@
 #include "meta_backup_service.h"
 #include "meta_state_service_utils.h"
 #include "block_service/block_service_manager.h"
+#include "meta_server_failure_detector.h"
 
 namespace dsn {
 namespace replication {
 
 class server_state;
-class meta_server_failure_detector;
 class server_load_balancer;
 class meta_duplication_service;
 class meta_split_service;
@@ -128,13 +128,31 @@ public:
 
     dsn::task_tracker *tracker() { return &_tracker; }
 
+    template <typename TRpcHolder>
+    bool check_status(TRpcHolder rpc, /*out*/ rpc_address *forward_address = nullptr)
+    {
+        int result = check_leader(rpc, forward_address);
+        if (result == 0)
+            return false;
+        if (result == -1 || !_started) {
+            if (result == -1) {
+                rpc.response().err = ERR_FORWARD_TO_OTHERS;
+            } else if (_recovering) {
+                rpc.response().err = ERR_UNDER_RECOVERY;
+            } else {
+                rpc.response().err = ERR_SERVICE_NOT_ACTIVE;
+            }
+            ddebug("reject request with %s", rpc.response().err.to_string());
+            return false;
+        }
+
+        return true;
+    }
+
 private:
     void register_rpc_handlers();
     void register_ctrl_commands();
     void unregister_ctrl_commands();
-
-    // client => meta server
-    void on_query_configuration_by_index(configuration_query_by_index_rpc rpc);
 
     // partition server => meta server
     void on_config_sync(configuration_query_by_node_rpc rpc);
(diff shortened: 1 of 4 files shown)


src/meta/server_state.cpp
```diff
@@ -2830,5 +2835,42 @@ void server_state::clear_app_envs(const app_env_rpc &env_rpc)
                    new_envs.c_str());
         });
 }
+
+void server_state::register_rpc_handlers()
+{
+    static std::once_flag flag;
```
neverchanje, line 2841: What's call_once used for? What if more than one `server_state` instance is created?

src/meta/meta_service.h
```diff
@@ -195,9 +213,29 @@ class meta_service : public serverlet<meta_service>
     // if return -1 and `forward_address' != nullptr, then return leader by `forward_address'.
     int check_leader(dsn::message_ex *req, dsn::rpc_address *forward_address);
     template <typename TRpcHolder>
-    int check_leader(TRpcHolder rpc, /*out*/ rpc_address *forward_address);
-    template <typename TRpcHolder>
-    bool check_status(TRpcHolder rpc, /*out*/ rpc_address *forward_address = nullptr);
+    int check_leader(TRpcHolder rpc, rpc_address *forward_address)
```
neverchanje, line 216: What did this modification fix? Why move the implementation to the header file?

levy5307: If we don't move this to header file, the link will produce error. It seems it can't do the template instantiation work if we don't move it to header file.

src/meta/server_state.cpp
levy5307: If one rpc handler is registered twice more times, It will produce coredump. Maybe you can have a look of the implementation of `register_rpc_handler`:
```
bool rpc_server_dispatcher::register_rpc_handler(dsn::task_code code,
                                                 const char *extra_name,
                                                 const rpc_request_handler &h)
{
    std::unique_ptr<handler_entry> ctx(new handler_entry{code, extra_name, h});

    utils::auto_write_lock l(_handlers_lock);
    auto it = _handlers.find(code.to_string());
    auto it2 = _handlers.find(extra_name);
    if (it == _handlers.end() && it2 == _handlers.end()) {
        _handlers[code.to_string()] = ctx.get();
        _handlers[ctx->extra_name] = ctx.get();

        {
            utils::auto_write_lock l(_vhandlers[code.code()]->second);
            _vhandlers[code.code()]->first = std::move(ctx);
        }
        return true;
    } else {
        dassert(false, "rpc registration confliction for '%s' '%s'", code.to_string(), extra_name);
        return false;
    }
}
```

src/meta/meta_service.h
levy5307: I moved the implementation of `check_status` and `check_leader` out of class `meta_service`, but they are still being put in the header file.

src/meta/server_state.h
```diff
@@ -291,6 +291,12 @@ class server_state
     void process_one_partition(std::shared_ptr<app_state> &app);
     void transition_staging_state(std::shared_ptr<app_state> &app);
 
+    void register_rpc_handlers();
+    void unregister_rpc_handlers();
+
+    // client => meta server
+    void on_query_configuration_by_index(configuration_query_by_index_rpc rpc);
```
Shuo-Jia, line 298: > all rpc_handlers in meta server is registered in meta_service. And then the rpc request is forwarded to server_state、bulk_load_service and so on

I think the previous structure is clear, only move `on_query_configuration_by_index` into server_state breaking the design, and I didn't see some merit in it

levy5307: Break what desgin? I think the main logic is implemented by `server_state` but not `meta_service`. The work of `meta_service` is forwarding. And there are a lot of other rpc handlers about `backup_service`、 `split_service`, and so on.

Shuo-Jia: > ...meta_service. The work of meta_service is forwarding

So my means is: on_xxx_() in meta_service forward to xxx() in server_state should be an appropriate design

Shuo-Jia (dismissed):

src/meta/server_state.cpp
levy5307: Yes. In meta_test there maybe more than one `server_state`, so one same rpc handler may register more than one times.

src/runtime/rpc/rpc_engine.cpp
```diff
@@ -38,12 +38,18 @@
...
 #include <dsn/cpp/serialization.h>
 #include <dsn/utility/rand.h>
+#include <dsn/utility/flags.h>
 #include <set>
 
 namespace dsn {
 
 DEFINE_TASK_CODE(LPC_RPC_TIMEOUT, TASK_PRIORITY_COMMON, THREAD_POOL_DEFAULT)
 
+DSN_DEFINE_bool("core",
```
acelyc111, line 48: what does this config item used for? when to config true and when false?

levy5307: this configuration is used for unit test to disable register rpc handlers. Because in some unit tests, one same rpc handler will register twice more times, and this will produce coredump. So you can set this configuration to false to avoid this situation.

PR #598, opened 2020-08-20
refactor: use member func instead of macro RPC_CHECK_STATUS

levy5307:
In previous implementation, we use `RPC_CHECK_STATUS` to check the status of meta server. But In cpp, macro has no type check,  so we can't prevent it from calling with `rpc_holder`. But it will cause the response to be sent repeatedly if we do this.
So in this pull request, I reimplement it using member function(function has type check). 
I name this function as `check_status_with_msg`, but not `check_status`, because it may produce template function instantiated with ambiguity if we use `check_status`.


neverchanje (dismissed):

src/meta/meta_service.h
```diff
@@ -196,8 +196,14 @@ class meta_service : public serverlet<meta_service>
...
     template <typename TRpcHolder>
     int check_leader(TRpcHolder rpc, /*out*/ rpc_address *forward_address);
+    // ret:
+    //    false: check failed
+    //    true:  check succeed
     template <typename TRpcHolder>
     bool check_status(TRpcHolder rpc, /*out*/ rpc_address *forward_address = nullptr);
+    template <typename TRespType>
+    bool check_status_with_msg(message_ex *req, TRespType response_struct);
+
```
neverchanje, line 206:
```suggestion
    template <typename T, typename TRespType=std::remove_reference<T>::type>
    bool check_status_with_msg(message_ex *req, /*out*/ TRespType& response_struct);
```

neverchanje (approved):

PR #599, opened 2020-08-21
feat(security): init sasl

levy5307:
If the config of enable_auth is true, we should init sasl.


src/runtime/security/sasl_utils.cpp
```diff
@@ -0,0 +1,157 @@
...
+    }
+}
+
+int sasl_simple_logger(void *context, int level, const char *msg)
+{
+    if (SASL_LOG_NONE == level || nullptr == msg) {
+        return SASL_OK;
+    }
+
+    ddebug_f("sasl log info: log level = {}, message = {}", logger_level_to_string(level), msg);
```
acelyc111, line 64: All logs are in DEBUG level? I think it;s needed to convert `level` to related rdsn's log level.

levy5307: I have thought about this. But the log level in sasl is not a one-to-one correspondence log level in rdsn.
log level in rdsn:
```
typedef enum dsn_log_level_t {
    LOG_LEVEL_INFORMATION,
    LOG_LEVEL_DEBUG,
    LOG_LEVEL_WARNING,
    LOG_LEVEL_ERROR,
    LOG_LEVEL_FATAL,
    LOG_LEVEL_COUNT,
    LOG_LEVEL_INVALID
} dsn_log_level_t;
```
log level in sasl:
```
define SASL_LOG_NONE  0	/* don't log anything */
#define SASL_LOG_ERR   1	/* log unusual errors (default) */
#define SASL_LOG_FAIL  2	/* log all authentication failures */
#define SASL_LOG_WARN  3	/* log non-fatal warnings */
#define SASL_LOG_NOTE  4	/* more verbose than LOG_WARN */
#define SASL_LOG_DEBUG 5	/* more verbose than LOG_NOTE */
#define SASL_LOG_TRACE 6	/* traces of internal protocols */
#define SASL_LOG_PASS  7	/* traces of internal protocols, including
```

acelyc111: How about?
```
SASL_LOG_ERR -> LOG_LEVEL_ERROR
SASL_LOG_FAIL/SASL_LOG_WARN -> LOG_LEVEL_WARNING
SASL_LOG_NOTE -> LOG_LEVEL_INFORMATION
othres -> LOG_LEVEL_DEBUG
```

src/runtime/security/sasl_utils.cpp
```diff
@@ -0,0 +1,156 @@
...
+    sasl_set_mutex(&sasl_mutex_alloc_local,
+                   &sasl_mutex_lock_local,
+                   &sasl_mutex_unlock_local,
+                   &sasl_mutex_free_local);
+}
+
+error_s init_sasl(bool is_server)
+{
+    sasl_set_mutex_local();
+    int err = sasl_client_init(&client_callbacks[0]);
```
hycdong, line 136: if `is_server=true`, why should call `sasl_client_init`?

src/runtime/security/sasl_utils.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+int sasl_mutex_unlock_local(void *m)
+{
+    static_cast<utils::ex_lock_nr *>(m)->unlock();
+    return 0;
+}
+
+void sasl_set_mutex_local()
+{
+    sasl_set_mutex(&sasl_mutex_alloc_local,
```
hycdong, line 127: I don't find `sasl_set_mutex`, I guess it is a function of `sasl/sasl.h`, could you please add some comments for this function or tell us where to find this function?

src/runtime/security/sasl_utils.cpp
```diff
@@ -0,0 +1,156 @@
...
+        return SASL_BADPARAM;
+    }
+}
+
+sasl_callback_t client_callbacks[] = {
+    {SASL_CB_USER, (sasl_callback_ft)&sasl_get_username, nullptr},
+    {SASL_CB_GETPATH, (sasl_callback_ft)&sasl_get_path, nullptr},
+    {SASL_CB_AUTHNAME, (sasl_callback_ft)&sasl_get_username, nullptr},
+    {SASL_CB_LOG, (sasl_callback_ft)&sasl_simple_logger, nullptr},
+    {SASL_CB_LIST_END, nullptr, nullptr}};
```
hycdong, line 102: What the meaning of `LIST_END` in `SASL_CB_LIST_END`?

src/runtime/security/sasl_utils.cpp
levy5307: because server is also a client to other server. for example: replica server is the client of meta server.

src/runtime/security/sasl_utils.cpp
levy5307: it means the end of the list. It just used for a smybol to show it is the end of the list.

src/runtime/security/kinit_context.cpp
```diff
@@ -74,14 +74,15 @@ error_s check_configuration()
...
-    virtual ~kinit_context();
+    ~kinit_context();
 
     // implementation of 'kinit -k -t <keytab_file> <principal>'
     error_s kinit();
-    const std::string username() const { return _user_name; }
+    const std::string &username() const { return _user_name; }
 
 private:
+    kinit_context() : _opt(nullptr) {}
```
acelyc111, line 84: `_opt` is initialized as `nullptr ` in line 108, no need to re-initialize it here.

src/runtime/security/sasl_utils.cpp
hycdong: OK, it means that server (including meta server and replica server) is both client and server, and only c++ client is actually client, I suggest adding comments for it.

acelyc111 (approved):

src/runtime/security/sasl_utils.cpp
levy5307:
```
SASL_LOG_ERR -> LOG_LEVEL_ERROR
SASL_LOG_FAIL/SASL_LOG_WARN -> LOG_LEVEL_WARNING
SASL_LOG_NOTE -> LOG_LEVEL_DEBUG
othres -> LOG_LEVEL_INFORMATION

```

neverchanje (approved):

hycdong (approved):

PR #601, opened 2020-08-24
feat(security):  make failure detector rpcs in white list

levy5307:
If the failure detector rpc is not in white list, this rpc will be rejected when the negotiation is in process.
So if in `lease period`  the negotiation is not succeed, the replica server will switch master to another meta server. And then start negotiation with the new meta server. In finally, the negotiation with the new meta server will be failed too.


acelyc111 (approved):

acelyc111 (approved):

include/dsn/tool-api/network.h
```diff
@@ -304,6 +304,7 @@ class rpc_session : public ref_counter
     void clear_send_queue(bool resend_msgs);
     bool on_disconnected(bool is_write);
     bool is_auth_success(message_ex *msg);
+    bool in_security_white_list(task_code code);
```
neverchanje, line 307:
```
bool rpc_session::in_security_white_list(task_code code)
{
    return security::is_negotiation_message(code) || fd::is_failure_detector_message(code);
}
```

This is totally static. Why make it a method rather than a free function?

```
namespace security {
bool in_white_list(task_code code)
{
    return is_negotiation_message(code) || fd::is_failure_detector_message(code);
}
} // 
```

src/runtime/rpc/network.cpp
```diff
@@ -281,7 +287,7 @@ void rpc_session::send_message(message_ex *msg)
         // Attention: here we only allow two cases to send message:
         //  case 1: session's state is SS_CONNECTED
         //  case 2: session is sending negotiation message
-        if ((SS_CONNECTED == _connect_state || security::is_negotiation_message(msg->rpc_code())) &&
+        if ((SS_CONNECTED == _connect_state || security::in_white_list(msg->rpc_code())) &&
```
hycdong, line 290: I suggest that you should update the comment, because a new case has been added.

src/runtime/rpc/network.cpp
```diff
@@ -42,6 +43,11 @@ namespace dsn {
 
 namespace security {
 DSN_DECLARE_bool(enable_auth);
+
+bool in_white_list(task_code code)
+{
+    return is_negotiation_message(code) || fd::is_failure_detector_message(code);
+}
```
hycdong, line 50: In my view, `in_white_list` is not a clear function name to show failure detector rpc needn't negotiation, I think it is better to use `is_negotiation_message(code) || fd::is_failure_detector_message(code)` directly in function `send_message` and `on_recv_message`, rather than the function `in_white_list`.

src/runtime/rpc/network.cpp
levy5307: Yes, you are right. sorry I forgot it.

src/runtime/rpc/network.cpp
levy5307: I don't think so. White list means the rpc needn't negotiation.

acelyc111 (approved):

src/runtime/rpc/network.cpp
neverchanje: What about `in_negotiation_white_list`? I agree with @hycdong. The name doesn't point out the white list for what.

src/runtime/rpc/network.cpp
```diff
@@ -42,6 +43,11 @@ namespace dsn {
 
 namespace security {
 DSN_DECLARE_bool(enable_auth);
+
+bool in_white_list(task_code code)
```
neverchanje, line 47:
```suggestion
// Whether `code` is in the set of RPCs that are out of the security control.
bool in_white_list(task_code code)
```

src/runtime/rpc/network.cpp
levy5307: The accurate  name is `in_security_white_list`, which means these rpcs doesn't need to do security process. 
But I think the name with namespace `security::in_white_list` is engough to represent it.

hycdong: I agree with @neverchanje , I think `in_negotiation_white_list` may be better, and `in_negotiation_allow_list` is much better.

PR #604, opened 2020-08-26
feat(security): implement sasl_wrapper which is a wrapper to operate sasl

levy5307:
implement `sasl_wrapper`


src/runtime/security/kinit_context.cpp
```diff
@@ -45,6 +45,8 @@ DSN_DECLARE_bool(enable_auth);
 DSN_DEFINE_string("security", krb5_keytab, "", "absolute path of keytab file");
 DSN_DEFINE_string("security", krb5_config, "", "absolute path of krb5_config file");
 DSN_DEFINE_string("security", krb5_principal, "", "kerberos principal");
+DSN_DEFINE_string("security", service_fqdn, "", "service fqdn");
```
hycdong, line 48: How about adding some comments about the usage of `fqdn`?

levy5307: I add it to the desc param of `DSN_DEFINE_string`

neverchanje (approved):

neverchanje (approved):

acelyc111 (approved):

PR #606, opened 2020-08-27
refactor(security): close the connection when something is wrong with sever_negotiation

levy5307:
In previous implementation, when something is wrong in `server_negotiation`, it send a message to `client_negotiation`, and the `client_negotiation` will close the connection when it receives the message.
So I think we can directly close the connection in this case, no need to send a message.

### DIFF
--- a/src/runtime/security/server_negotiation.cpp
+++ b/src/runtime/security/server_negotiation.cpp
@@ -50,7 +50,7 @@ void server_negotiation::handle_request(negotiation_rpc rpc)
         // TBD(zlw)
         break;
     default:
-        fail_negotiation(rpc, "wrong status");
+        fail_negotiation();
     }
 }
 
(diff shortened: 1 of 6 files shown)


acelyc111 (approved):

src/runtime/security/server_negotiation.cpp
```diff
@@ -112,12 +112,5 @@ error_s server_negotiation::do_sasl_server_init()
     return error_s::make(ERR_OK);
 }
 
-void server_negotiation::fail_negotiation(negotiation_rpc rpc, const std::string &reason)
-{
-    negotiation_response &response = rpc.response();
-    _status = response.status = negotiation_status::type::SASL_AUTH_FAIL;
-    response.msg = reason;
```
hycdong, line 119: In previous implementation, there will be error reason. Is it totally useless?

levy5307: It is only used for sending a message to client to show why it is failed. I think it is useless

hycdong (approved):

PR #594, opened 2020-08-15
refactor(http): replace service/method with simple HTTP path

neverchanje:
This PR I replace `http_request::service_method` with a simple URL path.

```diff
struct http_request
{
    static error_with<http_request> parse(dsn::message_ex *m);

-    // http://ip:port/<service>/<method>
-    std::pair<std::string, std::string> service_method;
+    std::string path;
```

And I decouple the registration of HTTP calls from `http_server` to `http_call_registry`.

- `http_server` is a special RPC server that handles HTTP requests.
- `http_call_registry` manages the HTTP calls.

I will in later PR provide a simpler HTTP service API based on  `http_call_registry`

PR #610, opened 2020-08-31
refactor: hide http_server from http api

neverchanje:
This PR simplifies the registration of HTTP services. Previously, to enable HTTP you must create an instance of `http_server`, because it's required for 1. network RPC 2. service registration. 

In https://github.com/XiaoMi/rdsn/pull/594 I decoupled service registration from `http_server` into a new class `http_call_registry`. Consider that `http_server` is now responsible only for network RPC, it should be a singleton in every case. 

If `http_server` is a singleton, we can hide it from the http api, and expose a pure function:

```cpp
void start_http_server();
```


src/http/test/http_server_test.cpp
```diff
@@ -8,6 +8,7 @@
 #include "http/http_message_parser.h"
 #include "http/root_http_service.h"
 #include "http/server_info_http_services.h"
+#include "http/http_server_impl.h"
```
levy5307, line 11: Why did you add this line? I think it's useless

src/http/http_server_impl.h
```diff
@@ -0,0 +1,47 @@
...
+
+#pragma once
+
+#include <dsn/tool-api/rpc_message.h>
+#include <dsn/cpp/serverlet.h>
+#include <dsn/http/http_server.h>
+
+namespace dsn {
+
+class http_server : public serverlet<http_server>
```
levy5307, line 26: If it's a singleton, why did you not make it inheritance from `utils::singleton`?

src/http/http_server.cpp
```diff
@@ -178,25 +178,41 @@ void http_server::add_service(http_service *service)
...
+/*extern*/ void start_http_server()
+{
+    // starts http server as a singleton
+    static http_server server;
+}
+
+/*extern*/ void register_http_service(http_service *svc)
+{
+    // simply hosting the memory of these http services.
+    static std::vector<std::shared_ptr<http_service>> services_holder;
```
levy5307, line 211: `std::unique_ptr`

src/http/http_server_impl.h
```diff
@@ -0,0 +1,47 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#pragma once
+
+#include <dsn/tool-api/rpc_message.h>
```
levy5307, line 20: I think there is no need to include `rpc_message.h`. use predeclare is ok

src/http/http_server_impl.h
```diff
@@ -0,0 +1,47 @@
...
+    http_server();
+
+    ~http_server() override = default;
+
+    void add_service(http_service *service);
+
+    void serve(message_ex *msg);
+
+private:
+    std::map<std::string, std::unique_ptr<http_service>> _service_map;
```
levy5307, line 38: in you previous implementation, I think `_service_map` is useless now.

src/http/http_server.cpp
```diff
@@ -13,6 +13,7 @@
 #include "perf_counter_http_service.h"
 #include "uri_decoder.h"
 #include "http_call_registry.h"
+#include "http_server_impl.h"
 
 namespace dsn {
```
levy5307, line 19: Could you please replace all of the `add_service()` in this file?

src/http/http_server.cpp
```diff
@@ -178,25 +178,41 @@ void http_server::add_service(http_service *service)
...
     writer.flush();
 
-    return resp;
+    dsn_rpc_reply(resp_msg.get());
+}
+
+/*extern*/ void start_http_server()
+{
+    // starts http server as a singleton
+    static http_server server;
```
acelyc111, line 205: After http_server object constructed, then it can serve request immediatly? without calling any functions like 'start_work'?

src/http/http_server_impl.h
neverchanje: This is a impl header, not a api header. It will not lead to any more compilation.

src/http/test/http_server_test.cpp
neverchanje: It's not useless. It imports the class `http_server`.

src/http/http_server_impl.h
neverchanje: I would like to use the least possible code. There're ways to implement one semantic in c++. I just choose whatever that can work.

src/http/http_server.cpp
neverchanje: No. There's no such `start_work` thing. The http_server actually does not manage the TCP conns. It only registers the RPC code to the underlying RPC manager.

acelyc111 (approved):

levy5307 (approved):

PR #615, opened 2020-09-07
refactor(http): add http handler without creating http service

neverchanje:
## What problem do I solve?

To register an HTTP handler, we previously needed to define an `http_service` subclass before writing the exact handler.

```cpp
class recent_start_time_http_service : public http_service
{
public:
    recent_start_time_http_service()
    {
        register_handler("",
                         std::bind(&recent_start_time_http_service::get_recent_start_time_handler,
                                   this,
                                   std::placeholders::_1,
                                   std::placeholders::_2),
                         "ip:port/recentStartTime");
    }

    std::string path() const override { return "recentStartTime"; }

    void get_recent_start_time_handler(const http_request &req, http_response &resp);
};
```

As we can see, `http_service` is too heavy to implement a lightweight handler.

## How do I solve this problem

We can simplify this short type handler by using `register_http_call`, a new API of the library.

```cpp
    register_http_call("recentStartTime")
        .with_callback([](const http_request &req, http_response &resp) {
            get_recent_start_time_handler(req, resp);
        })
        .with_help("Gets the server start time.");
```

Remember that we use `register_http_service` to register an `http_sevice`. I did to `http_call` likewise. `register_http_call` is designed to be a convenience utility that could boost our future development of new HTTP API.

## Known issue

In this PR I removed `version_http_service`. This http call will be registered during a `pegasus_service_app` is created. So I will commit another PR for this.


src/http/http_server.cpp
```diff
@@ -38,6 +38,21 @@ DSN_DEFINE_bool("http", enable_http_server, true, "whether to enable the embedde
     }
 }
 
+/*extern*/ http_call &register_http_call(std::string full_path)
+{
+    auto call_ptr = dsn::make_unique<http_call>();
+    call_ptr->path = std::move(full_path);
+    http_call &call = *call_ptr;
+    // register this call
```
levy5307, line 46: I think the comment doesn't make sense. Maybe you can remove it.

neverchanje: Sure. Any other problems?

levy5307: no

src/http/http_server.cpp
```diff
@@ -64,13 +79,7 @@ http_server::http_server() : serverlet<http_server>("http_server")
     tools::register_message_header_parser<http_message_parser>(NET_HDR_HTTP, {"GET ", "POST"});
 
     // add builtin services
-    register_http_service(new root_http_service());
-
-#ifdef DSN_ENABLE_GPERF
-    register_http_service(new pprof_http_service());
```
hycdong, line 70: I don't find out where to register pprof http serivce, do we need to register it?

include/dsn/dist/replication/meta_service_app.h
```diff
@@ -59,7 +59,6 @@ class meta_service_app : public service_app
 private:
     friend class ::dsn::replication::test::test_checker;
     std::unique_ptr<dsn::replication::meta_service> _service;
-    version_http_service *_version_http_service;
```
hycdong, line 62: I see that this pull request remove `version_http_service`, is it useless?

src/http/http_server.cpp
neverchanje:
```
/*extern*/ void register_builtin_http_calls()
{
#ifdef DSN_ENABLE_GPERF
    static pprof_http_service pprof_svc;
#endif
```

No. The constructor of pprof_http_service actually does the registration.

include/dsn/dist/replication/meta_service_app.h
neverchanje: No. I will create `version_http_service` in pegasus side. I've included "Known issue" section on my description.

hycdong (approved):

levy5307 (approved):

PR #616, opened 2020-09-07
fix(security): fix bug in building cyrus-sasl

levy5307:
fix bug in building cyrus-sasl


src/runtime/test/config-test.ini
```diff
@@ -133,3 +133,14 @@ run = true
 [uri-resolver.http://localhost:8080]
 factory = partition_resolver_simple
 arguments = 127.0.0.1:8080
+
+[security]
+  krb5_keytab = /etc/pegasus_tst-pegasus.keytab
```
acelyc111, line 138: These files may not exist on test environment, will the test cases fail?

levy5307: Yes, these configs is used for init kerberos. Or we should delete this unit test.
By the way, I prefer to delete the unit test

neverchanje (approved):

Shuo-Jia (approved):

PR #620, opened 2020-09-09
refactor(security): refactor the send func of client_negotiation

levy5307:
refactor the send func of client_negotiation


src/runtime/security/client_negotiation.cpp
```diff
@@ -129,10 +128,8 @@ void client_negotiation::on_mechanism_selected(const negotiation_response &resp)
     std::string start_output;
     err_s = _sasl->start(_selected_mechanism, "", start_output);
     if (err_s.is_ok() || ERR_SASL_INCOMPLETE == err_s.code()) {
-        auto req = dsn::make_unique<negotiation_request>();
-        _status = req->status = negotiation_status::type::SASL_INITIATE;
-        req->msg = start_output;
-        send(std::move(req));
+        _status = negotiation_status::type::SASL_INITIATE;
+        send(_status, start_output);
```
neverchanje, line 132: Use std::move here if you do not want a string copy.

```cpp
void send(..., std::string msg) { // use value-typed string, allowing string to be move constructed.
```

neverchanje (approved):

src/runtime/security/client_negotiation.cpp
```diff
@@ -172,16 +167,18 @@ void client_negotiation::on_challenge(const negotiation_response &challenge)
 void client_negotiation::select_mechanism(const std::string &mechanism)
 {
     _selected_mechanism = mechanism;
+    _status = negotiation_status::type::SASL_SELECT_MECHANISMS;
 
-    auto req = dsn::make_unique<negotiation_request>();
-    _status = req->status = negotiation_status::type::SASL_SELECT_MECHANISMS;
-    req->msg = mechanism;
-    send(std::move(req));
+    send(_status, std::move(mechanism));
```
acelyc111, line 172: `mechanism` is a 'const &', is it movable? there is not any compile complains?

levy5307: I didn't see any compile complains. And I write a test program like this:
```
    std::string test = "123123";
    const std::string &ref_test = test;
    const std::string &&r_test = std::move(ref_test);
    std::cout << r_test << std::endl;
```
It works well.

acelyc111 (approved):

PR #621, opened 2020-09-09
fix: set the tracer type when init

Shuo-Jia:
`_is_sub` control whether to dump the point when deconstructer, if one tracer isn't marked as `is_sub` at initialization, but wait `set_sub_tracer`, which will cause tracer that hoped `is_sub` but not be `set_sub_tracer` dump points. for example, `learn` and `log write` process will create `aio_task`, but only `log write` process `set_sub_tracer`,  the result is that `aio_task` of `learn` will dump points, which is not expected. So if we hope one tracer `is_sub`, we should init it at first constructer

### DIFF
--- a/include/dsn/utils/latency_tracer.h
+++ b/include/dsn/utils/latency_tracer.h
@@ -72,7 +72,7 @@ public:
     //  threshold < 0: don't dump any trace points
     //  threshold = 0: dump all trace points
     //  threshold > 0: dump the trace point when time_used > threshold
-    latency_tracer(const std::string &name, uint64_t threshold = 0);
+    latency_tracer(const std::string &name, bool is_sub = false, uint64_t threshold = 0);
 
     ~latency_tracer();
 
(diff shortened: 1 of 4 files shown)


acelyc111 (dismissed):

include/dsn/utils/latency_tracer.h
```diff
@@ -72,7 +72,7 @@ class latency_tracer
     //  threshold < 0: don't dump any trace points
     //  threshold = 0: dump all trace points
     //  threshold > 0: dump the trace point when time_used > threshold
-    latency_tracer(const std::string &name, uint64_t threshold = 0);
+    latency_tracer(const std::string &name, bool is_sub = false, uint64_t threshold = 0);
```
hycdong, line 75: I suggest adding some comments for the new `is_sub` parameter.

include/dsn/utils/latency_tracer.h
```diff
@@ -69,10 +69,14 @@ DSN_DECLARE_bool(enable_latency_tracer);
 class latency_tracer
 {
 public:
+    //-is_sub:
+    //  if `is_sub`=true means its points will be dumped by parent tracer and won't be dumped
+    //  repeatedly when decontructor
```
levy5307, line 74: when it destructs

hycdong (approved):

PR #626, opened 2020-09-15
refactor(security): add join point to start negotiation

levy5307:
Previously, the network module is coupled from security, because we should start negotiation when rpc session is connected. To make network decoupled from security,  in this pr, I use joint point for starting negotiation. And if you enable authentiation, a joint point, which is implemented in security module, is put_back to  `on_rpc_session_connected`. 

And to make network  module simply, I removed the `SS_NEGOTIATING`. Because I think the authentication code should not  invade network modules


src/runtime/security/negotiation_service.cpp
```diff
@@ -44,10 +48,42 @@ void negotiation_service::on_negotiation_request(negotiation_rpc rpc)
...
+}
+
+void negotiation_service::on_rpc_disconnected(rpc_session *session)
+{
+    {
+        utils::auto_lock<utils::ex_lock_nr> l(_lock);
+        const auto iter = _negotiations.find(session);
+        if (iter != _negotiations.end()) {
+            _negotiations.erase(iter);
+        }
```
acelyc111, line 77: Why not erase session directly?

src/runtime/security/negotiation_service.cpp
```diff
@@ -44,10 +48,42 @@ void negotiation_service::on_negotiation_request(negotiation_rpc rpc)
...
+    }
+}
+
+void init_join_point()
+{
+    rpc_session::on_rpc_session_connected.put_back(negotiation_service::on_rpc_connected,
+                                                   "security");
+    rpc_session::on_rpc_session_disconnected.put_back(negotiation_service::on_rpc_disconnected,
+                                                      "security");
```
acelyc111, line 86: 👍

src/runtime/security/negotiation_service.cpp
```diff
@@ -44,10 +48,42 @@ void negotiation_service::on_negotiation_request(negotiation_rpc rpc)
         return;
     }
 
-    server_negotiation *srv_negotiation =
-        static_cast<server_negotiation *>(rpc.dsn_request()->io_session->get_negotiation());
+    server_negotiation *srv_negotiation = nullptr;
+    {
+        utils::auto_lock<utils::ex_lock_nr> l(_lock);
+        srv_negotiation =
```
acelyc111, line 54: Is it safe to add an `assert` to check the session is not exist in the map?

src/runtime/rpc/network.cpp
```diff
@@ -74,9 +74,7 @@ void rpc_session::set_connected()
 
     {
         utils::auto_lock<utils::ex_lock_nr> l(_lock);
-        dassert((_connect_state == SS_NEGOTIATING && security::FLAGS_enable_auth) ||
-                    (_connect_state == SS_CONNECTING && !security::FLAGS_enable_auth),
-                "wrong session state");
+        dassert(_connect_state == SS_CONNECTING, "wrong session state");
```
acelyc111, line 77: Use `dcheck_eq` then you can get exact value of `_connect_state ` when check failed.

src/runtime/rpc/network.cpp
```diff
@@ -478,29 +457,24 @@ bool rpc_session::on_recv_message(message_ex *msg, int delay_ms)
...
+    // todo(zlw): resend pending messages when negotiation is succeed
 }
 
-void rpc_session::auth_negotiation()
+bool rpc_session::is_negotiation_succeed()
 {
-    _negotiation = security::create_negotiation(is_client(), this);
-    _negotiation->start();
+    if (negotiation_succeed) {
+        return negotiation_succeed;
```
acelyc111, line 471: Why these sentense out of lock?

include/dsn/tool-api/network.h
```diff
@@ -267,13 +267,15 @@ class rpc_session : public ref_counter
...
     {
         SS_CONNECTING,
-        SS_NEGOTIATING,
         SS_CONNECTED,
         SS_DISCONNECTED
     };
     ::dsn::utils::ex_lock_nr _lock; // [
     volatile session_state _connect_state;
 
+    bool negotiation_succeed = false;
```
acelyc111, line 276: Is it need to leave it in the big lock, how about use atomic? Lock too many variables should be very careful to avoid dead lock problems.

include/dsn/tool-api/network.h
```diff
@@ -232,9 +232,10 @@ class rpc_session : public ref_counter
     bool delay_recv(int delay_ms);
     bool on_recv_message(message_ex *msg, int delay_ms);
 
-    /// for negotiation
-    void start_negotiation();
-    security::negotiation *get_negotiation() const;
+    /// interfaces for security authentication,
+    /// you can ignore them if you don't enable auth
+    void set_negotiation_succeed();
+    bool is_negotiation_succeed();
```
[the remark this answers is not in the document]
levy5307: we can't use const here, because the implementation of `is_negotiation_succeed ` has to lock `_lock`, which is a menber of `rpc_session`

src/runtime/security/negotiation_service.cpp
levy5307: Yes, it's safe, we can add a `assert` here. Because only a `rpc_session` who have a corresponding `negotiation` can send negotiation message

include/dsn/tool-api/network.h
levy5307: Yes, if we only add `negotiation_succeed ` here, there is no need to lock. But I will add a resend message queue, which is the messages sent in the duration of negotiation.  I should to ensure the access of `negotiatioin_succeed` and the resend message queue to be atomic.

src/runtime/rpc/network.cpp
levy5307: this is kind of double check. the first one didn't lock the `_lock`.
Because `negotiation_succeed` only transfered from `false` to `true`. So if it is true now, it will not change in the later. but if it is `false` now, maybe it will change soon. So we should use lock to protect it.
Why I use double check here is to be efficient

include/dsn/tool-api/network.h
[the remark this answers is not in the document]
acelyc111: In this case, we often define lock as a mutable member.

src/runtime/rpc/network.cpp
acelyc111: OK, better to add some comments here

src/runtime/security/negotiation_service.cpp
```diff
@@ -44,10 +48,43 @@ void negotiation_service::on_negotiation_request(negotiation_rpc rpc)
...
     srv_negotiation->handle_request(rpc);
 }
 
+void negotiation_service::on_rpc_connected(rpc_session *session)
+{
+    std::unique_ptr<negotiation> nego = security::create_negotiation(session->is_client(), session);
+    nego->start();
+    {
+        zauto_write_lock l(_lock);
+        _negotiations[session] = std::move(nego);
```
acelyc111, line 70: Is it safe to add an assert to check the session is not exist in the map?

levy5307: I think it's ok to not add assert here. If a rpc session connects twice, we should add a new negotiation, and the old one will be destroyed when `_negotiations[session] = std::move(nego)`

acelyc111 (dismissed):

src/runtime/rpc/network.cpp
levy5307: I removed the double check and the lock operation in `is_negotiation_succeed`. Because I think it is ok to get the value of `negotiation_succeed` without lock. 
The double check will be used in `set_negotiation_succeed` in the later pull request.

src/runtime/security/negotiation_service.h
```diff
@@ -20,22 +20,32 @@
...
 
 private:
     negotiation_service();
     void on_negotiation_request(negotiation_rpc rpc);
     friend class utils::singleton<negotiation_service>;
     friend class negotiation_service_test;
+
+    static zrwlock_nr _lock; // [
+    static negotiation_map _negotiations;
+    //]
```
hycdong, line 46: Why add '//]' here?

levy5307: `//[` and `//]` indicates that  the _lock is used for protecting the members locates between `[` and `]`

src/runtime/rpc/network.cpp
```diff
@@ -478,28 +457,15 @@ bool rpc_session::on_recv_message(message_ex *msg, int delay_ms)
...
-            set_negotiation();
-        }
-
-        auth_negotiation();
-    } else {
-        // set negotiation success if auth is disabled
-        on_success();
-    }
-}
+    utils::auto_lock<utils::ex_lock_nr> l(_lock);
```
acelyc111, line 462: You should also remove lock here?

src/runtime/rpc/network.cpp
acelyc111: Is `on_recv_message` in the same lock?

hycdong (approved):

src/runtime/rpc/network.cpp
levy5307: on_recv_message will not lock any locks.  It only calls `is_negotiation_succeed` to judge whether the negotiation is succeed

levy5307: Another better way to implement it is to change the type of `_lock` from `utils::ex_lock_nr` to `zrwlock_nr`. And in `is_negotiation_succeed` we use `zauto_read_lock l(_lock)` to lock read. And in other cases we should use `zauto_write_lock l(_lock)`

acelyc111 (approved):

src/runtime/rpc/network.cpp
```diff
@@ -400,7 +387,7 @@ bool rpc_session::on_disconnected(bool is_write)
 
 bool rpc_session::is_auth_success(message_ex *msg)
 {
-    if (security::FLAGS_enable_auth && !_negotiation->negotiation_succeed()) {
+    if (security::FLAGS_enable_auth && !is_negotiation_succeed()) {
```
neverchanje, line 390: From my point, if you want to decouple the negotiation entirely with `rpc_session`, you should remove this function and modify `rpc_session::on_recv_message`, where a session rejects the incoming requests when the negotiation is unfinished.

Nevertheless, there's no joint point for recv_message as far as I see. (There's one joint point at `task_spec::on_rpc_call`, but not suitable since it targets at customization of tasks, rather than general rpc control).

Maybe you can add such joint point later.

levy5307: Yes, I will do it in the next pull request. I split the refactor of network into 3 pull requests.

neverchanje (approved):

acelyc111 (approved):

PR #629, opened 2020-09-17
refactor(security): add join point to filter the sending message

levy5307:
Add join point to filter the sending message. If a message is filtered by `negotiation_service::on_rpc_send_msg`, we will put it into a pending message queue, which name is `_pending_messages` in `rpc_session`, and resend it when the negotiation is succeed.


acelyc111:
I found these PRs will add many if-statements, it's better to add some branch predictions, i.e. `dsn_likely`, `dsn_unlikely`

src/runtime/rpc/network.cpp
```diff
@@ -437,12 +440,45 @@ bool rpc_session::on_recv_message(message_ex *msg, int delay_ms)
...
+        negotiation_succeed = true;
+
+        _pending_messages.swap(swapped_pending_msgs);
+    }
 
-    // todo(zlw): resend pending messages when negotiation is succeed
+    // resend the pending messages
+    for (auto msg : swapped_pending_msgs) {
+        send_message(msg);
+        msg->release_ref();
```
hycdong, line 480: Is it safe to resend message without interval? In other words, will resending messages increase network load?

levy5307: The negotiation is succeed in about serveral seconds, it won't be too many messages pending.
And I think the work load is guaranteed by rate limiter of server.

acelyc111 (approved):

hycdong (approved):

include/dsn/tool-api/network.h
```diff
@@ -275,7 +281,10 @@ class rpc_session : public ref_counter
     volatile session_state _connect_state;
 
     bool negotiation_succeed = false;
-    // TODO(zlw): add send pending message
+    // when the negotiation of a session isn't succeed,
+    // all messages are queued in _pending_messages.
+    // after connected, all of them are moved to "_messages"
+    std::vector<message_ex *> _pending_messages;
```
neverchanje, line 287: Why need this pending queue of messages? The clients can buffer its messages before the negotiation completes. I think this is an over-optimzation.

PR #631, opened 2020-09-22
refactor(backup): straighten the execution path of download_file

neverchanje:
The block_service API provides only async interfaces. It causes the callers must write code from bottom-up unnaturally.
This PR turns the async calls (`block_file::download`) in `block_service_manager::download_file` into sync functions (`download_block_file_sync`).
It doesn't change the original behavior, because `download_file` is actually synchronous at all. But it reverses the code to top-down, which is more readable for humans and more debuggable.

Another significant change in this PR is that `download_file` does not verify checksum with remote FS anymore. 

A md5 verification of local file with remote:
```cpp
        // redundant process:
        std::string current_md5;
        error_code e = utils::filesystem::md5sum(local_file_name, current_md5);
        if (e != ERR_OK) {
            derror_f("calculate file({}) md5 failed", local_file_name);
            download_err = e;
            return;
        }
        if (current_md5 != bf->get_md5sum()) { // md5 of the remote file 
            derror_f("local file({}) is different from remote file({}), download failed, md5: "
                     "local({}) VS remote({})",
                     local_file_name,
                     bf->file_name(),
                     current_md5,
                     bf->get_md5sum());
            download_err = ERR_CORRUPTION;
            return;
        }
```

Since in both restore and bulk-load, there's a meta file storing the checksums, which we use to compare with the local file checksum after the download completes. We do not need another verification between the local file and the remote file. 

The remote file md5 is generally provided by the filesystem service. 
- aws s3: https://aws.amazon.com/cn/premiumsupport/knowledge-center/data-integrity-s3/
- aliyun oss: https://help.aliyun.com/knowledge_detail/39533.html

But it doesn't mean that every filesystem has such capability to produce md5. Some may use by native an uncommon checksum algorithm, HDFS for example: https://blog.woopi.org/wordpress/files/hadoop-2.6.0-javadoc/org/apache/hadoop/fs/MD5MD5CRC32FileChecksum.html.

So md5 should not be in our presumption of the filesystem abstraction. I may remove `block_file::get_md5sum` in future PR.


src/block_service/block_service_manager.cpp
```diff
@@ -84,133 +84,76 @@ block_filesystem *block_service_manager::get_block_filesystem(const std::string
...
-            } else {
-                download_err = ERR_OK;
-                download_file_size = bf->get_size();
-                ddebug_f("local file({}) has been downloaded, file size = {}",
-                         local_file_name,
-                         download_file_size);
-                return;
-            }
+    // Create a block_file object.
+    const std::string remote_file_name = utils::filesystem::path_combine(remote_dir, file_name);
```
levy5307, line 130: const std::string &remote_file_name

neverchanje: remote_file_name is a new value, not a reference to an existing value.

levy5307: so why did you use `  const std::string &local_file_name` in L121?

hycdong (approved):

src/block_service/block_service_manager.cpp
```diff
@@ -84,133 +84,76 @@ block_filesystem *block_service_manager::get_block_filesystem(const std::string
...
+    auto create_resp =
+        create_block_file_sync(remote_file_name, false /*ignore file meta*/, fs, &tracker);
+    error_code err = create_resp.err;
+    if (err != ERR_OK) {
+        derror_f("create file({}) failed with error({})", remote_file_name, err.to_string());
+        return err;
+    }
+    block_file *bf = create_resp.file_handle.get();
```
zhangyifan27, line 138: use `block_file_ptr` ?

zhangyifan27 (approved):

acelyc111 (approved):

PR #633, opened 2020-09-27
perf: optimizing write latency using independent IO queues replace of libaio

Shuo-Jia:
This PR is based on #569, thanks for @neverchanje offering an initial implementation of the new async-io.

The original async-io is based on [Linux AIO](https://github.com/littledan/linux-aio) which may cause some bottlenecks especially in **Learning** and **Compaction** situations which will slow down the write-path.  #568 has changed the callback task pool of LibAIO to optimize the latency during **Compaction**, but we still find latency spikes occasionally occur during **Compaction**, and it generally makes the cluster unavailable during **Learning** (the replica migration process typically happen during scaling-in-out nodes).

To avoid the influence of different AIO tasks sharing the same LibAIO queue, this PR removes the LibAIO module and introduces a new async-IO implementation based on rdsn task-queue. It handles different IO tasks in separate task queues, isolates low & high priority tasks.

In the latest design, the Learning IO task is assigned into `THREAD_POOL_DEFAULT` queue, the private-log IO task is assigned to  `THREAD_POOL_REPLICATION_LONG` queue, shared-log IO task is assigned to `THREAD_POOL_SLOG` queue. With the isolation of different IO queues, every io task can be executed efficiently.

**In addition**, for the [Learning request logic](https://github.com/XiaoMi/rdsn/blob/2597bfcd245bfe9f6b74bce2e3a48bb7a4b89896/src/nfs/nfs_client_impl.cpp#L215) may be assigned `THREAD_POOL_REPLICATION` when after [end_get_file_size](https://github.com/XiaoMi/rdsn/blob/2597bfcd245bfe9f6b74bce2e3a48bb7a4b89896/src/nfs/nfs_client_impl.cpp#L145)(for the `end_get_file_size` is assigned into `THREAD_POOL_REPLICATION`), which may block the write operation especially using [NFS-RateLimiter](https://github.com/XiaoMi/rdsn/pull/461) set low learning rate. Since the change is little, I fix it in this PR:

```diff
nfs_client_impl::end_get_file_size(){
...
- continue_copy();
+ tasking::enqueue(LPC_NFS_COPY_FILE, nullptr, [this]() { continue_copy(); }, 0);
}
```

And `LPC_NFS_COPY_FILE` is assigned into `THREAD_POOL_DEFAULT`

## Experiments

Here I tested cases during Learning and Compaction to see the effect of optimization. The pegasus configuration is 2 meta-server and 5 replica-server.

**Case 1:** Compaction-30thread*3client, load, 1KB length

#568 has shown the effect of most cases. This test shows the average result of multiple tests. In this test, I execute three times using YCSB.

io/latency | min | average | p95 | p99 | p999 | p9999 | max
-- | -- | -- | -- | -- | -- | -- | --
LibAIO | 340 | 1578 | 3157 | 6423 | 11119 | **325897** | **1291624**
NewAIO | 335 | 1558 | 3139 | 6343 | 10175 | **18079** | **177663**

We can find them achieving almost the same results below P9999. However, at P9999 and MAX results, `NewAIO` provides a better result.

**Case 2**: Learning of adding node-15thread*3client, write:read=3:1, 1KB length, data=16GB * 32 Partitions
LibAIOAddNodeLatency
NewAIOAddNodeLatency

**Case 3**: Learning of offline node-15thread*3client, write:read=3:1, 1KB length, data=16GB * 32 Partitions
LibAIOOfflineLatency
NewAIO500MBP999

From the above results, we can confirm that the new async-io implementation almost avoids the spikes during adding node, and greatly reduced the influence while offlining node.

Notice that not all the tests can have same exact latency value, but the conclusions of multiple tests are consistent.


src/aio/aio_task_queue.h
```diff
@@ -0,0 +1,170 @@
...
+
+struct io_event_t
+{
+public:
+    io_event_t(aio_task *tsk, bool async, aio_provider *provider) : _tsk(tsk), _provider(provider)
+    {
+        if (!async) {
+            _notify = make_unique<utils::notify_event>();
+        }
+    }
```
levy5307, line 46: add a blank line

src/aio/aio_task_queue.h
```diff
@@ -0,0 +1,170 @@
...
+        if (_notify) {
+            _notify->wait();
+        }
+    }
+    void complete()
+    {
+        _processed_bytes = 0;
+        auto aio_ctx = _tsk->get_aio_context();
+        _err = ERR_UNKNOWN;
+        if (aio_ctx->type == AIO_Read) {
```
levy5307, line 58: use switch is better here

src/aio/aio_task_queue.h
```diff
@@ -0,0 +1,170 @@
...
+        default:
+            dassert_f(false,
+                      "task priority must be set "
+                      "TASK_PRIORITY_HIGH/TASK_PRIORITY_COMMON/TASK_PRIORITY_LOW");
+            break;
+        }
+    };
+
+private:
+    std::vector<std::shared_ptr<io_event_loop_t>> _high_pri_workers;
```
levy5307, line 166: `map<dsn_task_priority_t, std::vector<std::shared_ptr<io_event_loop_t>>>`
so the func `enqueue` can be more simplify

hycdong (approved):

src/aio/native_linux_aio_provider.cpp
```diff
@@ -86,144 +63,70 @@ error_code native_linux_aio_provider::flush(dsn_handle_t fh)
...
-            io_prep_pwritev(
-                &aio->cb, static_cast<int>((ssize_t)aio->file), iov, iovcnt, aio->file_offset);
-        }
-        break;
-    default:
-        derror("unknown aio type %u", static_cast<int>(aio->type));
+    aio_context *aio_ctx = aio_tsk->get_aio_context();
+    error_code err = ERR_UNKNOWN;
+    uint32_t processed_bytes = 0;
+    if (aio_ctx->type == AIO_Read) {
```
levy5307, line 110: use switch is better

src/aio/native_linux_aio_provider.cpp
```diff
@@ -86,144 +63,70 @@ error_code native_linux_aio_provider::flush(dsn_handle_t fh)
...
-        } else {
-            aio->evt->wait();
-            delete aio->evt;
-            aio->evt = nullptr;
-            if (pbytes != nullptr) {
-                *pbytes = aio->bytes;
-            }
-            return aio->err;
-        }
+        std::unique_ptr<utils::notify_event> notify = std::make_unique<utils::notify_event>();
```
levy5307, line 125: `
utils::notify_event notify;
notify.notify();
`

src/aio/aio_provider.h
```diff
@@ -61,6 +61,8 @@ class aio_provider
 
     virtual error_code close(dsn_handle_t fh) = 0;
     virtual error_code flush(dsn_handle_t fh) = 0;
+    virtual error_code write(aio_context *aio_ctx, uint32_t *processed_bytes) = 0;
```
levy5307, line 64: `/*out*/ uint32_t *processed_bytes`

neverchanje: Yes, we can do better by:
```suggestion
    virtual error_code write(const aio_context &aio_ctx, /*out*/ uint32_t *processed_bytes) = 0;
```

src/nfs/nfs_client_impl.cpp
```diff
@@ -212,7 +212,7 @@ void nfs_client_impl::end_get_file_size(::dsn::error_code err,
             _copy_requests_low.push(std::move(copy_requests));
     }
 
-    continue_copy();
+    tasking::enqueue(LPC_NFS_COPY_FILE, nullptr, [this]() { continue_copy(); }, 0);
```
neverchanje, line 215: I'm uncertain if this is a good fix. I think `continue_copy` should not block at first.
As you pointed out, the blocking point is `folly::TokenBucket::consumeWithBorrowAndWait`.
We should use a non-blocking alternative `consumeWithBorrowNonBlocking` to prevent unnecessary block.
But anyway, we can leave it later.

Shuo-Jia: OK!

neverchanje (dismissed):

src/aio/disk_engine.h
```diff
@@ -28,6 +28,7 @@
 
 #include "aio_provider.h"
 
+#include <dsn/tool_api.h>
```
levy5307, line 31: Why did you add `#include <dsn/tool_api.h>`?

Shuo-Jia: https://github.com/XiaoMi/rdsn/blob/5cf7373e27c0730a8ae7cc11b9780d23c3dc774d/src/aio/disk_engine.cpp#L40-L42 need it, otherwise compile error

levy5307: so you can add it in `disk_engine.cpp`, but not `disk_engine.h`

Shuo-Jia: Yes

levy5307 (dismissed):

neverchanje (approved):

levy5307 (approved):

PR #635, opened 2020-09-30
fix: fix fds_service_test

zhangyifan27:
- fix tests, now test could pass with correct fds settings.
- do not check meta before read/download, because we would call `get_file_meta()` in `fds_file_object::get_content_in_batches()`.
- when download failed, we should remove existing local file.
- other code cleanup.

### DIFF
--- a/src/block_service/fds/fds_service.cpp
+++ b/src/block_service/fds/fds_service.cpp
@@ -398,20 +398,6 @@ fds_file_object::fds_file_object(fds_service *s,
 {
 }
 
-fds_file_object::fds_file_object(fds_service *s,
-                                 const std::string &name,
-                                 const std::string &fds_path,
-                                 const std::string &md5,
-                                 uint64_t size)
-    : block_file(name),
-      _service(s),
-      _fds_path(fds_path),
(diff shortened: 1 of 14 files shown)


acelyc111 (approved):

src/block_service/fds/fds_service.cpp
```diff
@@ -722,9 +692,16 @@ dsn::task_ptr fds_file_object::download(const download_request &req,
         resp.err =
             get_content_in_batches(req.remote_pos, req.remote_length, *handle, transfered_size);
         resp.downloaded_size = 0;
-        if (handle->tellp() != -1)
+        if (resp.err == ERR_OK && handle->tellp() != -1) {
             resp.downloaded_size = handle->tellp();
+        }
         handle->close();
+        if (resp.err != ERR_OK && dsn::utils::filesystem::file_exists(req.output_local_name)) {
+            derror("fail to download file(%s) from fds, remove localfile(%s)",
```
levy5307, line 700: derror_f

acelyc111 (approved):

levy5307 (approved):

PR #638, opened 2020-10-10
refactor(split): refactor meta start partition split

hycdong:
This pull request is a refactor of meta start partition split.


src/runtime/security/security_types.cpp
```diff
@@ -275,6 +275,5 @@ void negotiation_response::printTo(std::ostream &out) const
         << "msg=" << to_string(msg);
     out << ")";
 }
-
-} // namespace security
-} // namespace dsn
+}
```
acelyc111, line 278: namespace should be kept

src/runtime/security/security_types.h
```diff
@@ -147,8 +147,7 @@ inline std::ostream &operator<<(std::ostream &out, const negotiation_response &o
     obj.printTo(out);
     return out;
 }
-
-} // namespace security
-} // namespace dsn
+}
```
acelyc111, line 150: namespace should be kept

src/runtime/security/security_types.cpp
hycdong: src/runtime/security/security_types.h and src/runtime/security/security_types.cpp are auto-generated by thirft, it will be removed each time when we update thift sturctures, the namespace maybe manually added before. Should we add it manually when we change thrift sturctures?

acelyc111: That's OK, not needed.

acelyc111 (approved):

levy5307 (approved):

PR #641, opened 2020-10-14
feat(split): add split_status

hycdong:
This pull request adds two sturctures to show split status.
### split_status (in replication.thirft)
Define five split_status for partition (not_split, splitting, pausing, paused, canceling), both replica and meta will use it.
- For replica, we used to use `_child_gpid` to distinguish whether a replica is executing splitting, if `_child_gpid.get_app_id() > 0`, the replica is splitting, otherwise not. It can only show two split status. Now, we use `_split_status` to describe partition split status.
- For meta, see split_state
### split_state (in meta_data.h)
Define sturct `split_state` to show app split progress, this sturcture include splitting partition count and each split_status for those splitting partitions. If `splitting_count = 0`, it means current app is not splitting or has finished splitting. 

This sturcture is helpful to query app split progress and easier to understand. In previous implementation, when an app is executing splitting, tha ballot of not-ready child partition is `invalid_ballot`( < 0), so we should traverse `app->partitions` to check if any partition whose ballot is invalid. Now, we just check the value of `splitting_count`.

This pr also update related unit tests.

### DIFF
--- a/include/dsn/dist/replication/replication_enums.h
+++ b/include/dsn/dist/replication/replication_enums.h
@@ -21,6 +21,7 @@ ENUM_REG(replication::partition_status::PS_ERROR)
 ENUM_REG(replication::partition_status::PS_PRIMARY)
 ENUM_REG(replication::partition_status::PS_SECONDARY)
 ENUM_REG(replication::partition_status::PS_POTENTIAL_SECONDARY)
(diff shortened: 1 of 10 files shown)


acelyc111 (approved):

Shuo-Jia (approved):

Shuo-Jia:
The file seem not be named `xxxmanager`, since it does not any `manager` work, may you can named better name later

PR #644, opened 2020-10-20
refactor: reduce netprovider unit test executing time

hycdong:
As [pegasus issue 567](https://github.com/apache/incubator-pegasus/issues/567) shows, there are some unit tests are time-cosuming, ` tools_common.asio_network_provider_connection_threshold` is one of them, it will cost 70017ms. This pull request is a refactor of this unit test to reduce its executing time, it will cost 2077ms to run this unit test.

### DIFF
--- a/src/runtime/test/netprovider.cpp
+++ b/src/runtime/test/netprovider.cpp
@@ -107,7 +107,7 @@ void rpc_server_response(dsn::message_ex *request)
 void wait_response()
 {
     while (wait_flag == 0)
-        std::this_thread::sleep_for(std::chrono::seconds(1));
+        std::this_thread::sleep_for(std::chrono::milliseconds(5));
 }
 
 void rpc_client_session_send(rpc_session_ptr client_session, bool reject = false)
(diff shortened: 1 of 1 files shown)


acelyc111 (approved):

src/runtime/test/netprovider.cpp
```diff
@@ -272,9 +272,12 @@ TEST(tools_common, asio_network_provider_connection_threshold)
...
     start_result = asio_network->start(RPC_CHANNEL_TCP, TEST_PORT, false);
     ASSERT_TRUE(start_result == ERR_OK);
-    asio_network->change_test_cfg_conn_threshold_per_ip(10);
 
-    for (int count = 0; count < 20; count++) {
+    auto CONN_THRESHOLD = 3;
+    asio_network->change_test_cfg_conn_threshold_per_ip(CONN_THRESHOLD);
+
+    // not exceed threshold
+    for (int count = 0; count < CONN_THRESHOLD + 2; count++) {
```
neverchanje, line 280: Why doesn't it exceed threshold when sessions count == CONN_THRESHOLD +2?

hycdong: Acutually, the count doesn't matter in not-exceed case, it will first create a client to create a new session, then close the session, so it will not exceed connection threshold, the count here only means the loop count.

neverchanje (approved):

PR #652, opened 2020-10-28
fix(security): fix bug in negotiation_service::on_negotiation_request when rpc_session is closed

levy5307:
1. Previously, if one rpc session is disconnected, we should remove the corresponding `negotiation` in `negotiation_service`.

In multi-thread environment, if thread A receive a negotiation message, and pass this message to `negotiation_service::on_negotiation_request`, which will find the correspoing `negotiation` to deal with this negotiation message. 

And before this message is passed to `negotiation_service::on_negotiation_reques`, thread B receive a message which is illegal to process(for example: get operation),  which will produce the rpc_session disconnected. In the meanwhile, the negotiation is removed. 

So in `negotiation_service::on_negotiation_request` of thead A, the corresponding `negotiation` will be `nullptr`, this will produce a coredump.

Here is the process progress:
```
thread A:     -->receive nego msg--------------------------------------------------------------->get negotiation----->......---->

thread B: --------> receive get mesage-->close rpc_session-->remove negotiation----->......----->
```

Same situation with client_negotiation processing response.

2. The ownership of the `negotiation` instance is held by `rpc_session`, because I refactor it before. security is on the upper level of network now. So `negotiation` is better to have the ownership of `rpc_session`.

3. change the name from `negotiaiton_service` to `negotiation_manager`


src/runtime/security/negotiation_service.cpp
```diff
@@ -67,9 +67,10 @@ void negotiation_service::on_negotiation_request(negotiation_rpc rpc)
             static_cast<server_negotiation *>(_negotiations[rpc.dsn_request()->io_session].get());
     }
 
-    dassert(srv_negotiation != nullptr,
-            "negotiation is null for msg: {}",
-            rpc.dsn_request()->rpc_code().to_string());
+    if (nullptr == srv_negotiation) {
+        ddebug("negotiation is null for msg: {}", rpc.dsn_request()->rpc_code().to_string());
```
neverchanje, line 71:
```suggestion
        derror("negotiation is null for msg: {}", rpc.dsn_request()->rpc_code().to_string());
```

This case appears when the client is incorrectly sending requests before negotiation completes. In such a case, I think logging in error-level could help us find the cause.

neverchanje: Could we just reply the unnegotiated messages using ERR_UNNEGOTIATED, without shutting down the session?

levy5307: I think shut down the session is the best practise. Because if we don't close it, the client can also send message to server. It is waste of network bandwidth

src/runtime/security/negotiation_service.cpp
```diff
@@ -67,12 +69,32 @@ void negotiation_service::on_negotiation_request(negotiation_rpc rpc)
...
-            "negotiation is null for msg: {}",
-            rpc.dsn_request()->rpc_code().to_string());
+    if (nullptr == srv_negotiation) {
+        derror_f("negotiation is null for msg: {}", rpc.dsn_request()->rpc_code().to_string());
+        return;
+    }
     srv_negotiation->handle_request(rpc);
 }
 
+void negotiation_service::on_negotiation_response(error_code err, negotiation_rpc rpc)
```
neverchanje, line 79: The name `negotiation_service` is ambiguous. It conflicts with `server_negotiation` that easily gets people confused about the difference. Apart from adding an explanation in the comment, I recommend you rename the class, to "negotiation_manager" or so.

levy5307: done.

src/runtime/security/negotiation_manager.cpp
```diff
@@ -67,13 +69,33 @@ void negotiation_service::on_negotiation_request(negotiation_rpc rpc)
...
+
+    client_negotiation *cli_negotiation = nullptr;
+    {
+        utils::auto_read_lock l(_lock);
+        cli_negotiation =
+            static_cast<client_negotiation *>(_negotiations[rpc.dsn_request()->io_session].get());
+    }
+
+    if (nullptr == cli_negotiation) {
+        derror_f("negotiation is null for msg: {}", rpc.dsn_request()->rpc_code().to_string());
```
neverchanje, line 92: What do we gonna do for such a message? Isn't there any error code? Any response? Does that mean the client will keep waiting but the server just ignores it?

levy5307: in this case, the session is closed already. so we don't need send a response here.

neverchanje: I can't see why it's closed? If it's closed, why it wasn't removed from `_negotiations`?

levy5307: If `client_negotiation` sends a message to `server_negotiation`,  and the `server_negotiation` find something is wrong when it receives this message, it will close the connection. 
On the client side, it will remove the negotiation from `_negotiation`.
And then the message which is send by client_negotiation  is timeout. It will calls the corresponding callback, which is `on_negotiation_response`. And in this function, it will not find the negotiation, because it is already removed. So we should add a judgement here.

neverchanje:
```
    {
        cli_negotiation =
            static_cast<client_negotiation *>(_negotiations[rpc.dsn_request()->io_session].get());
    }
    if (nullptr == cli_negotiation) {
        derror_f("negotiation is null for msg: {}", rpc.dsn_request()->rpc_code().to_string());
```

This is not proper usage of std::map. Actually `_negotiations[rpc.dsn_request()->io_session]` is not a read-only operation. It WRITES to the map. I think you misunderstood this, a very common mistake that would be made for new c++er. This code will incur thread-conflict as the section was wrapped by read-lock only, rather than write-lock. 

If you want to check if the `client_negotiation` was removed, use `std::map::find`.

```
    {
        auto it = _negotiations.find(rpc.dsn_request()->io_session);
        if(it == _negotiations.end()) {
            return; // ignore
        }
        cli_negotiation = static_cast<client_negotiation *>(it->second.get())
    }
    cli_negotiation->handle_response(err, std::move(rpc.response()));
```

src/runtime/security/negotiation_manager.cpp
```diff
@@ -38,80 +40,100 @@ inline bool in_white_list(task_code code)
...
-    {
-        utils::auto_read_lock l(_lock);
-        srv_negotiation =
-            static_cast<server_negotiation *>(_negotiations[rpc.dsn_request()->io_session].get());
+    std::shared_ptr<negotiation> nego = get_negotiation(rpc);
+    if (nullptr != nego) {
+        server_negotiation *srv_negotiation = static_cast<server_negotiation *>(nego.get());
+        srv_negotiation->handle_request(rpc);
     }
+}
```
hycdong, line 70: Is it necessary to add error log even error handling for nego is nullptr? The same in on_negotiation_response function.

src/runtime/security/negotiation_manager.cpp
```diff
@@ -38,80 +40,100 @@ inline bool in_white_list(task_code code)
...
-    dassert(srv_negotiation != nullptr,
-            "negotiation is null for msg: {}",
-            rpc.dsn_request()->rpc_code().to_string());
-    srv_negotiation->handle_request(rpc);
+void negotiation_manager::on_negotiation_response(error_code err, negotiation_rpc rpc)
+{
+    dassert(rpc.dsn_request()->io_session->is_client(),
+            "only client session receives negotiation response");
+
+    std::shared_ptr<negotiation> nego = get_negotiation(rpc);
```
neverchanje, line 77: Why use shared_ptr here... The memory of `nego` is still owned by `_negotiations`. You are transferring the memory to std::shared_ptr... And when this function completes, `nego` will free the internal memory because no one reference it.

I don't understand why such simple mistake happens but tests still pass.

neverchanje: Add test something like the follows:

```
TEST(negotiation_manager_test, client_negotiation) {
  negotiation_rpc fake_rpc;
  ... // mock data
  on_negotiation_response(ERR_OK, fake);

  // ensure not null, ensure memory valid.
  ASSERT_EQ(get_negotiation(rpc), xxx);
}
```

src/runtime/security/negotiation_manager.cpp
levy5307: I think it's ok to add a log here. And there won't be too much log here, because it doesn't happen very often

.

src/runtime/security/negotiation_manager.cpp
levy5307: I think we must use shared_ptr here. because if we use unique_ptr, the negotiation maybe destroyed by `negotiation_manager::on_rpc_disconnected`.  so in L77 we must declare a shared_ptr to add it's ref count

neverchanje (approved):

hycdong (approved):

PR #645, opened 2020-10-21
feat(split): add update_child_group_partition_count

hycdong:
## Simple partition split process
1. meta receives client partition split request, and change partition count #286
2. replica notices partition count changed during on_config_sync
3. parent partition create child partition #291
4. parent prepare states for child to learn #299
5. child partition async learn states from parent #309 #319
6. child notify parent catch up #390
7. **update child group partition count**
8. register child partition #391
9. update parent group partition count

More partition split discussion in issue #69 and partition split design doc
This pr solves the part of seventh step of partition split, which is bold in process description.

## What this pr solve
When all children have caught up its parent's state, parent will send rpc to child to let them update partition count, child will update its partition count in memory and update `.app_info` file on disk. If all children update partition_count succeed, it will send register child rpc to meta server. Besides, this pr also adds related unit tests.

PR #391, opened 2020-02-11
feat(split): register child partition

hycdong:
## Simple partition split process
1. meta receives client partition split request, and change partition count #286
2. replica notices partition count changed during on_config_sync
3. parent partition create child partition #291
4. parent prepare states for child to learn #299 
5. child partition async learn states from parent #309 #319 
6. child notify parent catch up #390 
7. update child group partition count
8. **register child partition**
9. parent update group partition count and recover read and write

More partition split discussion in issue #69 and [partition split design doc](https://pegasus-kv.github.io/2020/02/06/partition-split-design.html)
This pr solves the part of fifth step of partition split, which is bold in process description.

## What this pr solved
- `replica::register_child_on_meta`: primary parent register child partition on meta_server and parent will reject read write client.
- `meta_split_service::register_child_on_meta`: meta server update child partition_configuration structure on remote storage and local.
- `child_partition_active`: when child registered, child of primary parent will be the primary of the child replica group, its status will turn from `PS_PARTITION_SPLIT` to `PS_PRIMARY`, child of secondary will update its status by group_check.

PR #390, opened 2020-02-10
feat(split): child notify parent catch up

hycdong:
## Simple partition split process
1. meta receives client partition split request, and change partition count #286
2. replica notices partition count changed during on_config_sync
3. parent partition create child partition #291
4. parent prepare states for child to learn #299 
5. child partition async learn states from parent #309 #319 
6. **child notify parent catch up**
7. meta server register child partitions
8. child partition active, and parent recover read and write

More partition split discussion in issue #69 and [partition split design doc](https://pegasus-kv.github.io/2020/02/06/partition-split-design.html)
This pr solves the part of fifth step of partition split, which is bold in process description.

## What this pr solved
- `child_notify_catch_up`: child partition sent `notify_catch_up_request` to primary parent
- `parent_handle_child_catch_up`: primary parent handle child's catch_up_request, if all child partitions caught up, parent will copy mutation to child **synchronously**.
- `sync_point`: sync_point is the first decree after parent send write request to child synchronously. When `sync_point` committed, parent consider child has learned all data, and primary parent will update partition_count of child replica group (not in this pr, but in #392)

PR #286, opened 2019-07-30
split: meta start partition split

hycdong:
### Simple partition split process
1. **meta receives client partition split request, and change partition count**
2. replica notices partition count changed during on_config_sync
3. parent partitions create child partitions, child partitions async learn mutations and private logs from parents
4. all child partitions in group finish learn process, parent will stop read and write
5. meta server register child partitions
6. child partition active, and parent recover read and write

More partition split discussion in  issue #69 
This pr solves the first step of partition split, which is bold in process description.

### Main code logic
When meta server receives a partition split request:
- meta will firstly validate parameters, including app status, partition count etc
- if validation passed, meta will update new partition_count on zk
- then update new partition_count on meta server locally
- then create child partitions' partition_configuration structure locally, and we assume child partition init balliot = invalid_ballot(-1), which means the partition is currently invalid

### Other explanations
- add init_partition_count in app_info, it will be used when meta server failed during partition split, I will explain it in another new pull request
- add unit tests for function ''app_partition_split"

PR #653, opened 2020-10-28
feat(split): add splitting_replicas while on_config_sync

hycdong:
## Simple partition split process
1. meta receives client partition split request, and change partition count #286
2. **replica notices partition count changed during on_config_sync**
3. parent partition create child partition #291
4. parent prepare states for child to learn #299
5. child partition async learn states from parent #309 #319
6. child notify parent catch up #390
7. update child group partition count #645 
8. register child partition #391
9. update parent group partition count #654

More partition split discussion in issue #69 and partition split design doc
This pr solves the part of second step of partition split, which is bold in process description.

## What this pr solve
Replica server will send on_config_sync rpc to meta server periodically to get the newest configuration from meta server, we use it to transfer the split status from meta to replica.
- for meta server, if any partition on this node is splitting, it will be added into splitting_replicas in response ( function `on_config_sync` in server_state.cpp )
- for replica server,  it will parse split_status from response ( function `on_node_query_reply` in replica_stub.cpp ), then do possible split actions ( function `on_config_sync` in replica_config.cpp and function `control_split` in replica_split_manager.cpp ). This pr only add **primary start split**, the more split action will be added in further pull request.

This pr also adds related unit tests.


src/replica/replica_stub.cpp
```diff
@@ -1331,11 +1334,20 @@ void replica_stub::on_node_query_reply(error_code err,
         }
 
         for (auto it = resp.partitions.begin(); it != resp.partitions.end(); ++it) {
+            auto meta_split_status = split_status::NOT_SPLIT;
```
levy5307, line 1337: maybe you can calculate `meta_split_status` in meta server. So we don't need to calculate it in each replica server

hycdong: If an app is not splitting, `splitting_replicas` in `configuration_query_by_node_response` won't be set. The config_sync rpc will happen every 30 seconds for all replica servers, and partition split is not a frequent action, for most cases it will reduce the network cost.

src/meta/test/meta_split_service_test.cpp
```diff
@@ -202,5 +211,52 @@ TEST_F(meta_split_service_test, register_child_test)
...
+    // - partition is not splitting
+    // - TODO(heyuchen): partition split is paused({false, true, 1})
+    struct config_sync_test
+    {
+        bool mock_child_registered;
+        bool mock_parent_paused;
+        int32_t expected_count;
+    } tests[] = {{false, false, 1}, {true, false, 0}};
+
+    for (auto test : tests) {
```
levy5307, line 247: const auto &test

src/replica/split/test/replica_split_test.cpp
```diff
@@ -713,5 +723,38 @@ TEST_F(replica_split_test, register_child_reply_test)
...
+        int32_t old_partition_version;
+        split_status::type old_split_status;
+        split_status::type expected_split_status;
+    } tests[]{{true,
+               split_status::SPLITTING,
+               OLD_PARTITION_COUNT - 1,
+               split_status::NOT_SPLIT,
+               split_status::NOT_SPLIT}};
+
+    for (auto test : tests) {
```
levy5307, line 750: const auto &test

src/replica/split/replica_split_manager.cpp
```diff
@@ -1103,6 +1103,48 @@ void replica_split_manager::child_handle_async_learn_error() // on child partiti
     _replica->_split_states.async_learn_task = nullptr;
 }
 
+// ThreadPool: THREAD_POOL_REPLICATION
+void replica_split_manager::trigger_primary_parent_split(
+    const int32_t meta_partition_count,
+    split_status::type meta_split_status) // on primary parent partition
```
levy5307, line 1109: const

acelyc111 (approved):

src/replica/replica_stub.cpp
```diff
@@ -1393,11 +1409,12 @@ void replica_stub::set_replica_state_subscriber_for_test(replica_state_subscribe
...
 // ThreadPool: THREAD_POOL_REPLICATION
 void replica_stub::on_node_query_reply_scatter(replica_stub_ptr this_,
-                                               const configuration_update_request &req)
+                                               const configuration_update_request &req,
+                                               split_status::type meta_split_status)
 {
     replica_ptr replica = get_replica(req.config.pid);
     if (replica != nullptr) {
-        replica->on_config_sync(req.info, req.config);
+        replica->on_config_sync(req.info, req.config, meta_split_status);
```
neverchanje, line 1417: It's ugly to particularly add an argument for one requirement. It makes the function responsibility obscure. I'd prefer adding `meta_split_status` to `configuration_update_request`, even it's ugly too.

Think about it, dozens of parameters in a function, if we keep doing that.

hycdong: Done

acelyc111 (approved):

neverchanje (approved):

PR #655, opened 2020-11-02
feat(security): implement meta server access controller

levy5307:
```diff
+ [security]
+ enable_acl = false
+ super_users = 
+ meta_acl_rpc_allow_list = 
```


src/meta/meta_service.cpp
```diff
@@ -76,6 +77,8 @@ meta_service::meta_service()
         "replica server disconnect count in the recent period");
     _unalive_nodes_count.init_app_counter(
         "eon.meta_service", "unalive_nodes", COUNTER_TYPE_NUMBER, "current count of unalive nodes");
+
+    _access_controller = security::create_access_controller(true);
```
acelyc111, line 81: The parameter is always true?

src/runtime/security/sasl_wrapper.cpp
```diff
@@ -39,6 +40,23 @@ sasl_wrapper::~sasl_wrapper()
...
+{
+    FAIL_POINT_INJECT_F("sasl_wrapper_retrive_username", [](dsn::string_view str) {
+        error_code err = error_code::try_get(str.data(), ERR_UNKNOWN);
+        return error_s::make(err);
+    });
+
+    char *username = nullptr;
+    error_s err_s = wrap_error(sasl_getprop(_conn, SASL_USERNAME, (const void **)&username));
+    if (err_s.is_ok()) {
+        output = username;
```
acelyc111, line 53: Could you please add some comment?

src/runtime/security/access_controller.h
```diff
@@ -0,0 +1,54 @@
...
+{
+public:
+    access_controller();
+    virtual ~access_controller() = 0;
+
+    /**
+     * reset the access controller
+     *    acls - the new acls to reset
+     **/
+    virtual void reset(const std::string &acls){};
```
acelyc111, line 37:
```suggestion
    virtual void reset(const std::string &acls) {}
```

src/runtime/security/access_controller.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include "meta_access_controller.h"
+
+namespace dsn {
+namespace security {
+DSN_DEFINE_bool("security", enable_acl, false, "whether enable access controller or not");
+DSN_DEFINE_string("security", super_users, "", "super user for access controller");
+
+access_controller::access_controller() { utils::split_args(FLAGS_super_users, _super_users, ','); }
+
+access_controller::~access_controller() {}
```
acelyc111, line 32: You defined it as a pure virtual function in .h file

src/runtime/security/access_controller.h
```diff
@@ -0,0 +1,54 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#pragma once
+
+#include <memory>
```
acelyc111, line 20: Should be removed?

src/runtime/security/meta_access_controller.h
```diff
@@ -0,0 +1,39 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#pragma once
+
+#include "access_controller.h"
+#include <unordered_set>
```
acelyc111, line 21:
```suggestion
#include <unordered_set>

#include "access_controller.h"
```

src/runtime/security/meta_access_controller.h
```diff
@@ -0,0 +1,39 @@
...
+
+namespace dsn {
+class message_ex;
+namespace security {
+
+class meta_access_controller : public access_controller
+{
+public:
+    meta_access_controller();
+    bool allowed(message_ex *msg);
```
acelyc111, line 31:
```suggestion
    bool allowed(message_ex *msg) override;
```

src/runtime/security/meta_access_controller.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include <dsn/tool-api/rpc_message.h>
+#include <dsn/utility/flags.h>
+#include <dsn/tool-api/network.h>
+
+namespace dsn {
+namespace security {
+DSN_DEFINE_string("security",
+                  rpc_code_white_list,
+                  "",
+                  "white list of rpc codes for meta_access_controller");
```
acelyc111, line 29: Why it's a user specified parameter?

src/runtime/security/meta_access_controller.cpp
```diff
@@ -0,0 +1,57 @@
...
+        register_white_list("RPC_CM_QUERY_PARTITION_CONFIG_BY_INDEX");
+    } else {
+        utils::split_args(FLAGS_rpc_code_white_list, _white_list, ',');
+    }
+}
+
+bool meta_access_controller::allowed(message_ex *msg)
+{
+    if (pre_check(msg->io_session->get_client_username()) ||
+        _white_list.find(msg->rpc_code().to_string()) != _white_list.end()) {
```
acelyc111, line 46: int is more efficient than string, better to use int directly, this function will be called frequently in read/write path.

src/runtime/security/meta_access_controller.h
```diff
@@ -0,0 +1,39 @@
...
+
+private:
+    void register_white_list(const std::string &rpc_code);
+
+    std::unordered_set<std::string> _white_list;
```
acelyc111, line 36: _allowed_rpc_code_list?

src/meta/meta_service.cpp
levy5307: In this pr, it is always true. but it will be false when we want to create a replica_access_controller, which will be added in next pr.

src/runtime/security/access_controller.h
levy5307: clang-format changes it like this way, which without a ' '.

src/runtime/security/access_controller.cpp
levy5307: We can't put it in .h file. because the factory function `create_access_controller` returns a `unique_ptr<access_controller>`, and `unique_ptr` needs to know destor of `access_controller`, or it will produce compile error. Ref: https://www.zhihu.com/question/53878904

src/runtime/security/access_controller.h
levy5307: no, because there is a factory function which uses unique_ptr:
```
std::unique_ptr<access_controller> create_access_controller(bool is_meta);

```

src/runtime/security/meta_access_controller.cpp
levy5307: I make `rpc_code_white_list` configurable. so if we should only change the config file, but not the cpp file if we want to add a rpc code to this white list.

src/runtime/security/meta_access_controller.cpp
```diff
@@ -0,0 +1,68 @@
...
+namespace dsn {
+namespace security {
+DSN_DEFINE_string("security",
+                  rpc_code_allowed_list,
+                  "",
+                  "allowed list of rpc codes for meta_access_controller");
+
+meta_access_controller::meta_access_controller()
+{
+    if (strlen(FLAGS_rpc_code_allowed_list) == 0) {
```
neverchanje, line 34:
```suggestion
    // MetaServer serves the allow-list RPC from all users. RPCs unincluded are accessible to only superusers.
    if (strlen(FLAGS_rpc_code_allowed_list) == 0) {
```

src/runtime/security/meta_access_controller.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+#include <dsn/tool-api/rpc_message.h>
+#include <dsn/utility/flags.h>
+#include <dsn/tool-api/network.h>
+#include <dsn/dist/fmt_logging.h>
+
+namespace dsn {
+namespace security {
+DSN_DEFINE_string("security",
+                  rpc_code_allowed_list,
```
neverchanje, line 28:
```suggestion
                  meta_acl_rpc_allow_list,
```

I can't read from the config name what it's used for. Rename it with the keyword "meta" and "acl".

include/dsn/tool-api/network.h
```diff
@@ -328,6 +331,10 @@ class rpc_session : public ref_counter
     rpc_client_matcher *_matcher;
 
     std::atomic_int _delay_server_receive_ms;
+
+    // _client_username is only valid if it is a server rpc_session.
+    // it represents the name of the corresponding client
+    std::string _client_username;
```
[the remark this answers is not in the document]
levy5307: I think it's ok to add a `client_username` property for rpc_session. It's a stuff of the session itself, but not provided for security. security is just using it.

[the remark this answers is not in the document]
neverchanje: No authorization/negotiation then no username. The RPC module should be pure network RPC implementation that can be used in every place whether or not the auth enabled. We must insist open-close principle that prevents such modification to add responsibility to a class.

I will take some time thinking about how to dispose of this variable, if you got no idea then.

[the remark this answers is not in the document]
levy5307: OK, but I don't suggest add user_name to negotiation, because in this way we should get the corresponding negotiation, which will acquire a lock. It will produce low efficiency

acelyc111 (approved):

src/runtime/security/meta_access_controller.cpp
```diff
@@ -0,0 +1,70 @@
...
+    }
+    return false;
+}
+
+void meta_access_controller::register_allowed_list(const std::string &rpc_code)
+{
+    auto code = task_code::try_get(rpc_code, TASK_CODE_INVALID);
+    if (code == TASK_CODE_INVALID) {
+        dassert_f(
+            false, "invalid task code({}) in rpc_code_white_list of security section", rpc_code);
```
Shuo-Jia, line 64:
```c++
dassert_f(code != TASK_CODE_INVALID, "invalid task code({}) in rpc_code_white_list of security section", rpc_code)
```

src/meta/meta_service.cpp
```diff
@@ -120,6 +123,12 @@ int meta_service::check_leader(TRpcHolder rpc, rpc_address *forward_address)
 template <typename TRpcHolder>
 bool meta_service::check_status(TRpcHolder rpc, rpc_address *forward_address)
 {
+    if (!_access_controller->allowed(rpc.dsn_request())) {
+        rpc.response().err = ERR_ACL_DENY;
+        ddebug("reject request with ERR_ACL_DENY");
```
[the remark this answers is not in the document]
levy5307: I think use ddebug is ok here. because there is no format work to do here.

neverchanje (approved):

PR #658, opened 2020-11-03
chore: migrate ci from travis to github actions

neverchanje:
This PR configures the github action jobs of when a pull-request created. I've written a wiki here: https://cwiki.apache.org/confluence/display/PEGASUS/Github+Actions.

In addition, since the former formatting script `./scripts/linux/run-clang-format.sh` relies on git to compare if codes changed, which is unavailable on github actions, I introduce a new file `run-clang-format.py` copied directly from https://github.com/Sarcasm/run-clang-format.

As travis-ci is no longer used, `.travis.yml` is removed now.

Test `load_from_private_log_test.restart_duplication2` is removed because the case fails on CI. And I found it impossible to happen. Because learner always replaces the original log dir if it learns from primary's private log.


acelyc111:
good job!

zhangyifan27:
I think we could run build and tests for every pull request, instead of adding a  "ready-for-testing" manually for the PR needed to test.

neverchanje:
> I think we could run build and tests for every pull request, instead of adding a "ready-for-testing" label manually for the PR needed to test.

It's for the security of our self-hosted runner. Without such restriction, anyone can submit a pull-request that will possibly attack the server.

levy5307 (dismissed):

acelyc111:
> > I think we could run build and tests for every pull request, instead of adding a "ready-for-testing" label manually for the PR needed to test.
> 
> It's for the security of our self-hosted runner. Without such restriction, anyone can submit a pull-request that will possibly attack the server. And of course, it saves the resource as well.

I agree with @zhangyifan27 , and seems we can limit parallel jobs, https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobs

neverchanje:
> I agree with @zhangyifan27 , and seems we can limit parallel jobs, https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobs

@acelyc111 
We don't have parallel jobs. Jobs within the PR workflow run one by one. I guess you imply that many CI workflows could run in one runner. AFAIK, if our runner is occupied, the workflow will fallback to Github runner.

neverchanje:
@acelyc111 OK, now I removed ready-for-testing requirement for PR. Please take a look.

acelyc111 (approved):

acelyc111:
> @acelyc111 OK, now I removed ready-for-testing requirement for PR. Please take a look.

LGTM

levy5307 (approved):

PR #661, opened 2020-11-06
fix(backup): use existing block_service_manager when add backup policy

zhangyifan27:
The way we determine whether a backup provider is valid is to construct a `block_service_manager` and destroy it, which is not reasonable and may cause problems when use hdfs as backup provider.

### DIFF
--- a/src/block_service/block_service_manager.cpp
+++ b/src/block_service/block_service_manager.cpp
@@ -36,8 +36,8 @@ block_service_manager::block_service_manager()
 block_service_manager::~block_service_manager()
 {
     zauto_write_lock l(_fs_lock);
-    ddebug("close block service manager.");
     _fs_map.clear();
+    ddebug("close block service manager.");
 }
 
 block_filesystem *block_service_manager::get_block_filesystem(const std::string &provider)
--- a/src/block_service/block_service_manager.cpp
+++ b/src/block_service/block_service_manager.cpp
@@ -33,54 +33,49 @@ block_service_manager::block_service_manager()
 {
 }
 
+block_service_manager::~block_service_manager()
+{
+    zauto_write_lock l(_fs_lock);
+    ddebug("close block service manager.");
+    _fs_map.clear();
+}
+
 block_filesystem *block_service_manager::get_block_filesystem(const std::string &provider)
 {
-    {
-        zauto_read_lock l(_fs_lock);
-        auto iter = _fs_map.find(provider);
-        if (iter != _fs_map.end())
-            return iter->second.get();
+    zauto_write_lock l(_fs_lock);
+    auto iter = _fs_map.find(provider);
+    if (iter != _fs_map.end()) {
+        return iter->second.get();
     }
 
-    {
-        zauto_write_lock l(_fs_lock);
-        auto iter = _fs_map.find(provider);
-        if (iter != _fs_map.end())
-            return iter->second.get();
+    const char *provider_type = dsn_config_get_value_string(
+        (std::string("block_service.") + provider).c_str(), "type", "", "block service type");
 
-        const char *provider_type = dsn_config_get_value_string(
-            (std::string("block_service.") + provider).c_str(), "type", "", "block service type");
+    block_filesystem *fs =
+        utils::factory_store<block_filesystem>::create(provider_type, PROVIDER_TYPE_MAIN);
+    if (fs == nullptr) {
+        derror("acquire block filesystem failed, provider = %s, provider_type = %s",
+               provider.c_str(),
+               provider_type);
+        return nullptr;
+    }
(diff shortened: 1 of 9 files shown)


src/block_service/block_service_manager.cpp
```diff
@@ -33,54 +33,49 @@ block_service_manager::block_service_manager()
...
+}
+
 block_filesystem *block_service_manager::get_block_filesystem(const std::string &provider)
 {
-    {
-        zauto_read_lock l(_fs_lock);
-        auto iter = _fs_map.find(provider);
-        if (iter != _fs_map.end())
-            return iter->second.get();
+    zauto_write_lock l(_fs_lock);
```
acelyc111, line 45: Is this lock's scope too large?

src/block_service/block_service_manager.cpp
```diff
@@ -33,54 +33,49 @@ block_service_manager::block_service_manager()
...
 }
 
+block_service_manager::~block_service_manager()
+{
+    zauto_write_lock l(_fs_lock);
+    _fs_map.clear();
+    ddebug("close block service manager.");
+}
+
 block_filesystem *block_service_manager::get_block_filesystem(const std::string &provider)
```
acelyc111, line 43: How about rename to `get_or_create_block_filesystem `?

src/block_service/block_service_manager.cpp
```diff
@@ -33,54 +33,52 @@ block_service_manager::block_service_manager()
...
-        }
+    const char *arguments = dsn_config_get_value_string(
+        (std::string("block_service.") + provider).c_str(), "args", "", "args for block_service");
+
+    std::vector<std::string> args;
+    utils::split_args(arguments, args);
+    dsn::error_code err = fs->initialize(args);
+
+    if (dsn::ERR_OK == err) {
+        ddebug("create block filesystem ok for provider(%s)", provider.c_str());
```
levy5307, line 73: ddebug_f

src/block_service/block_service_manager.cpp
```diff
@@ -33,54 +33,52 @@ block_service_manager::block_service_manager()
...
-        if (fs == nullptr) {
-            derror("acquire block filesystem failed, provider = %s, provider_type = %s",
-                   provider.c_str(),
-                   provider_type);
-            return nullptr;
-        }
+    block_filesystem *fs =
+        utils::factory_store<block_filesystem>::create(provider_type, PROVIDER_TYPE_MAIN);
+    if (fs == nullptr) {
+        derror("acquire block filesystem failed, provider = %s, provider_type = %s",
```
levy5307, line 59: derror_f

src/block_service/block_service_manager.cpp
```diff
@@ -33,54 +33,52 @@ block_service_manager::block_service_manager()
...
 
-        block_filesystem *fs =
-            utils::factory_store<block_filesystem>::create(provider_type, PROVIDER_TYPE_MAIN);
-        if (fs == nullptr) {
-            derror("acquire block filesystem failed, provider = %s, provider_type = %s",
-                   provider.c_str(),
-                   provider_type);
-            return nullptr;
-        }
+    block_filesystem *fs =
```
levy5307, line 56: I think it better to use `std::unique_ptr<block_filesystem>` here

zhangyifan27: The `block_filesystem` will be inserted into `_fs_map` and returned, so maybe using `unique_ptr` is not appropriate.

src/block_service/block_service_manager.cpp
```diff
@@ -33,55 +33,53 @@ block_service_manager::block_service_manager()
...
+        (std::string("block_service.") + provider).c_str(), "args", "", "args for block_service");
+
+    std::vector<std::string> args;
+    utils::split_args(arguments, args);
+    dsn::error_code err = fs->initialize(args);
+
+    if (dsn::ERR_OK == err) {
+        ddebug_f("create block filesystem ok for provider {}", provider);
+        zauto_write_lock l(_fs_lock);
+        _fs_map.emplace(provider, std::unique_ptr<block_filesystem>(fs));
```
acelyc111, line 75: Better to check there is no existed block_filesystem with the same name in the map

acelyc111, line 75: Better to check there is no existed block_filesystem with the same name in the map. If it exist, the former will be overwrited and the returned pointer is invalid to access.

zhangyifan27: We checked it in L45-L51.

neverchanje (approved):

acelyc111 (approved):

hycdong (approved):

PR #662, opened 2020-11-10
fix(http): fix message buffers size check

hycdong:
In previous pr #593, http server can parse content-type of HTTP request. The incoming http request will parse into message_ex structure, msg->buffers[3] indicate the content-type, the msg->buffers will be resized as 4. When parsing message into http _request, the size of buffers will be checked. However, function `parse` still consider the right size of buffers is 3, it will reject all requests. This pull request fix it, add BUFFER_SIZE and update related unit tests.

### DIFF
--- a/src/http/http_message_parser.h
+++ b/src/http/http_message_parser.h
@@ -38,6 +38,8 @@ namespace dsn {
 
 DEFINE_CUSTOMIZED_ID(network_header_format, NET_HDR_HTTP)
 
+#define BUFFER_SIZE 4
+
 // Incoming HTTP requests will be parsed into:
 //
 //    msg->header->rpc_name = "RPC_HTTP_SERVICE"
--- a/src/http/http_message_parser.h
+++ b/src/http/http_message_parser.h
@@ -38,7 +38,8 @@ namespace dsn {
 
 DEFINE_CUSTOMIZED_ID(network_header_format, NET_HDR_HTTP)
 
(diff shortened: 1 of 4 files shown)


src/http/http_message_parser.h
```diff
@@ -38,6 +38,8 @@ namespace dsn {
 
 DEFINE_CUSTOMIZED_ID(network_header_format, NET_HDR_HTTP)
 
+#define BUFFER_SIZE 4
```
neverchanje, line 41:
```suggestion
// Number of blobs that a message_ex contains.
#define HTTP_MSG_BUFFER_NUM 4
```

acelyc111 (approved):

src/http/http_message_parser.h
```diff
@@ -38,6 +38,9 @@ namespace dsn {
 
 DEFINE_CUSTOMIZED_ID(network_header_format, NET_HDR_HTTP)
 
+// Number of blobs that a message_ex contains.
+#define HTTP_MSG_BUFFERS_NUM 4
```
levy5307, line 42: I think `const uint8_t` is better here

neverchanje (approved):

src/http/http_message_parser.h
neverchanje: Using constexpr or macro is basically the same, but using `const` is not a good practice, because the var is loaded in runtime.

PR #663, opened 2020-11-10
fix(http): fix parse path from message

hycdong:
In #594, it simplified http path, but it left a problem. For example:
The http request sent to meta server to list all apps is like `ip:port/meta/apps`, http server will add a tuple like `<'meta/apps', 'apps_handler'>`, when server receives a http request, it will parse path from request, and find the handler by path. 
In current implementation, the registered path is like `meta/apps`, and the path parsed from request is `/meta/apps`, as a result, the handler can not be found. 
This pr fix this problem, the path parsed from request will be `meta/apps` as the implementation before #594.

### DIFF
--- a/src/http/http_server.cpp
+++ b/src/http/http_server.cpp
@@ -151,7 +151,27 @@ void http_server::serve(message_ex *msg)
     if (!unresolved_path.empty() && *unresolved_path.crbegin() == '\0') {
         unresolved_path.pop_back();
     }
-    ret.path = std::move(unresolved_path);
+
+    // parse path
+    std::vector<std::string> args;
+    boost::split(args, unresolved_path, boost::is_any_of("/"));
+    std::vector<std::string> real_args;
+    for (std::string &arg : args) {
+        if (!arg.empty()) {
+            real_args.emplace_back(std::move(arg));
+        }
+    }
+    if (real_args.size() == 0) {
+        ret.path = "";
+    } else {
+        std::string path = real_args[0];
+        for (int i = 1; i < real_args.size(); i++) {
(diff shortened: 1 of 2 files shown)


src/http/http_server.cpp
```diff
@@ -151,7 +151,27 @@ void http_server::serve(message_ex *msg)
...
+    }
+    if (real_args.size() == 0) {
+        ret.path = "";
+    } else {
+        std::string path = real_args[0];
+        for (int i = 1; i < real_args.size(); i++) {
+            path += '/';
+            path += real_args[i];
+        }
+        ddebug("hyc: path = %s", path.c_str());
```
acelyc111, line 172: Remove it?

acelyc111 (approved):

src/http/http_server.cpp
```diff
@@ -151,7 +151,26 @@ void http_server::serve(message_ex *msg)
...
+    // parse path
+    std::vector<std::string> args;
+    boost::split(args, unresolved_path, boost::is_any_of("/"));
+    std::vector<std::string> real_args;
+    for (std::string &arg : args) {
+        if (!arg.empty()) {
+            real_args.emplace_back(std::move(arg));
+        }
+    }
+    if (real_args.size() == 0) {
```
levy5307, line 164:
```
 std::string path;
 for (int i = 0; i < real_args.size(); i++) {
            path += real_args[i];
            path += '/';
}
ret.path = std::move(path.empty() ? "": path.substr(value.size() - 1));
```

neverchanje (approved):

PR #667, opened 2020-11-11
fix: compile error of zookeeper 3.4.10 on gcc9

neverchanje:
## compilation error

```
src/zookeeper.c: In function 'format_endpoint_info':
src/zookeeper.c:3469:21: error: '%d' directive writing between 1 and 5 bytes into a region of size between 0 and 127 [-Werror=format-overflow=]
 3469 |     sprintf(buf,"%s:%d",addrstr,ntohs(port));
      |                     ^~
src/zookeeper.c:3469:17: note: directive argument in the range [0, 65535]
 3469 |     sprintf(buf,"%s:%d",addrstr,ntohs(port));
      |                 ^~~~~~~
In file included from /usr/include/stdio.h:867,
                 from ./include/zookeeper.h:29,
                 from src/zookeeper.c:27:
/usr/include/x86_64-linux-gnu/bits/stdio2.h:36:10: note: '__builtin___sprintf_chk' output between 3 and 134 bytes into a destination of size 128
   36 |   return __builtin___sprintf_chk (__s, __USE_FORTIFY_LEVEL - 1,
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   37 |       __bos (__s), __fmt, __va_arg_pack ());
      |       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
```

GCC9 generates errors on the zookeeper c client 3.4.10. So I try to upgrade it to 3.5.8, and it produces no error then.
One potential risk 3.5.8 clients could be minor incompatibilities towards 3.4.x zk-servers.

Some threads had related discussions: 
- [zookeeper 3.4 is end of life](http://zookeeper-user.578899.n2.nabble.com/ANNOUNCE-Apache-ZooKeeper-3-4-End-of-Life-1st-June-2020-td7584875.html)
- [zookeeper 3.5 is theoretically compatible with 3.4 server](http://zookeeper-user.578899.n2.nabble.com/Is-3-5-5-client-compatible-with-3-4-x-servers-td7584161.html)

### DIFF
--- a/include/dsn/tool-api/aio_task.h
+++ b/include/dsn/tool-api/aio_task.h
@@ -51,8 +51,6 @@ public:
     // filled by apps
     dsn_handle_t file;
     void *buffer;
-    bool support_write_vec; // if the aio provider supports write buffer vector
-    std::vector<dsn_file_buffer_t> *write_buffer_vec; // only used if support_write_vec is true
     uint32_t buffer_size;
     uint64_t file_offset;
 
@@ -64,8 +62,6 @@ public:
     aio_context()
         : file(nullptr),
           buffer(nullptr),
-          support_write_vec(false),
-          write_buffer_vec(nullptr),
           buffer_size(0),
           file_offset(0),
           type(AIO_Invalid),
--- a/src/aio/aio_provider.cpp
+++ b/src/aio/aio_provider.cpp
@@ -33,12 +33,9 @@ aio_provider::aio_provider(disk_engine *disk) : _engine(disk) {}
 
 service_node *aio_provider::node() const { return _engine->node(); }
 
-void aio_provider::complete_io(aio_task *aio,
-                               error_code err,
-                               uint32_t bytes,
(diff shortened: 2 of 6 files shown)


acelyc111 (approved):

neverchanje:
Well, https://github.com/zk-ruby/zookeeper/issues/85#issuecomment-465333458 there's also an approach that could simply fix this problem.

```diff
- ./configure --enable-static=yes --enable-shared=no --prefix=${TP_OUTPUT} --with-pic=yes
+ CFLAGS=-Wno-error=format-overflow ./configure --enable-static=yes --enable-shared=no --prefix=${TP_OUTPUT} --with-pic=yes
```

Suppressing the error so that compilation will pass. I would prefer fixing with minimum modifications though it seems tricky.

PR #670, opened 2020-11-20
feat(security): implement replica server access controller

levy5307:


src/replica/replica_2pc.cpp
```diff
@@ -40,6 +41,10 @@ void replica::on_client_write(dsn::message_ex *request, bool ignore_throttling)
 {
     _checker.only_one_thread_access();
 
+    if (!_access_controller->allowed(request)) {
```
neverchanje, line 44: Remote commands are not controlled. You restrict only the write/read.

neverchanje (approved):

hycdong (approved):

PR #672, opened 2020-11-24
feat(security): implement reset interface for replica access controller

levy5307:


src/replica/replica.h
```diff
@@ -127,6 +127,7 @@ class replica : public serverlet<replica>, public ref_counter, public replica_ba
     void update_throttle_env_internal(const std::map<std::string, std::string> &envs,
                                       const std::string &key,
                                       throttling_controller &cntl);
+    void update_ac_allowed_users(const std::map<std::string, std::string> &envs);
```
acelyc111, line 130: I think use 'access' instead of 'ac' is ok, not good to short partial words of a function name.

levy5307: `ac` means `access_controller`, but it's too long if we use `update_access_controller_allowed_users`

neverchanje: `ac` is acceptable to me. I'd recommend adding a comment above the function declaration.

A must-do is to validate this env here: https://github.com/XiaoMi/rdsn/blob/master/src/meta/app_env_validator.cpp.

hycdong: I agree with @neverchanje , I will expect adding comment for this function in this pull request and the validation of `allowed_user` in your further pull request.

src/runtime/security/replica_access_controller.h
```diff
@@ -27,10 +27,12 @@ class replica_access_controller : public access_controller
 public:
     replica_access_controller(const std::string &name);
     bool allowed(message_ex *msg);
+    void reset(const std::string &users);
```
hycdong, line 30: I think `update` or other name may be better, `reset` sounds like reset to default value.

hycdong (approved):

neverchanje (approved):

PR #654, opened 2020-10-29
feat(split): parent group update partition count

hycdong:
## Simple partition split process
1. meta receives client partition split request, and change partition count #286
2. replica notices partition count changed during on_config_sync #653 
3. parent partition create child partition #291
4. parent prepare states for child to learn #299
5. child partition async learn states from parent #309 #319
6. child notify parent catch up #390
7. update child group partition count #645 
8. register child partition #391
9. **update parent group partition count**

More partition split discussion in issue #69 and partition split design doc
This pr solves the part of ninth step of partition split, which is bold in process description.

## What this pr solve
When primary parent registers child succeed, it will update its own partition_count, then broadcast this to secondary through group_check. Besides, this pull request also adds related unit test.

PR #675, opened 2020-11-26
feat(split): secondary start split

hycdong:
## Simple partition split process
1. meta receives client partition split request, and change partition count #286
2. replica notices partition count changed during on_config_sync #653 
3. **parent partition create child partition** #291
4. parent prepare states for child to learn #299
5. child partition async learn states from parent #309 #319
6. child notify parent catch up #390
7. update child group partition count #645 
8. register child partition #391
9. update parent group partition count #654

More partition split discussion in issue #69 and partition split design doc
This pr solves the part of third step of partition split, which is bold in process description.

## What this pr solve
Primary parent notices that it should start partition split by `on_config_sync`, then it will pass it to secondary through group_check by calling `broadcast_group_check`, and put split related filed in `group_check_request`, when secondary receives this request, it will call `trigger_secondary_parent_split`.


src/replica/split/replica_split_manager.h
```diff
@@ -33,6 +33,12 @@ class replica_split_manager : replica_base
     int32_t get_partition_version() const { return _partition_version.load(); }
     gpid get_child_gpid() const { return _child_gpid; }
     void set_child_gpid(gpid pid) { _child_gpid = pid; }
+    bool is_splitting() const
+    {
+        return _child_gpid.get_app_id() > 0 && _child_init_ballot > 0 &&
+               _split_status == split_status::SPLITTING;
+    }
+    split_status::type get_meta_split_status() { return _meta_split_status; }
```
neverchanje, line 41: Why don't simply call it `get_split_status`?

hycdong: There are two different status, `_split_status` is replica local split status, and `_meta_split_status` is split_status from meta server through `on_config_sync`.

src/replication.thrift
```diff
@@ -180,8 +180,9 @@ struct group_check_request
     // their WALs after this decree.
     5:optional i64          confirmed_decree;
 
-    // Used to deliver child gpid during partition split
-    6:optional dsn.gpid     child_gpid;
+    // Used to deliver child gpid and meta_split_status during partition split
+    6:optional dsn.gpid     child_gpid;
```
neverchanje, line 184: How primary knows its child_gpid? I think it certainly can calculates out on its own gpid.
Such as parent_gpid(1, 4) => child_gpid(1, 4+partition-count)
If primary can, the secondary can too.

neverchanje:
```
enum split_status
{
    // idle state
    NOT_SPLIT,
    // A replica is splitting into two replicas, one call parent, one call child.
    SPLITTING,
    PAUSING,
    PAUSED,
    // After the split successfully cancelled, the state turns into NOT_SPLIT.
    CANCELING
}
```

src/replication.thrift
hycdong: Primary calculate child_gpid in `trigger_primary_parent_split`, if it found the partition_count sync from meta server is double of its own value: 
https://github.com/XiaoMi/rdsn/blob/d850ee74917e1145fc6125bb174ccf28233dc3fe/src/replica/split/replica_split_manager.cpp#L1136-L1140
It is a safe solution to sync it to secondary, if any bug happened, such as duplicated split, secondary may get wrong child_gpid, leading dangerous result.

neverchanje (approved):

acelyc111 (approved):

PR #676, opened 2020-11-26
feat(split): child copy mutation asynchronously

hycdong:
During partition split, parent partition should send mutation to child partition, child will copy this mutation and prepare it. This pull request is about child partition copying mutations asynchronously.
When parent is splitting, no matter it is primary or secondary, it should copy mutation to child, by calling function `copy_mutation`, child partition will put it into its own prepare list and add it into shared log and private log asynchronously.


src/replica/split/replica_split_manager.cpp
```diff
@@ -1172,5 +1172,87 @@ void replica_split_manager::trigger_secondary_parent_split(
...
+        _stub->split_replica_error_handler(
+            _replica->_split_states.parent_gpid, [mu](replica_split_manager *split_mgr) {
+                split_mgr->parent_cleanup_split_context();
+                split_mgr->on_copy_mutation_reply(
+                    ERR_OK, mu->data.header.ballot, mu->data.header.decree);
+            });
+        return;
+    }
+
+    if (!_replica->_split_states.is_prepare_list_copied) {
```
neverchanje, line 1207:
```suggestion
    if (!_replica->_split_states.is_prepare_list_copied) {
```
What does this statement mean? If prepare-list hasn't copied, why don't we proceed with copying? Why return?

src/replica/split/replica_split_manager.cpp
```diff
@@ -1172,5 +1172,87 @@ void replica_split_manager::trigger_secondary_parent_split(
...
+
+    mu->data.header.pid = get_gpid();
+    _replica->_prepare_list->prepare(mu, partition_status::PS_SECONDARY);
+    if (!mu->is_sync_to_child()) { // child copy mutation asynchronously
+        if (!mu->is_logged()) {
+            mu->set_logged();
+        }
+        mu->log_task() = _stub->_log->append(
+            mu, LPC_WRITE_REPLICATION_LOG, tracker(), nullptr, get_gpid().thread_hash());
+        _replica->_private_log->append(mu, LPC_WRITE_REPLICATION_LOG_COMMON, tracker(), nullptr);
```
neverchanje, line 1235:
```suggestion
        _replica->_private_log->append(mu, LPC_WRITE_REPLICATION_LOG_COMMON, tracker(), nullptr, get_gpid().thread_hash());
```

src/replica/split/replica_split_manager.cpp
hycdong: When parent partition set valid child_gpid, it will be considered doing partition split, child partition should copy each mutation. At the same time, child partition will be created, and its status is PS_PARTITION_SPLIT, parent will prepare states for child including prepare list. Parent partition and child partition may not execute in the same thread, it is possible for parent partition start to send mutation but child partition has not copied parent prepare list. In this case, child partition should not copy mutation.

src/replica/split/replica_split_manager.cpp
```diff
@@ -1160,5 +1160,87 @@ void replica_split_manager::trigger_secondary_parent_split(
...
 }
 
+// ThreadPool: THREAD_POOL_REPLICATION
+void replica_split_manager::copy_mutation(mutation_ptr &mu) // on parent partition
+{
+    dassert_replica(_child_gpid.get_app_id() > 0, "child_gpid({}) is invalid", _child_gpid);
+
+    // TODO(hyc): if copy mutation synchronously, add flags
+
+    mutation_ptr new_mu = mutation::copy_no_reply(mu);
```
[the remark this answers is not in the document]
hycdong: Well, `copy_no_reply` will only create a new mutation, child partition will apply it in function `on_copy_mutation`. Besides, no matter copying asynchronously or synchronously, parent will call function `copy_mutation`, child will call function `on_copy_mutation`.

neverchanje (approved):

acelyc111 (approved):

PR #678, opened 2020-11-26
fix: cluster_name of ddl client from zk root

empiredan:
Currently cluster_name from ddl client is extracted from zk root. However, Zk root format like "/pegasus/<cluster_name>" is not necessarily adopted by all users. Once it's changed to another format, pegasus shell just prints "unknown" when logged in. Meta service should response with the real cluster name instead.


neverchanje (approved):

neverchanje:
Hi, @empiredan 

We are implementing a brand new pegasus-shell called [admin-cli](https://github.com/pegasus-kv/admin-cli). It will be released very soon, perhaps this week or next.

I'm considering if we should keep the command `cluster-name` or not. There will be better ways to display the cluster name than typing a command.

empiredan:
> Hi, @empiredan
> 
> We are implementing a brand new pegasus-shell called [admin-cli](https://github.com/pegasus-kv/admin-cli). It will be released very soon, perhaps this week or next.
> 
> I'm considering if we should keep the command `cluster-name` or not. There will be better ways to display the cluster name than typing a command.

@neverchanje That's wonderful ! I'll learn about admin-cli.

By the way, has there ever been a command named `cluster-name` in shell ? As far as I known the cluster name is only shown immediately after the shell is started. Or the problem is that whether it's necessary for users to check cluster name while they are working in shell ? Or any better way instead ?

acelyc111 (approved):

PR #679, opened 2020-11-27
feat(split): add meta control split

hycdong:
Partition split may copy lots of data, making heavy load, it is safe to provide the function pause or cancel it. As a result, we provide three types to control split: pause, restart, cancel. 
For pause and restart, admin user can set `parent_pidx` as a valid partition index, server will only try to pause or restart that specific partition split process. If `parent_pidx = -1`, server will pause all spitting partition or restart all paused partition. 
For cancel, `parent_pidx` will always be -1, but it should set `old_partition_count` to avoid duplicated cancel request.
This pull request implement ddl-client control split interface and how meta server handle control split request. What meta server does is update `split_status` for the partition, and this value will pass to primary replica through on_config_sync. How replica server handle different split_status will be added in further pull request.


src/meta/meta_split_service.h
```diff
@@ -48,6 +57,19 @@ class meta_split_service
     void
     on_add_child_on_remote_storage_reply(error_code ec, register_child_rpc rpc, bool create_new);
 
+    static const std::string control_type_str(split_control_type::type type)
```
levy5307, line 60: What about use `ENUM_REG` for `split_control_type`?

hycdong: If `split_control_type` is defined by ENUM_REG, the string will be like `dsn::replication::split_control_type::PSC_PAUSE`, the function `control_type_str` is only used when printing log or hint message to show the control type, I think the string using ENUM_REG is a little bit too long.

src/replication.thrift
```diff
@@ -890,6 +890,38 @@ struct start_partition_split_response
     2:string            hint_msg;
 }
 
+enum split_control_type
+{
+    PSC_PAUSE,
```
acelyc111, line 895: Now that the enums are defined in shrift, mean we will use them as `split_control_type:: PSC_PAUSE ` , is it needed to add the confusing prefix `PSC`？

src/meta/meta_split_service.cpp
```diff
@@ -293,5 +293,165 @@ void meta_split_service::on_add_child_on_remote_storage_reply(error_code ec,
...
+        derror_f("{} split for app({}) failed, {}",
+                 control_type_str(control_type),
+                 app_name,
+                 response.hint_msg);
+        return;
+    }
+
+    auto iter = app->helpers->split_states.status.find(parent_pidx);
+    if (iter == app->helpers->split_states.status.end()) {
+        response.err = control_type == split_control_type::PSC_PAUSE ? ERR_CHILD_REGISTERED
```
Shuo-Jia, line 347: Why return different err code?  shoudn't it be all `ERR_INVALID_STATE`?

src/meta/meta_split_service.cpp
```diff
@@ -293,5 +293,165 @@ void meta_split_service::on_add_child_on_remote_storage_reply(error_code ec,
...
+            derror_f("{} split for app({}) failed, {}",
+                     control_type_str(control_type),
+                     app_name,
+                     response.hint_msg);
+            return;
+        }
+
+        if (iter->second == split_status::PAUSING || iter->second == split_status::PAUSED) {
+            dwarn_f("duplicated pause request for app({}), partition[{}]", app_name, parent_pidx);
+            response.err = ERR_OK;
```
Shuo-Jia, line 370: `ERR_OK`?

src/meta/meta_split_service.cpp
```diff
@@ -293,5 +293,165 @@ void meta_split_service::on_add_child_on_remote_storage_reply(error_code ec,
...
+            return;
+        }
+
+        iter->second = split_status::PAUSING;
+        response.err = ERR_OK;
+        ddebug_f("app({}) partition[{}] pausing split succeed", app_name, parent_pidx);
+    } else { // restart
+        if (iter->second == split_status::SPLITTING) {
+            dwarn_f("duplicated restart request for app({}) partition[{}]", app_name, parent_pidx);
+            response.err == ERR_OK;
```
Shuo-Jia, line 380: same, I think it should return `err` like `err_invalid_stat` but not `ok`

Shuo-Jia: And the `log` show is `duplicated restart`, the `SPLITTING` may be not trigered by `restart`

src/meta/meta_split_service.cpp
```diff
@@ -293,5 +293,165 @@ void meta_split_service::on_add_child_on_remote_storage_reply(error_code ec,
...
+                                                dsn::enum_to_string(iter->second)));
+            derror_f("{} split for app({}) failed, {}",
+                     control_type_str(control_type),
+                     app_name,
+                     response.hint_msg);
+            return;
+        }
+
+        iter->second = split_status::SPLITTING;
+        response.err = ERR_OK;
```
Shuo-Jia, line 397: I think you only validate the state whether invalid or not but no need so many if...also `state` case, like:
```c++
if (splliting) {
pause
} else {
return invalid state
}

if pause {
restart
} else {
invalid state
}
```

src/replication.thrift
hycdong: PSC is short for partition split control, actually it is redundant, I will remove it.

src/meta/meta_split_service.h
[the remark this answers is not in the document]
hycdong: As I left the comment to @levy5307, this function is used to print short name of control_type, and it is hardly to add another new type, I think it is okay to remain the implementation here.

include/dsn/dist/replication/replication_ddl_client.h
```diff
@@ -194,6 +194,11 @@ class replication_ddl_client
     // partition split
     error_with<start_partition_split_response> start_partition_split(const std::string &app_name,
                                                                      int partition_count);
+    error_with<control_split_response>
+    control_partition_split(const std::string &app_name,
+                            split_control_type::type control_type,
+                            const int32_t parent_pidx,
+                            const int32_t old_partition_count);
```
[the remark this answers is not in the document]
hycdong: Yes, you're right~ I have already added three interfaces easier  to use.

src/meta/meta_split_service.cpp
hycdong: If user tries to pause split has already finished, ERR_CHILD_REGISTERED may be better to descibe split is finished.

Shuo-Jia: it's also ok(actually the hint message is enough)

src/meta/meta_split_service.cpp
hycdong: This is an old implementation problem, I have already updated and refacored it as your comment.

acelyc111 (approved):

neverchanje (approved):

PR #681, opened 2020-12-04
feat(split): replica server handle pause and cancel status

hycdong:
When meta server pause or cancel split, partition's split_status will be `pausing` or `canceling`(#679), and this split_status will transfer replica server through on_config_sync(#653). This pr implements how replica server handle those two split_status.
When primary parent partition receives `pausing` or `canceling` split_status from meta server, it will set split_status into `not_split` and broadcast it through `group_check`, secondary parent partition also set its split_status into` not_split` and set `is_split_stopped = true` in `group_check_response`. Primary parent partition will check if all partitions in its group have already paused or canceled split, if all stop succeed, it will send notification to meta, which will be implmented in next pull request.


neverchanje (approved):

src/replica/split/replica_split_manager.cpp
```diff
@@ -1257,5 +1273,81 @@ void replica_split_manager::on_copy_mutation_reply(error_code ec,
...
+// ThreadPool: THREAD_POOL_REPLICATION
+void replica_split_manager::parent_stop_split(
+    split_status::type meta_split_status) // on parent partition
+{
+    dassert_replica(status() == partition_status::PS_PRIMARY ||
+                        status() == partition_status::PS_SECONDARY,
+                    "wrong partition_status({})",
+                    enum_to_string(status()));
+    dassert_replica(_split_status == split_status::SPLITTING ||
+                        _split_status == split_status::NOT_SPLIT,
```
Shuo-Jia, line 1285: Why allow stop under `NOT_SPLIT`?

src/replica/split/replica_split_manager.cpp
```diff
@@ -1257,5 +1273,81 @@ void replica_split_manager::on_copy_mutation_reply(error_code ec,
...
+    const std::shared_ptr<group_check_request> &req,
+    const std::shared_ptr<group_check_response> &resp) // on primary parent partition
+{
+    if (!req->__isset.meta_split_status || (req->meta_split_status != split_status::PAUSING &&
+                                            req->meta_split_status != split_status::CANCELING)) {
+        // partition is not executing split or not stopping split
+        return;
+    }
+
+    if (!resp->__isset.is_split_stopped || !resp->is_split_stopped) {
```
Shuo-Jia, line 1323: `__isset.is_split_stopped` and `is_split_stopped` defference?

src/replica/split/replica_split_manager.cpp
```diff
@@ -1257,5 +1273,81 @@ void replica_split_manager::on_copy_mutation_reply(error_code ec,
...
+        // secondary has not stopped split
+        return;
+    }
+
+    _replica->_primary_states.split_stopped_secondary.insert(req->node);
+    auto count = 0;
+    for (auto &iter : _replica->_primary_states.statuses) {
```
Shuo-Jia, line 1330: can `count`  be stored as `global` variable, but not repeat compute when check?

src/replica/split/replica_split_manager.cpp
Shuo-Jia: And, if we send `cancel` or `pause` by mistake, dassert will cause crash, is your expect?

hycdong: When learn happened, parent partition may stop split, set it _split_status as `NOT_SPLIT`, meta server won't know it, so it is possible when a pause or cancel split request sync to parent partition, its split_status is NOT_SPLIT.

src/replica/split/replica_split_manager.cpp
hycdong: It's safety to check it each time, because `_primary_states.statuses` may change during learn, and this case is not always happened, it's okay to check it.

src/replica/split/replica_split_manager.cpp
hycdong: > And, if we send cancel or pause by mistake, dassert will cause crash? is your expect?
Pause or cancel split request will send to meta server, meta server should check if this table is splitting, you can reference pr[679](https://github.com/XiaoMi/rdsn/pull/681). Besides, parent partition will not set split_status as PAUSING and CANCELING, when it receives pause or cancel request, it will set it NOT_SPLIT. I don't know if I explain it clearly, you can comment to me if you have any questions.

Shuo-Jia (approved):

PR #682, opened 2020-12-07
feat: support to modify configs without restart

levy5307:
Support to modify configs without restart. 

***NOTE***

1. String modifications is not supported.

2. A example for http interface to update config:
```
http://localhost:34801/updateConfig?max_concurrent_bulk_load_downloading_count=5
```
3. Please add `DSN_TAG_VARIABLE(name, FT_MUTABLE)` if you want to make a config mutable.

4. I will add validator for config update, in next pull request.


acelyc111 (approved):

hycdong:
I review this pull request roughly, I think this pr implement a http get interface to support user update some configurations dynamically, we don't need to restart server any more. I recommend that you add some comments in code or descriptions, including:
- The example of how to update the flag, the sample request and response
- The restriction of this pr, such as string config is not supported, the name of configuration should be `FLAGS_##name`, old configuration is not supported etc.

By the way, would you consider adding another interface for listing names of the config who can be updated through http?

levy5307:
> I review this pull request roughly, I think this pr implement a http get interface to support user update some configurations dynamically, we don't need to restart server any more. I recommend that you add some comments in code or descriptions, including:
> 
> * The example of how to update the flag, the sample request and response
> * The restriction of this pr, such as string config is not supported, the name of configuration should be `FLAGS_##name`, old configuration is not supported etc.
> 
> By the way, would you consider adding another interface for listing names of the config who can be updated through http?

Yes, I will add a http interface to list names of all of the configs.

acelyc111 (dismissed):

hycdong (dismissed):

src/utils/flags.cpp
```diff
@@ -37,6 +40,19 @@ class flag_data
...
+        type tmpval_##type_enum;                                                                   \
+        if (!dsn::buf2##suffix(val, tmpval_##type_enum)) {                                         \
+            return error_s::make(ERR_INVALID_PARAMETERS, fmt::format("{} in invalid", val));       \
+        }                                                                                          \
+        value<type>() = tmpval_##type_enum;                                                        \
+        break
+
+#define FLAG_DATA_UPDATE_STRING()                                                                  \
+    case FV_STRING:                                                                                \
+        return error_s::make(ERR_NO_PERMISSION, "string modifications are not supported")
```
neverchanje, line 54: NO_PERMISSION literally means you have no authorization to operate something. Use ERR_INVALID_PARAMETERS instead, which is existing code.

neverchanje (approved):

hycdong (approved):

PR #683, opened 2020-12-08
feat(security): enable access replica when appenv `replica_access_controller.allowed_users` is empty

levy5307:
enable access replica when appenv `replica_access_controller.allowed_users` is empty.
So the access to replica is enabled when the appenv is not set.

### DIFF
--- a/src/runtime/security/replica_access_controller.cpp
+++ b/src/runtime/security/replica_access_controller.cpp
@@ -34,7 +34,7 @@ bool replica_access_controller::allowed(message_ex *msg)
 
     {
         utils::auto_read_lock l(_lock);
-        if (_users.find(user_name) == _users.end()) {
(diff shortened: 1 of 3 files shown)


Shuo-Jia (dismissed):

src/runtime/security/replica_access_controller.cpp
```diff
@@ -34,7 +34,7 @@ bool replica_access_controller::allowed(message_ex *msg)
 
     {
         utils::auto_read_lock l(_lock);
-        if (_users.find(user_name) == _users.end()) {
+        if (!_users.empty() && _users.find(user_name) == _users.end()) {
```
hycdong, line 37: When users are empty, we always allow it, why would we do it?

neverchanje (approved):

Shuo-Jia (approved):

PR #686, opened 2020-12-15
feat(bulk-load): support user-define remote storage root path

hycdong:
In previous implementation, bulk load will download files from specific remote storage path, which is defined in config file, name is `bulk_load_provider_root`. This pull request removes this config, and adds `remote_root_path` for starting bulk load request, server will use it as the root path of remote file storage.

### DIFF
--- a/src/meta/meta_bulk_load_service.cpp
+++ b/src/meta/meta_bulk_load_service.cpp
@@ -59,6 +59,7 @@ void bulk_load_service::on_start_bulk_load(start_bulk_load_rpc rpc)
     error_code e = check_bulk_load_request_params(request.app_name,
                                                   request.cluster_name,
                                                   request.file_provider_type,
+                                                  request.remote_root_path,
                                                   app->app_id,
                                                   app->partition_count,
                                                   hint_msg);
@@ -86,6 +87,7 @@ void bulk_load_service::on_start_bulk_load(start_bulk_load_rpc rpc)
 error_code bulk_load_service::check_bulk_load_request_params(const std::string &app_name,
                                                              const std::string &cluster_name,
                                                              const std::string &file_provider,
+                                                             const std::string &remote_root_path,
                                                              const int32_t app_id,
                                                              const int32_t partition_count,
                                                              std::string &hint_msg)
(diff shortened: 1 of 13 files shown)


src/replica/bulk_load/replica_bulk_loader.cpp
```diff
@@ -409,6 +412,10 @@ error_code replica_bulk_loader::download_sst_files(const std::string &app_name,
                 uint64_t f_size = 0;
                 error_code ec = _stub->_block_service_manager.download_file(
                     remote_dir, local_dir, f_meta.name, fs, f_size);
+                if (ec == ERR_OK && f_size == 0) {
```
acelyc111, line 415: It's a bit of confused, how about define a new code like `ERR_DOWNLOADED`?

src/meta/meta_bulk_load_service.cpp
```diff
@@ -59,6 +59,7 @@ void bulk_load_service::on_start_bulk_load(start_bulk_load_rpc rpc)
     error_code e = check_bulk_load_request_params(request.app_name,
                                                   request.cluster_name,
                                                   request.file_provider_type,
+                                                  request.remote_root_path,
```
acelyc111, line 62: Is there any compatiable issue when an old shell tools send a bulk load request?

hycdong: Yes, the old shell tools will send invalid request lack of remote_root_path, I will add related code for shell command. Consider that bulk load is not already used in production and we plan to use admin-cli tool instead of shell tools, and bulk load related command has not already been implemented in admin-cli tool, I think this compatiable problem can be acceptable.

src/replica/bulk_load/replica_bulk_loader.cpp
hycdong: I remove this code, which is not related with this pr, I will add another pr to solve it.

acelyc111 (approved):

neverchanje (approved):

levy5307 (approved):

PR #690, opened 2020-12-16
feat: support to build a portable rocksdb binary

empiredan:
The process exits with "Illegal instruction (core dumped)" while trying to launch pegasus server on certain nodes. By trouble-shooting, we find "PORTABLE" is disabled while rocksdb is built. The pegasus server can be built and run normally on standard nodes with "PORTABLE" disabled. However, hardware varies in some other nodes where certain instructions are not supported. We have to achieve portability for these nodes at the cost of a little of performance.

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -61,6 +61,10 @@ function usage_build()
         echo "                         e.g., \"dsn_runtime_tests,dsn_meta_state_tests\","
         echo "                         if not set, then run all tests"
     fi
+
+    echo "   --enable_rocksdb_portable      build a portable rocksdb binary"
+    echo "   --enable_rocksdb_force_sse42   force building rocksdb binary with SSE4.2,"
+    echo "                                  even when portable is enabled"
 }
 function run_build()
 {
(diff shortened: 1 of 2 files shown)


neverchanje:
I prefer not to have a sse4.2 option, building it with portable only is acceptable to me. I had this idea too when I read the building script in nebulagraph: https://github.com/vesoft-inc/nebula/blob/master/third-party/externals/rocksdb.cmake#L22.

They also disable `-DWITH_TOOLS=OFF` btw. 

@acelyc111

empiredan:
> I prefer not to have a sse4.2 option, building it with portable & without SIMD is acceptable to me. I had this idea too when I read the building script in nebulagraph: https://github.com/vesoft-inc/nebula/blob/master/third-party/externals/rocksdb.cmake#L22.
> 
> They also disable `-DWITH_TOOLS=OFF` btw.
> 
> @acelyc111

Actually sse will be enabled automatically if cpu supports it while portable is disabled. It just forces sse if portable is enabled. It makes sense that if performance is important one will disable portable. I'll drop sse4.2 option.

acelyc111:
IMO, performance is more important than portable, we should build a platform optimized binary be default.
This patch just add a build option, I think it's acceptable.

acelyc111 (approved):

neverchanje (approved):

PR #693, opened 2020-12-17
feat(bulk_load): add start bulk load http interface

hycdong:
This pull request adds a new http interface for meta server to start bulk load.

### DIFF
--- a/src/meta/meta_http_service.cpp
+++ b/src/meta/meta_http_service.cpp
@@ -5,7 +5,6 @@
 #include <string>
 
 #include <dsn/c/api_layer1.h>
-#include <dsn/cpp/json_helper.h>
 #include <dsn/cpp/serialization_helper/dsn.layer2_types.h>
 #include <dsn/dist/replication/replication_types.h>
 #include <dsn/dist/replication/duplication_common.h>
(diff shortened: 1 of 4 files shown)


src/meta/meta_http_service.h
```diff
@@ -6,11 +6,15 @@
 
 #include <algorithm>
 
+#include <dsn/cpp/json_helper.h>
```
levy5307, line 9: you can move this line to the corresponding cpp file

src/meta/meta_http_service.cpp
```diff
@@ -611,6 +610,61 @@ void meta_http_service::query_duplication_handler(const http_request &req, http_
...
+    if (request.file_provider_type.empty()) {
+        resp.body = "file_provider_type should not be empty";
+        resp.status_code = http_status_code::bad_request;
+        return;
+    }
+    if (request.remote_root_path.empty()) {
+        resp.body = "remote_root_path should not be empty";
+        resp.status_code = http_status_code::bad_request;
+        return;
+    }
```
Shuo-Jia, line 651: Can this simple args validate put client side?

src/meta/meta_http_service.h
hycdong: This update is for further pull request, I will add structure in `.h` file using json serilization.

src/meta/meta_http_service.cpp
hycdong: In current implmetation, client and server both will validate arguments.

src/meta/meta_http_service.h
levy5307: So I think you can add this line in the future pull request.

hycdong: It is still useful for this pull request, I use `NON_MEMBER_JSON_SERIALIZATION` marco to make structure `start_bulk_load_request` json-serializable. As your comment shows, I can add this in `meta_http_service.cpp`, but considering it will finally move into `.h` file, I did it in this pull request.

levy5307: could you paste the definition of `NON_MEMBER_JSON_SERIALIZATION` here?

levy5307 (approved):

src/meta/meta_http_service.cpp
neverchanje: I will implement HTTP argument validation in later PR. I actually did it in my staging branch. But haven't ready for review yet. Got no time to make it steady.

neverchanje (approved):

PR #694, opened 2020-12-17
feat(security): treat negotiation succeed if server is old version

levy5307:
When you rolling_update a cluster, if the new config set enable_auth = true. The new version server will send negotiation message to the server which has old version. And it will return `ERR_HANDLER_NOT_FOUND` to the new version server, so the negotiation will always fail.

### DIFF
--- a/src/runtime/security/client_negotiation.cpp
+++ b/src/runtime/security/client_negotiation.cpp
@@ -50,10 +50,10 @@ void client_negotiation::list_mechanisms()
 void client_negotiation::handle_response(error_code err, const negotiation_response &&response)
 {
     if (err != ERR_OK) {
-        // ERR_HANDLER_NOT_FOUND means server is old version, which doesn't support authentiation
+        // ERR_HANDLER_NOT_FOUND means server is old version, which doesn't support authentication
         if (ERR_HANDLER_NOT_FOUND == err && !FLAGS_mandatory_auth) {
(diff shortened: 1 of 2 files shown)


acelyc111 (approved):

src/runtime/security/client_negotiation.cpp
```diff
@@ -50,7 +50,15 @@ void client_negotiation::list_mechanisms()
 void client_negotiation::handle_response(error_code err, const negotiation_response &&response)
 {
     if (err != ERR_OK) {
-        fail_negotiation();
+        // ERR_HANDLER_NOT_FOUND means server is old version, which doesn't support authentiation
+        if (ERR_HANDLER_NOT_FOUND == err && !FLAGS_mandatory_auth) {
+            ddebug_f("{}: treat negotiation succeed because server is old version, which doesn't "
+                     "support authentiation",
+                     _name);
```
neverchanje, line 57:
```suggestion
        // ERR_HANDLER_NOT_FOUND means server is old version, which doesn't support authentication
        if (ERR_HANDLER_NOT_FOUND == err && !FLAGS_mandatory_auth) {
            ddebug_f("{}: treat negotiation succeed because server is old version, which doesn't "
                     "support authentication",
                     _name);
```

neverchanje (approved):

Shuo-Jia (dismissed):

Shuo-Jia (approved):

neverchanje (approved):

PR #695, opened 2020-12-18
feat(bulk_load): add update compaction envs http interface

hycdong:
This pull request adds the http interface for meta server to update app_envs of manual_compaction.

### DIFF
--- a/src/meta/meta_http_service.cpp
+++ b/src/meta/meta_http_service.cpp
@@ -706,7 +706,7 @@ void meta_http_service::start_compaction_handler(const http_request &req, http_r
         return;
     }
 
-    // validate paramters
+    // validate parameters
     manual_compaction_info info;
     bool ret = json::json_forwarder<manual_compaction_info>::decode(req.body, info);
     if (!ret) {
@@ -765,7 +765,7 @@ void meta_http_service::start_compaction_handler(const http_request &req, http_r
         keys.emplace_back(replica_envs::MANUAL_COMPACT_MAX_CONCURRENT_RUNNING_COUNT);
         values.emplace_back(std::to_string(info.max_concurrent_running_count));
     }
-    update_app_env_handler(info.app_name, keys, values, resp);
+    update_app_env(info.app_name, keys, values, resp);
 }
 
 bool meta_http_service::redirect_if_not_primary(const http_request &req, http_response &resp)
@@ -791,10 +791,10 @@ bool meta_http_service::redirect_if_not_primary(const http_request &req, http_re
     return false;
 }
 
-void meta_http_service::update_app_env_handler(const std::string &app_name,
-                                               const std::vector<std::string> &keys,
-                                               const std::vector<std::string> &values,
-                                               http_response &resp)
+void meta_http_service::update_app_env(const std::string &app_name,
+                                       const std::vector<std::string> &keys,
+                                       const std::vector<std::string> &values,
+                                       http_response &resp)
 {
     configuration_update_app_env_request request;
     request.app_name = app_name;
--- a/src/meta/meta_http_service.cpp
+++ b/src/meta/meta_http_service.cpp
@@ -709,6 +709,7 @@ void meta_http_service::start_compaction_handler(const http_request &req, http_r
     // validate parameters
     manual_compaction_info info;
     bool ret = json::json_forwarder<manual_compaction_info>::decode(req.body, info);
+
     if (!ret) {
         resp.body = "invalid request structure";
         resp.status_code = http_status_code::bad_request;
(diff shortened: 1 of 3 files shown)


src/meta/meta_http_service.cpp
```diff
@@ -699,6 +700,74 @@ void meta_http_service::query_bulk_load_handler(const http_request &req, http_re
...
+        resp.body = "app_name should not be empty";
+        resp.status_code = http_status_code::bad_request;
+        return;
+    }
+    if (info.type.empty() || (info.type != "once" && info.type != "periodic")) {
+        resp.body = "type should ony be once or periodic";
+        resp.status_code = http_status_code::bad_request;
+        return;
+    }
+    if (info.target_level < -1) {
```
neverchanje, line 727:
```suggestion
    if (info.target_level <= -1) {
```

src/meta/meta_http_service.cpp
```diff
@@ -699,6 +700,74 @@ void meta_http_service::query_bulk_load_handler(const http_request &req, http_re
...
+        resp.status_code = http_status_code::bad_request;
+        return;
+    }
+    if (info.bottommost_level_compaction.empty() || (info.bottommost_level_compaction != "skip" &&
+                                                     info.bottommost_level_compaction != "force")) {
+        resp.body = "bottommost_level_compaction should ony be skip or force";
+        resp.status_code = http_status_code::bad_request;
+        return;
+    }
+    if (info.max_concurrent_running_count < 0) {
```
neverchanje, line 738:
```suggestion
    if (info.max_concurrent_running_count <= 0) {
```

src/meta/meta_http_service.h
```diff
@@ -89,11 +112,17 @@ class meta_http_service : public http_service
     void query_duplication_handler(const http_request &req, http_response &resp);
     void start_bulk_load_handler(const http_request &req, http_response &resp);
     void query_bulk_load_handler(const http_request &req, http_response &resp);
+    void start_compaction_handler(const http_request &req, http_response &resp);
 
 private:
     // set redirect location if current server is not primary
     bool redirect_if_not_primary(const http_request &req, http_response &resp);
 
+    void update_app_env_handler(const std::string &app_name,
```
neverchanje, line 121:
```suggestion
    void update_app_env(const std::string &app_name,
```

the "_handler" suffix is for http handler. this function is not handler.

src/meta/meta_http_service.cpp
hycdong: target_level is `number in range of [-1,num_levels], -1 means automatically, default is -1`

src/meta/meta_http_service.cpp
hycdong: max_concurrent_running_count default value is 0, it means no limit, you can see comments of structure `manual_compaction_info` in meta_http_service.h

src/meta/meta_http_service.cpp
neverchanje: OK, your comment said "greater than -1", so I supposed this is (-1, num_levels], which excludes -1. A proper and simple saying is ">= -1".

neverchanje (dismissed):

src/meta/test/meta_http_service_test.cpp
```diff
@@ -300,5 +309,49 @@ TEST_F(meta_bulk_load_http_test, query_bulk_load_request)
...
 
+TEST_F(meta_bulk_load_http_test, start_compaction_test)
+{
+    struct start_compaction_test
+    {
+        std::string request_json;
+        http_status_code expected_code;
+        std::string expected_response_json;
+    } tests[] = {
+        {"{\"app_name\":\"test_bulk_load\",\"type\":\"once\",\"target_level\":-1,\"bottommost_"
```
acelyc111, line 320: Better to use [string_literal](https://en.cppreference.com/w/cpp/language/string_literal) to simplify code.

src/meta/meta_http_service.cpp
```diff
@@ -699,6 +700,74 @@ void meta_http_service::query_bulk_load_handler(const http_request &req, http_re
...
+        resp.status_code = http_status_code::bad_request;
+        return;
+    }
+    if (info.app_name.empty()) {
+        resp.body = "app_name should not be empty";
+        resp.status_code = http_status_code::bad_request;
+        return;
+    }
+    if (info.type.empty() || (info.type != "once" && info.type != "periodic")) {
+        resp.body = "type should ony be once or periodic";
```
acelyc111, line 723:
```suggestion
        resp.body = "type should ony be 'once' or 'periodic'";
```

src/meta/meta_http_service.cpp
```diff
@@ -699,6 +700,74 @@ void meta_http_service::query_bulk_load_handler(const http_request &req, http_re
...
+        return;
+    }
+    if (info.target_level < -1) {
+        resp.body = "target_level should be greater than -1";
+        resp.status_code = http_status_code::bad_request;
+        return;
+    }
+    if (info.bottommost_level_compaction.empty() || (info.bottommost_level_compaction != "skip" &&
+                                                     info.bottommost_level_compaction != "force")) {
+        resp.body = "bottommost_level_compaction should ony be skip or force";
```
acelyc111, line 734:
```suggestion
        resp.body = "bottommost_level_compaction should ony be 'skip' or 'force'";
```

src/meta/meta_http_service.cpp
```diff
@@ -699,6 +700,74 @@ void meta_http_service::query_bulk_load_handler(const http_request &req, http_re
...
+    }
+    if (info.type == "periodic" && info.trigger_time.empty()) {
+        resp.body = "trigger_time should not be empty when type is periodic";
+        resp.status_code = http_status_code::bad_request;
+        return;
+    }
+
+    // create configuration_update_app_env_request
+    std::vector<std::string> keys;
+    std::vector<std::string> values;
```
acelyc111, line 751: There maybe problems when keys and values' size or order not matched, how about to use std::map ors std::pair?

hycdong: In struct configuration_update_app_env_request, key and value is std::vector, so I also use it here, besides, function `set_app_envs` will check if keys.size is equal to values.size, if keys size is not equal to values size, it will set `ERR_INVALID_PARAMETERS` in response.

acelyc111 (approved):

neverchanje (approved):

PR #696, opened 2020-12-18
feat(bulk_load): add query compaction status http interface

hycdong:
This pull request adds replica http interface for query app compaction status.

### DIFF
--- a/src/replica/replica.cpp
+++ b/src/replica/replica.cpp
@@ -443,6 +443,33 @@ std::string replica::query_compact_state() const
     return _app->query_compact_state();
 }
 
+const char *manual_compaction_status_to_string(manual_compaction_status status)
+{
+    switch (status) {
+    case CompactionFinish:
+        return "CompactionFinish";
+    case CompactionRunning:
+        return "CompactionRunning";
+    case CompactionQueue:
+        return "CompactionQueue";
+    default:
+        dassert(false, "");
+    }
+    return "wrongCompactionStatus";
+}
+
+manual_compaction_status replica::get_compact_status() const
+{
+    std::string compact_state = query_compact_state();
+    if (compact_state.find("recent start at") != std::string::npos) {
+        return CompactionRunning;
+    } else if (compact_state.find("recent enqueue at") != std::string::npos) {
+        return CompactionQueue;
+    } else {
+        return CompactionFinish;
+    }
+}
+
 // Replicas on the server which serves for the same table will share the same perf-counter.
 // For example counter `table.level.RPC_RRDB_RRDB_MULTI_PUT.latency(ns)@test_table` is shared by
 // all the replicas for `test_table`.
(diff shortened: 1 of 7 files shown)


src/replica/replica.cpp
```diff
@@ -443,6 +443,33 @@ std::string replica::query_compact_state() const
...
+        return "CompactionFinish";
+    case CompactionRunning:
+        return "CompactionRunning";
+    case CompactionQueue:
+        return "CompactionQueue";
+    default:
+        dassert(false, "");
+    }
+    return "wrongCompactionStatus";
+}
```
Shuo-Jia, line 459: `enum` convert `string` seem only register like this: 
https://github.com/XiaoMi/rdsn/blob/v2.1.1/include/dsn/dist/replication/replication_enums.h

hycdong: `replication_enums` will add prefix `dsn::replication::` for status, I simply it to avoid adding this prefix.

src/replica/replica.h
```diff
@@ -78,6 +78,14 @@ namespace test {
 class test_checker;
 }
 
+enum manual_compaction_status
+{
+    CompactionFinish = 0,
```
acelyc111, line 83: https://google.github.io/styleguide/cppguide.html#Enumerator_Names

src/replica/replica.cpp
```diff
@@ -443,6 +443,33 @@ std::string replica::query_compact_state() const
...
+    case CompactionFinish:
+        return "CompactionFinish";
+    case CompactionRunning:
+        return "CompactionRunning";
+    case CompactionQueue:
+        return "CompactionQueue";
+    default:
+        dassert(false, "");
+    }
+    return "wrongCompactionStatus";
```
acelyc111, line 458: Use `__builtin_unreachable();` instead

src/replica/replica.cpp
```diff
@@ -443,6 +443,33 @@ std::string replica::query_compact_state() const
...
+}
+
+manual_compaction_status replica::get_compact_status() const
+{
+    std::string compact_state = query_compact_state();
+    if (compact_state.find("recent start at") != std::string::npos) {
```
acelyc111, line 464: Add some comment to explain what does "recent start at" mean.

src/replica/replica.h
acelyc111: > https://google.github.io/styleguide/cppguide.html#Enumerator_Names

It says "Enumerators (for both scoped and unscoped enums) should be named like constants, not like macros. That is, use kEnumName not ENUM_NAME." :(

hycdong: That is my mistake, now it is fixed.

src/replica/replica.h
```diff
@@ -392,8 +400,12 @@ class replica : public serverlet<replica>, public ref_counter, public replica_ba
 
     void update_restore_progress(uint64_t f_size);
 
+    // Used for remote command
```
neverchanje, line 403:
```suggestion
    // Used for remote command
    // TODO: remove this interface and only expose the http interface
```

neverchanje (approved):

acelyc111 (approved):

Shuo-Jia (approved):

PR #703, opened 2020-12-23
fix: block service download file return error when file exists

hycdong:
When download fileA from remote file provider, if local file system has aleady have a file named fileA, server should check if local file is same as remote one, #692 implement it by compare md5 between files, however, hdfs_service don't implemet get_md5, it will always be "". So this pr recover download file before #692, and return ERR_PATH_ALREADY_EXIST when file exists, let the caller function to do validation.


zhangyifan27:
If two files have the same size, they could also be totally different files. I think we shouldn't determine whether the file should be downloaded based on this.

hycdong:
> If two files have the same size, they could also be totally different files. I think we shouldn't determine whether the file should be downloaded based on this.

I have already updated the code as comments show.

src/block_service/block_service_manager.cpp
```diff
@@ -118,6 +118,13 @@ error_code block_service_manager::download_file(const std::string &remote_dir,
                                                 block_filesystem *fs,
                                                 /*out*/ uint64_t &download_file_size)
 {
+    // local file exists
+    const std::string local_file_name = utils::filesystem::path_combine(local_dir, file_name);
+    if (utils::filesystem::file_exists(local_file_name)) {
+        ddebug_f("local file({}) exists", local_file_name);
+        return ERR_DOWNLOADED;
```
neverchanje, line 125:
```suggestion
        return ERR_PATH_ALREADY_EXIST;
```

Using the existing error code is enough, and is more accurate here. Because you didn't check carefully if the file is really downloaded, as you didn't check the md5sum.

neverchanje (dismissed):

src/replica/bulk_load/replica_bulk_loader.cpp
```diff
@@ -414,6 +414,9 @@ error_code replica_bulk_loader::download_sst_files(const std::string &remote_dir
                     remote_dir, local_dir, f_meta.name, fs, f_size);
                 const std::string &file_name =
                     utils::filesystem::path_combine(local_dir, f_meta.name);
+                if (ec == ERR_PATH_ALREADY_EXIST) {
```
zhangyifan27, line 417: Why we don't need to verify the existing file here?

hycdong: Fixed

zhangyifan27 (approved):

acelyc111 (approved):

PR #707, opened 2020-12-24
feat: add update rocksdb scenario interface

hycdong:
This pull request adds update rocksdb scenario envs interface for meta server.

### DIFF
--- a/src/meta/meta_http_service.cpp
+++ b/src/meta/meta_http_service.cpp
@@ -6,6 +6,7 @@
 
 #include <dsn/c/api_layer1.h>
 #include <dsn/cpp/serialization_helper/dsn.layer2_types.h>
+#include <dsn/dist/replication/replica_envs.h>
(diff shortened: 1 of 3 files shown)


neverchanje (approved):

src/meta/meta_http_service.cpp
```diff
@@ -769,6 +769,39 @@ void meta_http_service::start_compaction_handler(const http_request &req, http_r
...
+    if (!ret) {
+        resp.body = "invalid request structure";
+        resp.status_code = http_status_code::bad_request;
+        return;
+    }
+    if (info.app_name.empty()) {
+        resp.body = "app_name should not be empty";
+        resp.status_code = http_status_code::bad_request;
+        return;
+    }
```
acelyc111, line 790: There are many similar code like this, maybe we can use macros to simplify code in future refactor.

acelyc111 (approved):

PR #716, opened 2020-12-29
BREAKING CHANGE(backup): remove 'policy_name' from backup files' path

zhangyifan27:
This patch remove `policy_name` from backup files' path.
If we want to restore apps which is backed up in this new way, we don't need to use `-p` to specify the `policy_name`. If we want to restore old apps, we could also use `-p`.

I have tested this using onebox cluster, backup and restore some apps, all works well.

### DIFF
--- a/src/meta/server_state_restore.cpp
+++ b/src/meta/server_state_restore.cpp
@@ -55,11 +55,12 @@ void server_state::sync_app_from_backup_media(
         return;
     }
 
-    std::string app_metadata = cold_backup::get_app_metadata_file(request.cluster_name,
-                                                                  request.policy_name,
-                                                                  request.app_name,
-                                                                  request.app_id,
-                                                                  request.time_stamp);
+    std::string cluster_root = request.cluster_name;
+    if (!request.policy_name.empty()) {
+        cluster_root += ("/" + request.policy_name);
+    }
+    std::string app_metadata = cold_backup::get_app_metadata_file(
+        cluster_root, request.app_name, request.app_id, request.time_stamp);
 
     error_code err = ERR_OK;
     block_file_ptr file_handle = nullptr;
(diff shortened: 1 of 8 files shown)


src/meta/server_state_restore.cpp
```diff
@@ -55,11 +55,12 @@ void server_state::sync_app_from_backup_media(
...
-                                                                  request.policy_name,
-                                                                  request.app_name,
-                                                                  request.app_id,
-                                                                  request.time_stamp);
+    std::string cluster_root = request.cluster_name;
+    if (!request.policy_name.empty()) {
+        cluster_root += ("/" + request.policy_name);
+    }
+    std::string app_metadata = cold_backup::get_app_metadata_file(
+        cluster_root, request.app_name, request.app_id, request.time_stamp);
```
neverchanje, line 63: what does this mean? do we support restoring from the previous path with policy_name?
can we add a test for it?

src/replica/replica_restore.cpp
```diff
@@ -218,21 +218,23 @@ void replica::clear_restore_useless_files(const std::string &local_chkpt_dir,
...
-    const std::string &backup_root = req.cluster_name;
-    const std::string &policy_name = req.policy_name;
-    const int64_t &backup_id = req.time_stamp;
-    ddebug("%s: retore from policy_name(%s), backup_id(%lld)",
-           name(),
-           req.policy_name.c_str(),
-           req.time_stamp);
+    ddebug_f("{}: start to find valid checkpoint of app {}, backup_id {}",
+             name(),
+             req.app_id,
```
neverchanje, line 223: is app_id redundant? name() has included app_id already.

neverchanje (dismissed):

hycdong (approved):

neverchanje (approved):

PR #722, opened 2021-01-08
refactor: don't resend message if it is rejected by server

levy5307:
1. Don't resend message if it is rejected by server.  because it means that the server is overload if it reject this message. So if we resend the message, it will makes the situation even worse.
2. use cflags to define cluster_name. But we can't use flag validator, because cpp client will not set cluster name. so the validator will always fail for it

### DIFF
--- a/src/common/duplication_common.cpp
+++ b/src/common/duplication_common.cpp
@@ -54,7 +54,8 @@ DSN_DEFINE_string("replication", cluster_name, "", "name of this cluster");
     return it->second;
 }
 
-/*extern*/ const char *get_current_cluster_name() {
+/*extern*/ const char *get_current_cluster_name()
+{
     dassert(strlen(FLAGS_cluster_name) != 0, "cluster_name is not set");
     return FLAGS_cluster_name;
 }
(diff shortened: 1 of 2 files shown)


src/common/duplication_common.cpp
```diff
@@ -52,12 +54,9 @@ namespace replication {
...
 }
 
-/*extern*/ const char *get_current_cluster_name()
-{
-    static const char *cluster_name =
-        dsn_config_get_value_string("replication", "cluster_name", "", "name of this cluster");
-    dassert(strlen(cluster_name) != 0, "cluster_name is not set");
-    return cluster_name;
+/*extern*/ const char *get_current_cluster_name() {
+    dassert(strlen(FLAGS_cluster_name) != 0, "cluster_name is not set");
```
neverchanje, line 58:
```suggestion
    dassert(strlen(FLAGS_cluster_name) != 0, "cluster_name is not set");
```

Use flag validator instead

levy5307: We can't use flag validator, because cpp client will not set cluster name. so the validator will always fail for cpp client

neverchanje (approved):

hycdong (approved):

Shuo-Jia (approved):

PR #724, opened 2021-01-11
fix(backup): fix should_start_backup_unlocked()

zhangyifan27:
Now we start the first backup only when `cur_time == start_time`, then periodly start backup according to `backup_interval_time`.

### DIFF
--- a/src/meta/meta_backup_service.cpp
+++ b/src/meta/meta_backup_service.cpp
@@ -647,55 +647,16 @@ void policy_context::continue_current_backup_unlocked()
 bool policy_context::should_start_backup_unlocked()
 {
     uint64_t now = dsn_now_ms();
-    uint64_t recent_backup_start_time_ms = 0;
-    if (!_backup_history.empty()) {
-        recent_backup_start_time_ms = _backup_history.rbegin()->second.start_time_ms;
-    }
-
-    // the true start time of recent backup have drifted away with the origin start time of
-    // policy,
-    // so we should take the drift into consideration; if user change the start time of the
-    // policy,
-    // we just think the change of start time as drift
-    int32_t hour = 0, min = 0, sec = 0;
-    if (recent_backup_start_time_ms == 0) {
-        //  the first time to backup, just consider the start time
-        ::dsn::utils::time_ms_to_date_time(now, hour, min, sec);
+    if (_backup_history.empty()) {
+        // The first time to backup, just consider the start time
+        int32_t hour = 0, min = 0, sec = 0;
+        dsn::utils::time_ms_to_date_time(now, hour, min, sec);
         return _policy.start_time.should_start_backup(hour, min);
-    } else {
-        uint64_t next_backup_time_ms =
-            recent_backup_start_time_ms + _policy.backup_interval_seconds * 1000;
-        if (_policy.start_time.hour != 24) {
-            // user have specify the time point to start backup, so we should take the the
-            // time-drift into consideration
-
-            // compute the time-drift
-            ::dsn::utils::time_ms_to_date_time(recent_backup_start_time_ms, hour, min, sec);
-            int64_t time_dirft_ms = _policy.start_time.compute_time_drift_ms(hour, min);
-
-            if (time_dirft_ms >= 0) {
-                // hour:min(the true start time) >= policy.start_time :
-                //      1, user move up the start time of policy, such as 20:00 to 2:00, we just
-                //      think this case as time drift
-                //      2, the true start time of backup is delayed, compared the origin start time
-                //      of policy, we should process this case
-                //      3, the true start time of backup is the same with the origin start time of
-                //      policy
-                next_backup_time_ms -= time_dirft_ms;
-            } else {
-                // hour:min(the true start time) < policy.start_time:
-                //      1, user delay the start time of policy, such as 2:00 to 23:00
-                //
-                // these case has already been handled, we do nothing
-            }
-        }
-        if (next_backup_time_ms <= now) {
-            ::dsn::utils::time_ms_to_date_time(now, hour, min, sec);
-            return _policy.start_time.should_start_backup(hour, min);
-        } else {
-            return false;
-        }
     }
+    uint64_t recent_backup_start_time_ms = _backup_history.rbegin()->second.start_time_ms;
+    uint64_t next_backup_time_ms =
+        recent_backup_start_time_ms + _policy.backup_interval_seconds * 1000;
+    return now >= next_backup_time_ms;
 }
 
 void policy_context::issue_new_backup_unlocked()
@@ -716,6 +677,9 @@ void policy_context::issue_new_backup_unlocked()
     }
 
     if (!should_start_backup_unlocked()) {
+        ddebug_f("{}: it's not the right time to start backup now, retry to issue new backup 1 "
+                 "second later.",
+                 _policy.policy_name);
         tasking::enqueue(LPC_DEFAULT_CALLBACK,
                          &_tracker,
                          [this]() {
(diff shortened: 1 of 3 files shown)


src/meta/meta_backup_service.cpp
```diff
@@ -1540,15 +1515,28 @@ void backup_service::modify_backup_policy(configuration_modify_backup_policy_rpc
...
-            have_modify_policy = true;
+        if (!t_start_time.parse_from(request.start_time)) {
+            response.err = ERR_INVALID_PARAMETERS;
+            response.hint_message =
+                fmt::format("invalid start_time: {}, policy {} shouldn't be modified.",
+                            request.start_time,
+                            cur_policy.policy_name);
+            return;
         }
+        ddebug("%s: policy change start_time from (%s) to (%s)",
```
hycdong, line 1534: ddebug_f may be better.

src/meta/meta_backup_service.cpp
```diff
@@ -1540,15 +1515,28 @@ void backup_service::modify_backup_policy(configuration_modify_backup_policy_rpc
     }
 
     if (request.__isset.start_time) {
+        if (context_ptr->has_backup_history() || context_ptr->is_under_backuping()) {
```
hycdong, line 1518: Adding `context_ptr->has_backup_history()` condition, it means if the policy has already generated the backup, its start time can not be modified. Could you please explain why it should not be modified to me?

zhangyifan27: Because `start_time` only indicates the first time to start backup, if we have already started a backup, we shouldn't modify it.

hycdong: For once backup, start_time indicates the first time to start backup, for periodical backup, it should be allowed to be updated, for each time, it indicates this round starting backup.

src/meta/meta_backup_service.h
```diff
@@ -97,58 +92,23 @@ struct backup_start_time
...
-        if (hour == 24) {
-            // erase the restrict of backup_start_time, just for testing
+        if (hour == 24 && minute == 0) {
+            // only for tests
             return true;
         }
-        // NOTICE : if you want more precisely, you can use cur_min to implement
-        // now, we just ignore
-        return (cur_hour == hour);
+        return (cur_hour == hour) && (cur_min == minute);
```
hycdong, line 111: This raises me a question, backup used to have a problem that if backup interval less than one day, it will backup without control, it is caused by the hour check in function `should_start_backup`. Is my understanding correct? Besides, if we check hour and minute here, is that mean backup interval should always be 1 day or  integer multiple of days?

src/meta/meta_backup_service.cpp
zhangyifan27: The start time of subsequent backups is only related to `last_backup_start_time` and `backup_interval_secends`, if users want to modify next backup time, they could modify `backup_interval_secends`.

src/meta/meta_backup_service.h
zhangyifan27: We only check this when we want to start the first time backup.

src/meta/meta_backup_service.cpp
hycdong: Yes, you almost persuade me. Let me take an example to explain, there were a policy whose start time is 1:00, interval time is 1 day, if we would like update it to backup everyday 2:00. In this case, we wouldn't update interval time, if we don't provide updating start_time, we can only create another policy or update interval twice.

src/meta/meta_backup_service.h
```diff
@@ -79,15 +79,10 @@ struct backup_info
...
-// Attention: backup_start_time == 24:00 is represent no limit for start_time, 24:00 is mainly saved
-// for testing
-//
-// current, we don't support accurating to minute, only support accurating to hour, so
-// we just set minute to 0
 struct backup_start_time
 {
-    int32_t hour;   // [0 ~24)
-    int32_t minute; // [0 ~ 60)
+    int32_t hour;
```
levy5307, line 84: It's better to change the type of hour and minute to `uint8_t`

zhangyifan27: `dsn::utils::time_ms_to_date_time` use `int32_t` for hour, min, sec, so we also use `int32_t` here.
Maybe we should consider use `uint8_t` to replace `int32_t` in `dsn::utils::time_ms_to_date_time`, if not , even if we changed  interger type here, type conversions would still be needed.

src/meta/meta_backup_service.h
```diff
@@ -97,58 +92,23 @@ struct backup_start_time
...
+        return (hour >= 0 && hour < 24 && minute >= 0 && minute < 60);
     }
 
-    // judge whether we should start backup base current time
     bool should_start_backup(int32_t cur_hour, int32_t cur_min)
     {
-        if (hour == 24) {
-            // erase the restrict of backup_start_time, just for testing
+        if (hour == 24 && minute == 0) {
+            // only for tests
```
neverchanje, line 108: this is too tricky. I don't want any test code appears in such way.

Try this: https://github.com/XiaoMi/rdsn/blob/master/include/dsn/utility/fail_point.h

```
FAIL_POINT_INJECT_F(should_start_backup, []() -> bool {
  return true;
});
```

When you in test calling `fail::cfg("should_start_backup", return())`, and this func will return true.

src/meta/meta_backup_service.cpp
```diff
@@ -1254,11 +1221,20 @@ void backup_service::add_backup_policy(dsn::message_ex *msg)
...
     dassert(policy_context_ptr != nullptr, "invalid policy_context");
     policy p;
+    if (!p.start_time.parse_from(request.start_time)) {
+        derror_f("invalid start time: {}, policy {} shouldn't be added.",
+                 request.start_time,
+                 request.policy_name);
+        response.err = ERR_INVALID_PARAMETERS;
+        response.hint_message = "invalid start time: " + request.start_time;
+        _meta_svc->reply_data(msg, response);
+        msg->release_ref();
```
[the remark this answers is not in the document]
zhangyifan27: Agree. I'll use rpc_holder to replace message_ex later.

src/meta/meta_backup_service.h
zhangyifan27: Actually we didn't call `should_start_backup` directly in tests, we called `issue_new_backup_unlocked` and expected starting a new backup in `policy_context_test.test_app_dropped_during_backup`.

src/meta/meta_backup_service.cpp
```diff
@@ -716,17 +676,17 @@ void policy_context::issue_new_backup_unlocked()
     }
 
     if (!should_start_backup_unlocked()) {
+        dinfo_f("{}: it's not the right time to start backup now, retry to issue new backup 1 "
```
neverchanje, line 679: Better to remove the logs you don't need.

src/meta/meta_backup_service.cpp
```diff
@@ -835,6 +791,12 @@ std::vector<backup_info> policy_context::get_backup_infos(int cnt)
     return ret;
 }
 
+bool policy_context::has_backup_history()
+{
+    zauto_lock l(_lock);
+    return _backup_history.size() > 0;
```
acelyc111, line 797:
```suggestion
    return !_backup_history.empty();
```

src/meta/meta_backup_service.cpp
acelyc111: Let‘s imagine how would user use this feature, maybe it's almost like how would you set you clock, there is no such a configurable "interval" concept, the interval should always be fixed value like day(or week, month, etc).
If user want to backup multiple times in one day, she should set multiple policies. On the other hand, I believe backup multiple times a day isn't a regular requirement, we don't need to make our design and code too complex to satisfy it.

zhangyifan27: I don't think the design and code is complex. `start_time` is the first time to start a backup of a policy, `backup_interval_seconds` is interval between two backups, which is quite easy to understand. If users want to modify start_time, they should create another policy.

If we all agree that we should keep `backup_interval_seconds`, then any interval could be set, we shouldn't only support the fixed interval(day/week/month).

neverchanje:
@acelyc111 The new implementation is actually making the kernel design simpler than before, considering the deleted code that handles clock drift. We don't have to avoid the two backups being too close.

From the users perspective, this design is primitive and prone to misuse. I prefer this problem to be solved in Backup Manager. You can set up rules or limitations in Backup Manager.

PR #725, opened 2021-01-12
fix(split): message header should not be empty struct

hycdong:
Function `copy_message_no_reply` will be used during partition split when child copy parent's mutation, its header should not be an empty structure.

### DIFF
--- a/src/runtime/rpc/rpc_message.cpp
+++ b/src/runtime/rpc/rpc_message.cpp
@@ -172,7 +172,7 @@ message_ex *message_ex::copy_message_no_reply(const message_ex &old_msg)
         static_cast<char *>(dsn::tls_trans_malloc(sizeof(message_header))),
         [](char *c) { dsn::tls_trans_free(c); });
     msg->header = reinterpret_cast<message_header *>(header_holder.get());
-    msg->header = {}; // initialize to empty struct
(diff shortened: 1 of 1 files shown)


src/runtime/rpc/rpc_message.cpp
```diff
@@ -172,7 +172,7 @@ message_ex *message_ex::copy_message_no_reply(const message_ex &old_msg)
         static_cast<char *>(dsn::tls_trans_malloc(sizeof(message_header))),
         [](char *c) { dsn::tls_trans_free(c); });
     msg->header = reinterpret_cast<message_header *>(header_holder.get());
-    msg->header = {}; // initialize to empty struct
+    memset(static_cast<void *>(msg->header), 0, sizeof(message_header));
```
neverchanje, line 175: I don't see the difference between the two statements. Please write a unit-test for `copy_message_no_reply`

neverchanje (approved):

src/runtime/rpc/rpc_message.cpp
hycdong: As we discuss offline, if header is an empty strtuct, it will lead problem in Line195, which is `msg->header->body_length = msg->buffers[1].length();`.

acelyc111 (approved):

PR #727, opened 2021-01-14
feat(split): add child copy mutation synchronously

hycdong:
### What this pull request solves
During partition split, parent partition should send mutation to child partition, child will copy this mutation and prepare it. This pull request is about child partition copying mutations synchronously.(async-learn: #676)
When all child partitions caught up its parent, primary parent will set `sync_send_write_request=true`(#390), then each mutation parent received should wait for child prepare it succeed until this value is false. 

### Difference between normal 2pc and 2pc in sync-learn
- In normal write 2pc:
    - secondary: reply ack to primary when prepare succeed
    - primary: commit mutation when prepare succeed and receive all secondaries ack
- In sync-learn 2pc:
    - secondary: reply ack to primary when both its child partition and itself prepare succeed
    - primary: commit mutation when both its child partition and itself prepare succeed and receive all secondaries ack

### Other details
- For those mutations should be copied sychronously, its `_is_sync_to_child = true`, will be set in `init_prepare` and `on_prepare`.
- For child partition, after it prepared mutation, it would append it into shared log, sent ack to parent and appended it into private log
- Add `_wait_child` to avoid secondary parent reply to primary parent without waiting for its child prepare succeed, also avoid primary parent commit without waiting for its child prepare succeed
- Add `_is_error_acked` to avoid reply twice when secondary partition prepare failed


src/replica/mutation.h
```diff
@@ -155,7 +160,15 @@ class mutation : public ref_counter
         {
             unsigned int _not_logged : 1;
             unsigned int _left_secondary_ack_count : 15;
-            unsigned int _left_potential_secondary_ack_count : 16;
+            unsigned int _left_potential_secondary_ack_count : 14;
+            // Used for partition split
+            // _wait_child = 1 : child prepare mutation synchronously, its parent should wait for
+            // child ack
+            unsigned int _wait_child : 1;
```
neverchanje, line 167: You can use bool here, I've tested on c++98,c++11,c++14, with `-wall` flag on.

```
// Example program
#include <iostream>
#include <string>

struct CompactStruct {
    unsigned int _not_logged : 1;
    unsigned int _left_secondary_ack_count : 15;
    unsigned int _left_potential_secondary_ack_count : 16;
    // Used for partition split
    // _wait_child = 1 : child prepare mutation synchronously, its parent should wait for
    // child ack
    bool _wait_child : 1;
    // Used for partition split
    // when prepare failed when child prepare mutation synchronously, secondary may try to
    // ack to primary twice, we use _is_error_acked to restrict only ack once
    bool _is_error_acked : 1;
};

int main()
{
    CompactStruct t;
    t._is_error_acked = 1;
    std::cout << t._wait_child << std::endl;
    return 0;
}
```

hycdong: Done

neverchanje (approved):

neverchanje (approved):

src/replica/replica_2pc.cpp
```diff
@@ -349,6 +355,9 @@ void replica::on_prepare(dsn::message_ex *request)
         rpc_read_stream reader(request);
         unmarshall(reader, rconfig, DSF_THRIFT_BINARY);
         mu = mutation::read_from(reader, request);
+        mu->set_is_sync_to_child(rconfig.split_sync_to_child);
+        // set split_sync_to_child to false immediately
```
levy5307, line 359: the comment here is useless

src/replica/mutation.h
```diff
@@ -155,7 +160,15 @@ class mutation : public ref_counter
...
             unsigned int _left_secondary_ack_count : 15;
-            unsigned int _left_potential_secondary_ack_count : 16;
+            unsigned int _left_potential_secondary_ack_count : 14;
+            // Used for partition split
+            // _wait_child = true : child prepare mutation synchronously, its parent should wait for
+            // child ack
+            bool _wait_child : 1;
+            // Used for partition split
+            // when prepare failed when child prepare mutation synchronously, secondary may try to
+            // ack to primary twice, we use _is_error_acked to restrict only ack once
```
levy5307, line 170: Why secondary may try to ack to primary twice when prepare failed? could you explain here?

levy5307 (approved):

neverchanje (approved):

PR #728, opened 2021-01-15
feat: add throttling for read size and read qps

levy5307:
add throttling for read size and read qps


src/replica/replica_throttle.cpp
```diff
@@ -14,40 +14,49 @@
...
-                               int32_t request_units)
+#define THROTTLE_REQUEST(op_type, throttling_type, request, request_units)                         \
+    do {                                                                                           \
+        int64_t delay_ms = 0;                                                                      \
+        auto type = _##op_type##_##throttling_type##_throttling_controller.control(                \
+            request->header->client.timeout_ms, request_units, delay_ms);                          \
+        if (type != throttling_controller::PASS) {                                                 \
+            if (type == throttling_controller::DELAY) {                                            \
+                tasking::enqueue(                                                                  \
+                    LPC_THROTTLING_DELAY,                                                          \
```
neverchanje, line 25:
```suggestion
                    LPC_THROTTLING_DELAY,                                                          \
```

`LPC_WRITE_THROTTLING_DELAY` was previously in the replication thread pool. If read is also throttled, it should be retried in `THREAD_POOL_LOCAL_APP` thread pool.

include/dsn/dist/replication/replication.codes.h

```diff
// THREAD_POOL_LOCAL_APP
#define CURRENT_THREAD_POOL THREAD_POOL_LOCAL_APP
MAKE_EVENT_CODE(LPC_WRITE, TASK_PRIORITY_COMMON)
+MAKE_EVENT_CODE(LPC_READ_THROTTLING, TASK_PRIORITY_COMMON)
#undef CURRENT_THREAD_POOL
```

src/replica/replica_throttle.cpp
```diff
@@ -14,40 +14,49 @@
...
+                }                                                                                  \
+                _counter_recent_##op_type##_throttling_reject_count->increment();                  \
+            }                                                                                      \
+            return true;                                                                           \
+        }                                                                                          \
+    } while (0)
+
+bool replica::throttle_write_request(message_ex *request)
 {
-    if (!controller.enabled()) {
```
hycdong, line 21: I recommend remaining check controller enabled in function `throttle_write_request` and `throttle_read_request` to avoid calling function `control`, especially for read reuqest, most of read requests don't have throttle control.

src/replica/replica_throttle.cpp
```diff
@@ -14,40 +14,49 @@
...
+                if (delay_ms > 0) {                                                                \
+                    tasking::enqueue(LPC_##op_type##_THROTTLING_DELAY,                             \
+                                     &_tracker,                                                    \
+                                     [ this, req = message_ptr(request) ]() {                      \
+                                         response_client_##op_type(req, ERR_BUSY);                 \
+                                     },                                                            \
+                                     get_gpid().thread_hash(),                                     \
+                                     std::chrono::milliseconds(delay_ms));                         \
+                } else {                                                                           \
+                    response_client_##op_type(request, ERR_BUSY);                                  \
```
hycdong, line 41: I think you can update throttling error code in this pull request.

src/replica/replica_throttle.cpp
levy5307: I didn't remove it.  I just move this check into `src/utils/throttling_controller.cpp`.

src/replica/replica_throttle.cpp
levy5307: I think we can update it someday later. Modify at the same time as java client

src/replica/replica_throttle.cpp
hycdong: I knew that check is in function `control`, my suggestion is not moving the check in function `control`, remaining it in function `throttle_write_request` and `throttle_read_request` to avoid calling another function, especially for read requests.

levy5307: There are two contrllers for read and write operations, which are size throttling and qps throttling. we should call enable() function for each of them.
And I think It's better to move enable() into throttling_controller to simplify the usage of it.

hycdong: I got your point. How about moving this check in macro THROTTLE_REQUEST? It can avoid calling another function and simplify for usage.

levy5307: I don't think it's a good idea. We shouldn't make the `enable` of `throttling_controller` visiable. In other words,  for the user of `throttling_controller`,  it is not necessary to know whether `enable` is set or not. He only needs to call the function `control` and get the result which means whether it is passed or not.

levy5307: I don't think it's a good idea. We shouldn't make the `enable` of `throttling_controller` visiable. In other words,  for the user of `throttling_controller`,  it is not necessary to know whether `enable` is set or not. He only needs to call the function `control` and get the result which means whether it is passed or not.

hycdong (approved):

neverchanje (approved):

PR #733, opened 2021-01-20
fix(asan):  heap-use-after-free caused by replica_test_base::create_test_mutation

levy5307:
In function `create_test_mutation`, it uses `const std::string &data`as it's second param. And gets the pointer of data, reserving it. But the caller of `create_test_mutation` may release variable `data`. So the `mu` will reserves a wild pointer in this case. and the use of the reserved pointers will produce `heap-use-after-free` error.

### DIFF
--- a/src/replica/test/replica_test_base.h
+++ b/src/replica/test/replica_test_base.h
@@ -68,7 +68,9 @@ struct replica_test_base : replica_stub_test_base
         mu->data.updates.emplace_back(mutation_update());
         mu->data.updates.back().code =
             RPC_COLD_BACKUP; // whatever code it is, but never be WRITE_EMPTY
-        mu->data.updates.back().data.assign(data.data(), 0, data.length());
+        std::shared_ptr<char> ptr(dsn::utils::make_shared_array<char>(data.length()));
+        memcpy(ptr.get(), data.data(), data.length());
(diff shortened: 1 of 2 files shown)


hycdong (approved):

src/replica/test/replica_test_base.h
```diff
@@ -68,7 +68,9 @@ struct replica_test_base : replica_stub_test_base
         mu->data.updates.emplace_back(mutation_update());
         mu->data.updates.back().code =
             RPC_COLD_BACKUP; // whatever code it is, but never be WRITE_EMPTY
-        mu->data.updates.back().data.assign(data.data(), 0, data.length());
+        std::shared_ptr<char> ptr(dsn::utils::make_shared_array<char>(data.length()));
+        memcpy(ptr.get(), data.data(), data.length());
+        mu->data.updates.back().data.assign(ptr, 0, data.length());
```
neverchanje, line 73:
```suggestion
        mu->data.updates.back().data = blob::create_from_bytes(std::string(data));
```

https://github.com/XiaoMi/rdsn/blob/master/include/dsn/utility/blob.h#L73

neverchanje:
It would be better if you could paste the asan error on every PR that fixes the problem.

neverchanje (approved):

hycdong (approved):

PR #734, opened 2021-01-21
fix(command_manager): avoid using std::remove to erase elements from std::vector

zhangyifan27:
On account of `std::remove` could not actually erase elements from containers, 

> The removal is done by replacing the elements that compare equal to val by the next element that does not, ...  [1]

we may delete `command_instance *` repeatly when destroy `command_manager`.
```
void command_manager::deregister_command(dsn_handle_t handle)
{
    auto c = reinterpret_cast<command_instance *>(handle);
    dassert(c != nullptr, "cannot deregister a null handle");
    utils::auto_write_lock l(_lock);
    for (const std::string &cmd : c->commands) {
        ddebug("unregister command: %s", cmd.c_str());
        _handlers.erase(cmd);
    }
    std::remove(_commands.begin(), _commands.end(), c);	
    delete c;	
}

command_manager::~command_manager()
{	
    for (command_instance *c : _commands) {	
        delete c;	
    }	
}
```
The correct way to remove `command_instance` from the vector should be:
```
    auto iter = std::remove(_commands.begin(), _commands.end(), c);
    _comands.erase(iter, _commands.end());
    delete c;	
``` 
This patch remove `std::vector<command_instance *> _commands` from class `command_manager` because we already have `std::map<std::string, command_instance *> _handlers`, and also improve the use of `register_command` and `deregister_command`.

[1] https://www.cplusplus.com/reference/algorithm/remove/

### DIFF
--- a/src/runtime/service_engine.h
+++ b/src/runtime/service_engine.h
@@ -128,6 +128,9 @@ private:
     service_spec _spec;
     env_provider *_env;
 
+    dsn_handle_t _get_runtime_info_cmd;
+    dsn_handle_t _get_queue_info_cmd;
+
     // <port, servicenode>
     typedef std::map<int, service_node *>
         node_engines_by_port; // multiple ports may share the same node
--- a/src/utils/command_manager.cpp
+++ b/src/utils/command_manager.cpp
@@ -52,7 +52,6 @@ dsn_handle_t command_manager::register_command(const std::vector<std::string> &c
     c->help_long = help_long;
     c->help_short = help_one_line;
     c->handler = handler;
-    _commands.push_back(c);
 
     for (const std::string &cmd : commands) {
         if (!cmd.empty()) {
@@ -71,8 +70,6 @@ void command_manager::deregister_command(dsn_handle_t handle)
         ddebug("unregister command: %s", cmd.c_str());
         _handlers.erase(cmd);
     }
-    std::remove(_commands.begin(), _commands.end(), c);
-    delete c;
 }
 
 bool command_manager::run_command(const std::string &cmd,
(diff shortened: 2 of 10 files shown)


src/utils/command_manager.cpp
```diff
@@ -106,8 +103,8 @@ command_manager::command_manager()
 
                          if (args.size() == 0) {
                              utils::auto_read_lock l(_lock);
-                             for (auto c : this->_commands) {
-                                 ss << c->help_short << std::endl;
+                             for (auto c : this->_handlers) {
```
levy5307, line 106: const auto &handler

hycdong (approved):

src/runtime/service_engine.h
```diff
@@ -128,6 +128,9 @@ class service_engine : public utils::singleton<service_engine>
     service_spec _spec;
     env_provider *_env;
 
+    dsn_handle_t _get_runtime_info_cmd;
```
neverchanje, line 131: I would recommend wrapping `command_instance*` into a RAII class, so that we do not need to UNREGISTER_VALID_HANDLER manually. If we don't refactor in this PR, I will take this later.

src/utils/command_manager.cpp
```diff
@@ -52,7 +52,6 @@ dsn_handle_t command_manager::register_command(const std::vector<std::string> &c
     c->help_long = help_long;
     c->help_short = help_one_line;
     c->handler = handler;
-    _commands.push_back(c);
 
     for (const std::string &cmd : commands) {
         if (!cmd.empty()) {
```
neverchanje, line 57:
```
if (!cmd.empty()) {
   _handlers[cmd] = c;
} else {
  delete c;
}
```

There's a potential memleak.

acelyc111 (approved):

hycdong (approved):

PR #742, opened 2021-02-01
feat(split): supplement two cases after implementing cancel split

hycdong:
In pervious pull requests, cancel partition split is implemented, after adding those functions, we need to handle some cases specially.
1. In function `register_child_on_meta`
  - when all replicas in group finish split, primary parent will send rpc to register child on meta server, if split is canceled, this register rpc should be failed
2. In function `on_config_sync`
  - when partition split is canceled, the useless child replica is still in replica server memory, it will be included in on_config_sync rpc, meta server should not assert in this case, should let replica server gc this replica.

### DIFF
--- a/src/meta/server_state.cpp
+++ b/src/meta/server_state.cpp
@@ -834,13 +834,24 @@ void server_state::on_config_sync(configuration_query_by_node_rpc rpc)
                       rep.pid.get_partition_index());
                 std::shared_ptr<app_state> app = get_app(rep.pid.get_app_id());
                 if (app == nullptr || rep.pid.get_partition_index() >= app->partition_count) {
-                    // app is not recognized or partition is not recognized
-                    dassert(false,
-                            "gpid(%d.%d) on node(%s) is not exist on meta server, administrator "
-                            "should check consistency of meta data",
-                            rep.pid.get_app_id(),
-                            rep.pid.get_partition_index(),
-                            request.node.to_string());
+                    // This app has garbage partition after cancel split, the canceled child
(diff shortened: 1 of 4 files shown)


levy5307 (approved):

Shuo-Jia (approved):

src/meta/server_state.cpp
```diff
@@ -834,13 +834,24 @@ void server_state::on_config_sync(configuration_query_by_node_rpc rpc)
...
+                                request.node.to_string(),
+                                rep.pid);
+                    } else {
+                        // app is not recognized or partition is not recognized
+                        dassert(false,
+                                "gpid({}) on node({}) is not exist on meta server, administrator "
+                                "should check consistency of meta data",
+                                rep.pid,
+                                request.node.to_string());
+                    }
```
neverchanje, line 854: There are too many embedded levels, you should have sensed a very bad smell in this code. Try to make this big if into a function.

```cpp
// check_partition_index_validity returns false if the replica is should be GCed.
bool check_partition_index_validity(const shared_ptr<app_state>& app, gpid request_pid) {
  bool should_assert=false;
  if (app == nullptr) {
     should_assert=true;
  } else if (rep.status != partition_status::PS_ERROR && rep.pid.get_partition_index() >= app->partition_count * 2) {
    should_assert=true;
 }
  if (should_assert) {
    dassert(false);
  }
}

if (!check_partition_index_validity( req.pid)) {
  response.gc_replicas.push_back(rep);
}
```

PR #394, opened 2020-02-12
feat(split): add partition_version

hycdong:
`partition_version` is used to guarantee data consistence during partition split.
In normal cases, `partition_version = partition_count-1`, when this replica reject read and write request, `partition_version = -1`.
This pull request adds the definition of `partition_version`.

PR #747, opened 2021-02-02
feat(split): replica add validate_partition_hash

hycdong:
As pr #745 shows, we adds an app_env called `SPLIT_VALIDATE_PARTITION_HASH`, when partition split started, it will be set as true, this pr adds how replica handle it, including:
1. In function `update_app_envs_internal`, update replica's `_validate_partition_hash` value
2. In function `on_client_read` and `on_client_write`, check if reject all requests
As previous pr #391 shows, when parent partition is registering child partition, requests should be rejected during this stage (partition_version = -1)
3. In function `on_client_read` and `on_client_write`, check if read from correct partition, write into correct partition
As previous pr #394 shows, we adds `partition_version` to do this check, partition_version valid value is partition_count - 1. Let me take an example to explain the check:
> Before split:
> partition_count = 8, partition_version = 7, parent_partition = 1
> case1. write request partition_hash = 1, target_partition_index = (1 & 7) = 1
> case2. write request partition_hash = 9, target_partition_index = (9 & 7) = 1

When partition split finishs, the request whose partition_hash is 9 should not sent to parent partition, its should be sent to write to the child partition.

> After split:
> partition_count = 16, partition_version = 15, parent_partition = 1
> case1. write request partition_hash = 1, target_partition_index = (1 & 15) = 1
> case2. write request partition_hash = 9, target_partition_index = (9 & 15) = 9 -> should sent to child

When client receives error ERR_PARENT_PARTITION_MISUSED, it should trigger query_meta to let client learned the newest partition_count, this will be added in client.


src/replica/replica.cpp
```diff
@@ -176,6 +176,19 @@ void replica::on_client_read(dsn::message_ex *request, bool ignore_throttling)
         response_client_read(request, ERR_ACL_DENY);
     }
 
+    // validate partition_hash
```
levy5307, line 179: I think you can implement a func, and move these code into it

hycdong: I meant to add another function to do it, but it can only be implemented as macro, read request should call `on_client_read`, write request should call `on_client_write`. Besides, I think it clear to leave return error in `on_client_read` and `on_client_write`.

acelyc111 (dismissed):

src/replica/replica.cpp
```diff
@@ -176,6 +176,19 @@ void replica::on_client_read(dsn::message_ex *request, bool ignore_throttling)
...
+        if (_split_mgr->should_reject_request()) {
+            response_client_read(request, ERR_SPLITTING);
+            return;
+        }
+        if (!_split_mgr->check_partition_hash(
+                ((dsn::message_ex *)request)->header->client.partition_hash, "read")) {
+            response_client_read(request, ERR_PARENT_PARTITION_MISUSED);
+            return;
+        }
+    }
```
neverchanje, line 190:
```cpp
bool replica::reject_read_if_splitting(message_ex* request) {
    if (_validate_partition_hash) {
        if (_split_mgr->should_reject_request()) {
            response_client_read(request, ERR_SPLITTING);
            return true;
        }
        if (!_split_mgr->check_partition_hash(
                ((dsn::message_ex *)request)->header->client.partition_hash, "read")) {
            response_client_read(request, ERR_PARENT_PARTITION_MISUSED);
            return true;
        }
    }
    return false;
}
```

Separating the new code in a func is more readable. And of course it doesn't have to be a macro. Why you think that?

```
if (reject_read_if_splitting()) {
  return;
}
```

src/replica/replica_2pc.cpp
```diff
@@ -77,6 +77,19 @@ void replica::on_client_write(dsn::message_ex *request, bool ignore_throttling)
         return;
     }
 
+    // validate partition_hash
```
neverchanje, line 80: `reject_write_if_splitting`. If you want to eliminate the repeated code, you can use macro too:

```cpp
#define REJECT_REQUEST_IF_SPLITTING(op_type) \
if (_validate_partition_hash) { \
        if (_split_mgr->should_reject_request()) { \
            response_client##op_type##(request, ERR_SPLITTING); \
            return;
        }
        if (!_split_mgr->check_partition_hash( \
                ((dsn::message_ex *)request)->header->client.partition_hash, #op_type)) {
            response_client_write(request, ERR_PARENT_PARTITION_MISUSED);
            return;
        }
}  
```

levy5307: +1.

hycdong: As I comment to liwei, I think it clear to leave return error in `on_client_read` and `on_client_write`, and the repeated code is not too long.

src/replica/split/replica_split_manager.h
```diff
@@ -170,6 +170,22 @@ class replica_split_manager : replica_base
     ballot get_ballot() const { return _replica->get_ballot(); }
     decree last_committed_decree() const { return _replica->last_committed_decree(); }
     task_tracker *tracker() { return _replica->tracker(); }
+    bool should_reject_request() const { return get_partition_version() == -1; }
+    bool check_partition_hash(const uint64_t &partition_hash, const std::string &op) const
+    {
+        auto target_pidx = get_partition_version() & partition_hash;
+        if (target_pidx != get_gpid().get_partition_index()) {
```
levy5307, line 177: `dsn_unlikely`

levy5307 (approved):

acelyc111 (approved):

PR #752, opened 2021-02-05
fix: make disk_engine destructed after service_engine destructed

zhangyifan27:
When exiting the process, we need to close files using `disk_engine`, `disk_engine` should be destructed after `service_engine` destructed.

### DIFF
--- a/src/aio/file_io.cpp
+++ b/src/aio/file_io.cpp
@@ -42,10 +42,13 @@ namespace file {
 
 /*extern*/ error_code close(disk_file *file)
 {
+    error_code result = ERR_INVALID_HANDLE;
     if (file != nullptr) {
-        return disk_engine::provider().close(file->native_handle());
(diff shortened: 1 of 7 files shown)


src/aio/file_io.cpp
```diff
@@ -43,13 +42,10 @@ namespace file {
...
 /*extern*/ error_code close(disk_file *file)
 {
-    if (nullptr != file) {
-        auto ret = disk_engine::provider().close(file->native_handle());
-        delete file;
-        return ret;
-    } else {
-        return ERR_INVALID_HANDLE;
+    if (file != nullptr) {
+        return disk_engine::provider().close(file->native_handle());
```
neverchanje, line 46: Don't we need to delete the pointer?

levy5307 (approved):

acelyc111 (approved):

PR #762, opened 2021-02-22
feat(split): add query_child_state rpc

hycdong:
When partition split starts, meta server will transfer split_status and partition_count through `on_config_sync` to replica server.
When `meta partition_count = replica paritition_count * 2`, primary parent replica will do different things according to meta_split_status:
- `meta_split_status = SPLITTING`: start split
- `meta_split_status = PAUSING` or `meta_split_status = CANCELING`: pause or cancel split
- `meta_split_status = PAUSING`: split has been paused
- **`meta_split_status = NOT_SPLIT`: this pr solved**

This pull request is solve lastest case (meta partition_count = replica paritition_count * 2 && meta_split_status = NOT_SPLIT), there will be two cases:
1. when primary replica register child succeed, but replica server crashed, meta server will consider this parent partition not splitting,  in this case, child has been registered on meta server
2. when this parent partition is canceled, but other partitions is still canceling, in this case, child partition ballot is invalid_ballot

As above cases show, primary parent partition can not distinguish different cases,  it should send a query request to meta server called `query_child_state`
- If child partition is ready, meta server return ERR_OK to primary, primary will make child active and update parent group partition_count
- If is not splitting, meta server return ERR_INVALID_STATE to primary, primary will ignore this case, it split has already canceled, just waiting for other paritions.


src/meta/meta_service.cpp
```diff
@@ -1078,6 +1080,23 @@ void meta_service::on_notify_stop_split(notify_stop_split_rpc rpc)
...
+
+    if (_split_svc == nullptr) {
+        derror_f("meta doesn't support partition split");
+        rpc.response().err = ERR_SERVICE_NOT_ACTIVE;
+        return;
+    }
+    tasking::enqueue(LPC_META_STATE_NORMAL,
+                     tracker(),
+                     [this, rpc]() { _split_svc->query_child_state(std::move(rpc)); },
+                     server_state::sStateHash);
```
neverchanje, line 1097:
```suggestion
    _split_svc->query_child_state(std::move(rpc));
```
You don't have to enqueue because this is a read operation.

src/meta/meta_split_service.cpp
```diff
@@ -551,5 +551,40 @@ void meta_split_service::do_cancel_partition_split(std::shared_ptr<app_state> ap
...
 }
 
+void meta_split_service::query_child_state(query_child_state_rpc rpc)
+{
+    const auto &request = rpc.request();
+    const auto &app_name = request.app_name;
+    const auto &parent_pid = request.pid;
+    auto &response = rpc.response();
+
+    zauto_write_lock l(app_lock());
```
neverchanje, line 561: read lock is enough.

src/replica/split/replica_split_manager.cpp
```diff
@@ -1434,5 +1444,65 @@ void replica_split_manager::parent_send_notify_stop_request(
...
+    ddebug_replica("query child partition succeed, child partition[{}] has already been ready",
+                   response.child_config.pid);
+    // make child partition active
+    _stub->split_replica_exec(LPC_PARTITION_SPLIT,
+                              response.child_config.pid,
+                              std::bind(&replica_split_manager::child_partition_active,
+                                        std::placeholders::_1,
+                                        response.child_config));
+    update_local_partition_count(response.partition_count);
+    _replica->_primary_states.cleanup_split_states();
```
levy5307, line 1501: What about move `_replica->_primary_states.cleanup_split_states();` into `parent_cleanup_split_context();`

hycdong: It is a good recommend~ Actually, I will refactor it after I merge all function pull requests.

neverchanje (approved):

Shuo-Jia (approved):

PR #771, opened 2021-03-02
refactor: simplify nfs client and nfs server

neverchanje:
The file nfs_client.h, nfs_server.h are auto-generated previously, by a variant of thrift compiler that was made by microsoft/rdsn. 
Here I remove the files.


src/nfs/nfs_client_impl.h
```diff
@@ -24,32 +24,54 @@
...
+template <typename TCallback>
+task_ptr async_nfs_copy(const copy_request &request,
+                        TCallback &&callback,
+                        std::chrono::milliseconds timeout,
+                        rpc_address server_addr)
+{
+    return rpc::call(
+        server_addr, RPC_NFS_COPY, request, nullptr, std::forward<TCallback>(callback), timeout);
+}
+
```
Shuo-Jia, line 73: The APIs seem be not same with the origin(https://github.com/XiaoMi/rdsn/pull/771/files#diff-1eea91de75fadd0a10bf2538d55da66d9788516188eca733de36133898d7c8e0L107),  why not stay the same.

src/nfs/nfs_client_impl.h
```diff
@@ -24,32 +24,54 @@
...
+#include <iostream>
 #include <dsn/tool-api/task_tracker.h>
 #include <dsn/tool-api/zlocks.h>
 #include <dsn/perf_counter/perf_counter_wrapper.h>
 #include <dsn/dist/nfs_node.h>
 #include <dsn/utility/defer.h>
 #include <dsn/utility/TokenBucket.h>
 #include <dsn/utility/flags.h>
-#include "nfs_client.h"
+#include <dsn/utility/optional.h>
```
hycdong, line 39: It seems that `optional.h` is not used.

levy5307 (dismissed):

src/nfs/nfs_client_impl.h
neverchanje: Because it's now a free function not in any class. So to make it more clear, I prefixed the name with "async_nfs_" to ensure that the reader knows: 1. this api is asynchronous 2. this api belongs to the nfs module.

hycdong (approved):

Shuo-Jia (approved):

PR #772, opened 2021-03-03
feat(one-time backup): part-1 meta server handle rpc and init backup

zhangyifan27:
Prior to this patch, we could only backup tables by creating a `policy`. But there are some issues in `policy`:
- we don't know when the meta server would start a backup, even though we have set the `start_time`.
- we couldn't disable a policy in some cases.
- if the `interval time` is less than 24 hours, the behavior of `policy_context` would be uncontrollable.

In this patch, meta server could start backup tables immediately after received a backup rpc.

TODO:
- meta server backup app metadata.
- meta server send backup request to replica servers and handle response.
- add query_backup_status rpc.
- add tests.


src/meta/meta_backup_service.h
```diff
@@ -375,10 +422,11 @@ class backup_service
     meta_service *_meta_svc;
     server_state *_state;
 
-    // _lock is only used to lock _policy_states
+    // lock _policy_states and _backup_states.
     zlock _lock;
     std::map<std::string, std::shared_ptr<policy_context>>
         _policy_states; // policy_name -> policy_context
+    std::unordered_map<int64_t, std::shared_ptr<backup_engine>> _backup_states;
```
neverchanje, line 429: What is the map key (int64_t)?

src/common/backup.thrift
```diff
@@ -154,3 +154,15 @@ struct configuration_query_restore_response
     2:list<dsn.error_code>  restore_status;
     3:list<i32>             restore_progress;
 }
+
+struct start_backup_apps_request
+{
+    1:string        backup_provider_type;
+    2:set<i32>      app_ids;
```
neverchanje, line 161: I prefer each RPC handling only one app. This can make the implementation simpler.

src/meta/meta_backup_service.h
```diff
@@ -311,6 +312,51 @@ mock_private :
...
+    dist::block_service::block_filesystem *_block_service;
+    std::string _provider_type;
+    dsn::task_tracker _tracker;
+
+    // lock _cur_backup and _backup_status.
+    dsn::zlock _lock;
+    backup_info _cur_backup;
+    // app_id --> partition_id, backup_status
+    std::map<int32_t, std::map<int32_t, backup_status>> _backup_status;
+};
```
neverchanje, line 358: I prefer to move the lines into a new file, maybe backup_engine.h. It would be easier for readers to locate the class via file name.

src/meta/meta_backup_service.cpp
```diff
@@ -984,6 +984,285 @@ void policy_context::sync_remove_backup_info(const backup_info &info, dsn::task_
...
+}
+
+backup_engine::~backup_engine() { _tracker.cancel_outstanding_tasks(); }
+
+error_code backup_engine::init_backups(const std::set<int32_t> &app_ids)
+{
+    zauto_lock lock(_lock);
+    _cur_backup.app_names.clear();
+    _backup_status.clear();
+    for (const auto &app_id : app_ids) {
```
levy5307, line 999: I think there is no need to use reference for int32_t

src/meta/meta_backup_service.cpp
```diff
@@ -984,6 +984,285 @@ void policy_context::sync_remove_backup_info(const backup_info &info, dsn::task_
...
+                      [&err, &remote_file](const dist::block_service::create_file_response &resp) {
+                          err = resp.err;
+                          remote_file = resp.file_handle;
+                      })
+        ->wait();
+    if (err != dsn::ERR_OK) {
+        ddebug_f("create file {} failed", file_name);
+        return err;
+    }
+    dassert(remote_file != nullptr,
```
levy5307, line 1058: dassert_f

src/meta/meta_backup_service.cpp
```diff
@@ -1569,5 +1848,40 @@ std::string backup_service::get_backup_path(const std::string &policy_name, int6
     ss << _policy_meta_root << "/" << policy_name << "/" << backup_id;
     return ss.str();
 }
+
+void backup_service::start_backup_apps(start_backup_apps_rpc rpc)
+{
+    const start_backup_apps_request &request = rpc.request();
+    start_backup_apps_response &response = rpc.response();
+
+    std::shared_ptr<backup_engine> engine = std::make_shared<backup_engine>(this);
```
levy5307, line 1857: I think unique_ptr is better here

src/meta/meta_backup_service.cpp
```diff
@@ -984,6 +984,285 @@ void policy_context::sync_remove_backup_info(const backup_info &info, dsn::task_
...
+    {
+        zauto_read_lock l;
+        _backup_service->get_state()->lock_read(l);
+        const std::shared_ptr<app_state> &app = _backup_service->get_state()->get_app(app_id);
+        if (app == nullptr || app->status != app_status::AS_AVAILABLE) {
+            return ERR_INVALID_STATE;
+        }
+        // do not write app envs to backup file
+        app_state tmp = *app;
+        if (!tmp.envs.empty()) {
```
levy5307, line 1082: remove `if(!tmp.envs.empty()) {`

src/meta/meta_backup_service.cpp
zhangyifan27: The pointer would be used to query backup status, so it stored as a shared_ptr in the map `_backup_states`.

neverchanje (dismissed):

src/meta/backup_engine.h
```diff
@@ -0,0 +1,83 @@
...
+    int64_t end_time_ms;
+
+    bool is_backup_failed;
+
+    int32_t app_id;
+    std::string app_name;
+
+    app_backup_info() : backup_id(0), start_time_ms(0), end_time_ms(0), is_backup_failed(false) {}
+
+    DEFINE_JSON_SERIALIZATION(backup_id, start_time_ms, end_time_ms, app_id, app_name)
```
hycdong, line 46: `is_backup_failed` is not included in json serialization, why it is not necessary to be serialized?

hycdong (approved):

src/meta/backup_engine.h
```diff
@@ -0,0 +1,82 @@
...
+private:
+    error_code get_app_stat(int32_t app_id, std::shared_ptr<app_state> &app);
+
+    backup_service *_backup_service;
+    dist::block_service::block_filesystem *_block_service;
+    std::string _provider_type;
+    dsn::task_tracker _tracker;
+
+    // lock the following variables.
+    dsn::zlock _lock;
```
neverchanje, line 74: You can declare `_lock` a mutable so that you can let `is_backing_up` const. Making locks mutable is a convention in C++.

```
    bool is_backing_up() const;
```


```suggestion
    mutable dsn::zlock _lock;
```

neverchanje (approved):

neverchanje:
You can do the commented fixes in later pr. No big problems :)

PR #776, opened 2021-03-05
feat(one-time backup): part-3 meta server handle rpc response

zhangyifan27:


src/meta/backup_engine.cpp
```diff
@@ -184,6 +185,94 @@ void backup_engine::on_backup_reply(error_code err,
...
+                         0,
+                         std::chrono::seconds(1));
+    }
+}
+
+void backup_engine::complete_current_backup()
+{
+    {
+        zauto_lock l(_lock);
+        for (int i = 0; i < _backup_status.size(); ++i) {
```
levy5307, line 267:
```
for (auto &status : _backup_status)
```

src/meta/backup_engine.cpp
```diff
@@ -184,6 +185,94 @@ void backup_engine::on_backup_reply(error_code err,
...
+    if (response.err == ERR_LOCAL_APP_FAILURE) {
+        derror_f("backup_id({}): backup for partition {} failed.",
+                 _cur_backup.backup_id,
+                 pid.to_string());
+        zauto_lock l(_lock);
+        is_backup_failed = true;
+        _backup_status[partition] = backup_status::FAILED;
+        return;
+    }
+
```
hycdong, line 236: It seems that you don't handle rpc error specially. If rpc errors happen, it will call function `backup_app_partition`, and let `_backup_status[pidx] = backup_status::ALIVE;`, like normal not-finished case. How about adding some comments or log here?

src/meta/backup_engine.cpp
```diff
@@ -184,6 +185,94 @@ void backup_engine::on_backup_reply(error_code err,
...
+             pid.to_string(),
+             primary.to_string(),
+             err.to_string(),
+             response.err.to_string());
+    dassert_f(response.pid == pid,
+              "backup partition id {} vs {} don't match",
+              response.pid.to_string(),
+              pid.to_string());
+    dassert_f(response.backup_id == _cur_backup.backup_id,
+              "backup id {} vs {} don't match",
```
levy5307, line 200: dcheck_eq

src/meta/backup_engine.cpp
```diff
@@ -184,6 +185,94 @@ void backup_engine::on_backup_reply(error_code err,
...
+}
+
+void backup_engine::write_backup_info()
+{
+    std::string file_name =
+        cold_backup::get_backup_info_file(_backup_service->backup_root(), _cur_backup.backup_id);
+    blob buf = dsn::json::json_forwarder<app_backup_info>::encode(_cur_backup);
+    error_code err = write_backup_file(file_name, buf);
+    if (err != ERR_OK) {
+        ddebug_f("backup_id({}): write backup info failed, retry it later.", _cur_backup.backup_id);
```
hycdong, line 254: Warning or error is better to search error logs.

neverchanje (approved):

hycdong (approved):

PR #778, opened 2021-03-08
refactor: make aio decoupled from dsn_runtime

levy5307:


src/replica/replica_stub.cpp
```diff
@@ -60,6 +60,7 @@
 #endif
 #include <dsn/utility/fail_point.h>
 #include <dsn/dist/remote_command.h>
+#include <aio/disk_engine.h>
```
neverchanje, line 63:
```suggestion
#include "aio/disk_engine.h"
```

src/replica/replica_stub.cpp
```diff
@@ -87,6 +88,11 @@ replica_stub::replica_stub(replica_state_subscriber subscriber /*= nullptr*/,
       _fs_manager(false),
       _bulk_load_downloading_count(0)
 {
+
+    // make disk_engine destructed after replica_stub,
+    // because replica_stub relies on the former to close files.
+    disk_engine::instance();
```
neverchanje, line 94: The destruction order in c++ is the reverse of the initialization order

Init:
1. replica_stub
2. disk_engine

Destruct:
1. disk_engine
2. replica_stub

If you want replica_stub destructed before aio, you should place this line in replication_service_app

hycdong (approved):

src/aio/disk_engine.cpp
```diff
@@ -29,19 +29,25 @@
...
 const char *native_aio_provider = "dsn::tools::native_aio_provider";
 DSN_REGISTER_COMPONENT_PROVIDER(native_linux_aio_provider, native_aio_provider);
-DSN_REGISTER_COMPONENT_PROVIDER(sim_aio_provider, "dsn::tools::sim_aio_provider");
+
+struct disk_engine_initializer
+{
+    disk_engine_initializer() { disk_engine::instance(); }
+};
+
+// make disk_engine destructed after service_engine, which is inited in dsn_global_init,
```
zhangyifan27, line 48: How do you guarantee the order of destruction here?

src/aio/disk_engine.cpp
```diff
@@ -151,10 +157,7 @@ aio_task *disk_file::on_write_completed(aio_task *wk, void *ctx, error_code err,
 disk_engine::disk_engine()
 {
     aio_provider *provider = utils::factory_store<aio_provider>::create(
-        FLAGS_aio_factory_name, dsn::PROVIDER_TYPE_MAIN, this);
-    if (nullptr == provider) {
-        dassert_f(false, "The config value of aio_factory_name is invalid");
-    }
+        native_aio_provider, dsn::PROVIDER_TYPE_MAIN, this);
```
zhangyifan27, line 160: So we don't need config `aio_factory_name` now?

levy5307: Yes, of course. I will delete it in the later

zhangyifan27 (approved):

PR #796, opened 2021-03-18
feat(one-time backup): add query backup status function

zhangyifan27:


src/common/backup.thrift
```diff
@@ -170,3 +170,28 @@ struct start_backup_app_response
...
+}
+
+struct query_backup_status_response
+{
+    // Possible error:
+    // - ERR_INVALID_PARAMETERS: no available backup for requested app
+    // - ERR_SERVICE_NOT_ACTIVE: meta doesn't enable backup service
+    1:dsn.error_code                 err;
+    2:optional list<backup_item>     backup_items;
+    3:string                         hint_message;
```
neverchanje, line 196:
```suggestion
    3:optional string                hint_message;
```

neverchanje (commented):
This is a simple function. Please add multiple unit tests for that. Be test-driven.

src/common/backup.thrift
zhangyifan27: We have a `hint_message` no matter the query succeeds or fails.

src/meta/meta_backup_service.cpp
```diff
@@ -1592,19 +1592,25 @@ void backup_service::start_backup_app(start_backup_app_rpc rpc)
...
-            response.err = ERR_INVALID_STATE;
-            response.hint_message = fmt::format("Backup failed: app {} is backing up now.", app_id);
-            return;
+    {
+        zauto_lock l(_lock);
+        for (const auto &backup : _backup_states) {
+            if (app_id == backup->get_backup_app_id() && backup->is_backing_up()) {
+                response.err = ERR_INVALID_STATE;
+                response.hint_message =
+                    fmt::format("Backup failed: app {} is backing up now.", app_id);
```
levy5307, line 1601:
```
fmt::format("Backup failed: app {} is backing up now.", app_id);
```
=====>
fmt::format("Backup failed: app {} is making a backup now.", app_id);

neverchanje (dismissed):

src/meta/meta_backup_service.cpp
```diff
@@ -1617,5 +1623,31 @@ void backup_service::start_backup_app(start_backup_app_rpc rpc)
...
+    }
+
+    if (response.backup_items.empty()) {
+        response.err = ERR_INVALID_PARAMETERS;
+        response.hint_message = fmt::format("No available backup for app {}.", app_id);
+        return;
+    }
+    response.__isset.backup_items = true;
+    response.hint_message = fmt::format(
+        "There are {} available backups for app {}.", response.backup_items.size(), app_id);
```
hycdong, line 1648: What is the meaning of 'available backups'? Is `response.backup_items.size()` always equal to app partition_count?

zhangyifan27: `response.backup_items.size()` is the number of backups of the app stored in meta memory.

hycdong: Well, I misunderstand before, and I raise another question. Will response contain many history backups for this app if meta server runs long time and this app have many backups? The rpc caller such as backup manager should analyse those backups to find out who is the real backup it wants.

zhangyifan27: Yes. I added a optional field `backup_id` to help users get specific backup info. But if `backup_id` is not specified, meta will response with all backups for the requested app.

hycdong (approved):

levy5307 (approved):

PR #798, opened 2021-03-25
feat(split-client): c++ client split related implementation

hycdong:
This pull request contains c++ client partition split related implementation, including:
1. handle split related error code: `ERR_SPLITTING`, `ERR_PARENT_PARTITION_MISUSED`
  - when parent partition is registering child partition, it will reject read write requests by return `ERR_SPLITTING`, client won't query meta and will retry in remaining time
  - when child partition is active, the request should be sent to child, parent will return `ERR_PARENT_PARTITION_MISUSED`, client should update all partitions config to update both parent and child config
2. handle the condition that client query config for a splitting table
  - when client query config for a splitting table, it will get child partition config, if child is not ready, its requests should be redirected to parent
  -  it is a valid case that client header gpid is different from result gpid, update check in function `call_task`, not assert when gpid changed
3. handle app partition_count changed while query meta
  - in pervious implementation, if partition_count changed, client will core. Now server support split (new_count = old_count*2) and cancel split (new_count = old_count / 2), client should not assert in those cases.


src/client/partition_resolver_simple.cpp
```diff
@@ -48,7 +48,13 @@ void partition_resolver_simple::resolve(uint64_t partition_hash,
     if (_app_partition_count != -1) {
         idx = get_partition_index(_app_partition_count, partition_hash);
         rpc_address target;
-        if (ERR_OK == get_address(idx, target)) {
+        auto err = get_address(idx, target);
+        if (err == ERR_CHILD_NOT_READY) {
```
levy5307, line 52: dsn_unlikely

src/client/partition_resolver_simple.cpp
```diff
@@ -48,7 +48,13 @@ void partition_resolver_simple::resolve(uint64_t partition_hash,
...
+            // child partition is not ready, its requests should be sent to parent partition
+            idx -= _app_partition_count / 2;
+            err = get_address(idx, target);
+        }
+        if (err == ERR_OK) {
```
levy5307, line 57: dsn_likely

src/client/partition_resolver_simple.cpp
```diff
@@ -68,27 +74,36 @@ void partition_resolver_simple::resolve(uint64_t partition_hash,
...
+        err == ERR_BUSY //  busy (rpc busy or throttling busy)
+        ||
+        err == ERR_SPLITTING // partition is splitting, reject read and write
         ) {
+        return;
+    }
+
+    zauto_write_lock l(_config_lock);
+    if (err == ERR_PARENT_PARTITION_MISUSED) {
+        ddebug("clear all partition configuration cache due to access failure %s at %d.%d",
```
levy5307, line 94: ddebug_f

src/client/partition_resolver_simple.cpp
```diff
@@ -68,27 +75,36 @@ void partition_resolver_simple::resolve(uint64_t partition_hash,
...
+        _app_partition_count = -1;
+    } else {
+        ddebug_f("clear partition configuration cache {}.{} due to access failure {}",
+                 _app_id,
+                 partition_index,
+                 err);
+        auto it = _config_cache.find(partition_index);
+        if (it != _config_cache.end()) {
+            _config_cache.erase(it);
         }
```
acelyc111, line 108:
```suggestion
        _config_cache.erase(partition_index);
```

src/client/partition_resolver.cpp
```diff
@@ -105,13 +105,13 @@ void partition_resolver::call_task(const rpc_response_task_ptr &t)
...
                 if (hdr.gpid.value() != result.pid.value()) {
-                    dassert(hdr.gpid.value() == 0, "inconsistent gpid");
-                    hdr.gpid = result.pid;
-
-                    // update thread hash if not assigned by applications
-                    if (hdr.client.thread_hash == 0) {
+                    if (hdr.client.thread_hash == 0 // thread_hash is not assigned by applications
+                        ||
+                        hdr.gpid.value() != 0 // requests set to child redirect to parent
+                        ) {
```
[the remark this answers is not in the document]
hycdong: Sure, this is valid and pass the format validation.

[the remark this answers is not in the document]
acelyc111: Is the format I comment valid? If it is, you can improve it, a single `||` use one line seems ugly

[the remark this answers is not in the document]
hycdong: Oh, I misunderstand before. I tried your suggesting format, but the format validation will make `||` in a single line.

acelyc111 (approved):

levy5307 (approved):

PR #805, opened 2021-03-30
feat(http): update query manual compaction interface

hycdong:
#696 adds a http interface to query manual compaction status, this pr updates it:
- rename all `compact` and `compaction` into `manual_compact` and `manual_compaction`
- add a new status `kIdle` to distinguish not execute manual compaction and manual compaction finished
- update count in response from string into integer

The previous request and response is like:
```
curl 'http://localhost:34801/replica/compaction?app_id=1'
{
    "status": {
        "CompactionFinish": "8", 
        "CompactionQueue": "0", 
        "CompactionRunning": "0"
    }
}
```

The new request and response is like:
```
curl 'http://localhost:34801/replica/manual_compaction?app_id=1'
{
    "status": {
        "finished": 8,
        "idle": 0, 
        "queuing": 0, 
        "running": 0
    }
}
```

### DIFF
--- a/src/replica/replica.cpp
+++ b/src/replica/replica.cpp
@@ -471,7 +471,7 @@ void replica::close()
     ddebug("%s: replica closed, time_used = %" PRIu64 "ms", name(), dsn_now_ms() - start_time);
 }
 
-std::string replica::query_compact_state() const
+std::string replica::query_manual_compact_state() const
 {
     dassert_replica(_app != nullptr, "");
     return _app->query_compact_state();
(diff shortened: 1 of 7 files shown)


src/replica/replica.cpp
```diff
@@ -480,22 +480,24 @@ std::string replica::query_compact_state() const
...
     switch (status) {
-    case kFinish:
-        return "CompactionFinish";
-    case kRunning:
-        return "CompactionRunning";
+    case kIdle:
+        return "idle";
     case kQueue:
-        return "CompactionQueue";
+        return "queue";
```
neverchanje, line 486: I prefer queuing, or "waiting", rather than "queue"

src/replica/replica.cpp
```diff
@@ -480,22 +480,24 @@ std::string replica::query_compact_state() const
...
+    case kRunning:
+        return "running";
+    case kFinish:
+        return "finish";
```
neverchanje, line 490:
```suggestion
        return "finished";
```

neverchanje (approved):

levy5307 (approved):

PR #809, opened 2021-04-02
refactor: restrict remote-command to only one command name

neverchanje:


src/utils/command_manager.cpp
```diff
@@ -38,29 +39,19 @@ dsn_handle_t command_manager::register_command(const std::vector<std::string> &c
...
 
-    for (const std::string &cmd : commands) {
-        if (!cmd.empty()) {
-            is_valid_cmd = true;
-            auto it = _handlers.find(cmd);
-            dassert(it == _handlers.end(), "command '%s' already regisered", cmd.c_str());
-        }
-    }
-    dassert(is_valid_cmd, "should not register empty command");
+    utils::auto_write_lock l(_lock);
```
acelyc111, line 46:
```
    auto c = new command_instance();
    c->commands = commands;
    c->help_long = help_long;
    c->help_short = help_one_line;
    c->handler = std::move(handler);

    utils::auto_write_lock l(_lock);
    _handlers[cmd] = c;
```

src/utils/test/command_manager_test.cpp
```diff
@@ -24,40 +24,39 @@
...
- */
-
 #include <dsn/tool-api/command_manager.h>
 #include <gtest/gtest.h>
 
-using namespace ::dsn;
+namespace dsn {
 
-void command_manager_module_init()
+struct command_manager_test : public ::testing::Test
```
acelyc111, line 32:
```suggestion
class command_manager_test : public ::testing::Test
```

src/utils/command_manager.cpp
```diff
@@ -38,29 +39,19 @@ dsn_handle_t command_manager::register_command(const std::vector<std::string> &c
                                                const std::string &help_long,
                                                command_handler handler)
 {
-    utils::auto_write_lock l(_lock);
-    bool is_valid_cmd = false;
+    dcheck_eq(commands.size(), 1);
```
acelyc111, line 42: So `commands` can be replaced by `const std::string &command` ?

PR #814, opened 2021-04-09
feat(one-time backup): support user specified backup path

zhangyifan27:
This patch added a new optional field `backup_path` to `start_backup_app_request` so that users could backup an app to a specific path.

For local_service, suppose we had a configuration:
```
[[block_service.local_service]]
    type = local_service
    args = /usr/pegasus
```
If we specify `backup` as backup_path when use one-time backup, the previous backup directory is `/usr/pegasus`, now the backup directory is `/usr/pegasus/backup`. If we want to backup app to a totally user specified path, we should config args as '/'. 

For hdfs_service, the following configurations ard recommended, so that we could only constructed one `hdfsFS` object for the same namenode.
```
[[block_service.hdfs_tjwq]]
    type = hdfs_service
    args = hdfs://tjwqstaging-hdd

[[block_service.hdfs_zjy]]
    type = hdfs_service
    args = hdfs://zjyprc-hadoop
```
And if we use this configuration, users should specify a backup path every time they want to backup tables to hdfs, otherwise backup files would uploaded to the root path '/'. If both hdfs_service backup path and user specified backup path are set, we would not start a backup, and suggest users to modify the configuration.

I have tested this on onebox cluster and I will add function tests in the following patchs.

### DIFF
--- a/src/block_service/hdfs/hdfs_service.cpp
+++ b/src/block_service/hdfs/hdfs_service.cpp
@@ -70,8 +70,9 @@ error_code hdfs_service::initialize(const std::vector<std::string> &args)
     if (args.size() < 1) {
         return ERR_INVALID_PARAMETERS;
     }
+    // Name_node and root_path should be set in args of block_service configuration.
+    // If no path was configured, just use "/" as default root path.
     _hdfs_name_node = args[0];
-    // If no path was configured in 'block_service', just use "/" as default root path..
     _hdfs_path = args.size() >= 2 ? args[1] : "/";
     ddebug_f("hdfs backup root path is initialized to {}.", _hdfs_path);
 
--- a/src/block_service/hdfs/hdfs_service.cpp
+++ b/src/block_service/hdfs/hdfs_service.cpp
@@ -71,6 +71,7 @@ error_code hdfs_service::initialize(const std::vector<std::string> &args)
         return ERR_INVALID_PARAMETERS;
     }
     _hdfs_name_node = args[0];
+    // If no path is configured in 'block_service', just use "/" as default root path..
     _hdfs_path = args.size() >= 2 ? args[1] : "/";
     ddebug_f("hdfs backup root path is initialized to {}.", _hdfs_path);
 
--- a/src/block_service/hdfs/hdfs_service.cpp
+++ b/src/block_service/hdfs/hdfs_service.cpp
@@ -67,11 +67,13 @@ hdfs_service::~hdfs_service()
 
 error_code hdfs_service::initialize(const std::vector<std::string> &args)
 {
-    if (args.size() != 2) {
+    if (args.size() < 1) {
         return ERR_INVALID_PARAMETERS;
     }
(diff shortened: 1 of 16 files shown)


src/block_service/hdfs/hdfs_service.cpp
```diff
@@ -67,11 +67,13 @@ hdfs_service::~hdfs_service()
 
 error_code hdfs_service::initialize(const std::vector<std::string> &args)
 {
-    if (args.size() != 2) {
+    if (args.size() < 1) {
         return ERR_INVALID_PARAMETERS;
     }
     _hdfs_name_node = args[0];
-    _hdfs_path = args[1];
+    _hdfs_path = args.size() >= 2 ? args[1] : "/";
```
neverchanje, line 74: Please add a comment here: what do the arguments mean separately?

Shuo-Jia (dismissed):

src/meta/backup_engine.cpp
```diff
@@ -71,6 +72,12 @@ error_code backup_engine::set_block_service(const std::string &provider)
     return ERR_OK;
 }
 
+void backup_engine::set_backup_path(const std::string &path)
+{
+    ddebug_f("backup path is set to {}.", path);
```
Smityz, line 77: Why add a dot in the end?

levy5307:
```
 If we want to backup app to a totally user specified path, we shouldn't config block_service root path or configure it as '/'. 
```
Should we add a judgement to avoid this situation?

src/meta/backup_engine.cpp
zhangyifan27: It's just a ending period, sometimes this could be omitted for brevity.

zhangyifan27:
> ```
>  If we want to backup app to a totally user specified path, we shouldn't config block_service root path or configure it as '/'. 
> ```
> 
> Should we add a judgement to avoid this situation?



> ```
>  If we want to backup app to a totally user specified path, we shouldn't config block_service root path or configure it as '/'. 
> ```
> 
> Should we add a judgement to avoid this situation?

I didn't remove the default root path configured in `block_service` because I want to maintain compatibility with `policy backup`, which doesn't have `backup_path` field. And also this field is optional, so maybe sometimes users also want to use the configured default path.

levy5307:
> > ```
> >  If we want to backup app to a totally user specified path, we shouldn't config block_service root path or configure it as '/'. 
> > ```
> > 
> > 
> > Should we add a judgement to avoid this situation?
> 
> I didn't remove the default root path configured in `block_service` because I want to maintain compatibility with `policy backup`, which doesn't have `backup_path` field. And also this field is optional, so maybe sometimes users also want to use the configured default path.

I understood. What I mean is that if someone wants to specify user path, we should check if the block_service root path is configured or not.

levy5307 (dismissed):

neverchanje (approved):

levy5307 (approved):

PR #815, opened 2021-04-14
fix: fix data loss

ZhongChaoqiang:
### What problem does this PR solve?
Data loss becase wrong init_durable_decree in .init-info file.
For details, see apache/incubator-pegasus#719

### What is changed and how it works?
using last_durable_decree for init_durable_decree in .init-info file.

### Check List
Tests

Manual test (add detailed scripts or steps below)


zhangyifan27 (commented):
Could you add a test showing that this works the way you expect?

ZhongChaoqiang:
@neverchanje   @hycdong
Can you help to review the code?
Thanks!

ZhongChaoqiang:
> Could you add a test showing that this works the way you expect?

@zhangyifan27
OK，I will try to add a unit test. But this is a little complicated. Can you recommend a similar test case for me? Thanks!

Smityz:
mentioned in https://github.com/apache/incubator-pegasus/issues/719

PR #818, opened 2021-04-19
fix: support retry when io write incompletely

Shuo-Jia:
https://github.com/apache/incubator-pegasus/issues/659 has reported the io write operation bug and it only `retry` to decrease the probability of occurrence. This pr refer [stackovflow: handling-incomplete-write-calls](https://stackoverflow.com/questions/32683086/handling-incomplete-write-calls) ,  [rocksdb](https://github.com/facebook/rocksdb/blob/d89483098ff4a77bd0ef12454a64554805984cc5/env/io_posix.cc#L119) and add the retry operation.

## Note：
- This pr only decrease the probability but not resolve it completely
- This pr add `FAIL_POINT_INJECT_OFF_F` macro to mock `pwrite` to reproduce the bug

### DIFF
--- a/include/dsn/utility/fail_point.h
+++ b/include/dsn/utility/fail_point.h
@@ -22,8 +22,8 @@
 
 #include <dsn/utility/string_view.h>
 
-/// The only entry to define a fail point.
-/// When a fail point is defined, it's referenced via the name.
+/// The only entry to define a fail point with `return()` function: lambda function must be
+/// return non-void type;  When a fail point is defined, it's referenced via the name.
 #define FAIL_POINT_INJECT_F(name, lambda)                                                          \
     do {                                                                                           \
         if (dsn_likely(!::dsn::fail::_S_FAIL_POINT_ENABLED))                                       \
@@ -32,6 +32,19 @@
         auto __Res = ::dsn::fail::eval(name);                                                      \
         if (__Res != nullptr) {                                                                    \
             return __Func(*__Res);                                                                 \
+        }                                                                                          \
+    } while (0)
+
+/// The only entry to define a fail point with `off()` function: lambda  function must be
+/// return void type; When a fail point is defined, it's referenced via the name.
+#define FAIL_POINT_INJECT_OFF_F(name, lambda)                                                      \
+    do {                                                                                           \
+        if (dsn_likely(!::dsn::fail::_S_FAIL_POINT_ENABLED))                                       \
+            break;                                                                                 \
+        auto __Func = lambda;                                                                      \
+        auto __Res = ::dsn::fail::eval(name);                                                      \
+        if (__Res == nullptr) {                                                                    \
+            __Func();                                                                              \
         }                                                                                          \
     } while (0)
 
--- a/include/dsn/utility/fail_point.h
+++ b/include/dsn/utility/fail_point.h
@@ -23,7 +23,7 @@
 #include <dsn/utility/string_view.h>
 
 /// The only entry to define a fail point with `return()` function: lambda function must be
-/// return non-void type;  When a fail point is defined, it's referenced via the name.
+/// return non-void type. When a fail point is defined, it's referenced via the name.
 #define FAIL_POINT_INJECT_F(name, lambda)                                                          \
     do {                                                                                           \
         if (dsn_likely(!::dsn::fail::_S_FAIL_POINT_ENABLED))                                       \
@@ -36,7 +36,7 @@
     } while (0)
 
 /// The only entry to define a fail point with `off()` function: lambda  function must be
-/// return void type; When a fail point is defined, it's referenced via the name.
+/// return void type. When a fail point is defined, it's referenced via the name.
 #define FAIL_POINT_INJECT_OFF_F(name, lambda)                                                      \
     do {                                                                                           \
         if (dsn_likely(!::dsn::fail::_S_FAIL_POINT_ENABLED))                                       \
--- a/include/dsn/utility/fail_point.h
+++ b/include/dsn/utility/fail_point.h
@@ -22,7 +22,7 @@
 
 #include <dsn/utility/string_view.h>
 
-/// The only entry to define a fail point with `return()` function: lambda function must be
+/// The only entry to define a fail point with `return` function: lambda function must be
 /// return non-void type. When a fail point is defined, it's referenced via the name.
 #define FAIL_POINT_INJECT_F(name, lambda)                                                          \
     do {                                                                                           \
@@ -35,9 +35,9 @@
         }                                                                                          \
     } while (0)
 
-/// The only entry to define a fail point with `off()` function: lambda  function must be
+/// The only entry to define a fail point with `void` function: lambda  function must be
 /// return void type. When a fail point is defined, it's referenced via the name.
-#define FAIL_POINT_INJECT_OFF_F(name, lambda)                                                      \
+#define FAIL_POINT_INJECT_VOID_F(name, lambda)                                                     \
     do {                                                                                           \
         if (dsn_likely(!::dsn::fail::_S_FAIL_POINT_ENABLED))                                       \
             break;                                                                                 \
--- a/src/aio/native_linux_aio_provider.cpp
+++ b/src/aio/native_linux_aio_provider.cpp
@@ -69,18 +69,18 @@ error_code native_linux_aio_provider::write(const aio_context &aio_ctx,
                                             /*out*/ uint32_t *processed_bytes)
 {
     bool enable_retry = true;
-    uint32_t buffert_offset = 0;
+    uint32_t buffer_offset = 0;
     while (true) {
         uint32_t ret = pwrite(static_cast<int>((ssize_t)aio_ctx.file),
-                              aio_ctx.buffer + buffert_offset,
-                              aio_ctx.buffer_size - buffert_offset,
-                              aio_ctx.file_offset + buffert_offset);
+                              aio_ctx.buffer + buffer_offset,
+                              aio_ctx.buffer_size - buffer_offset,
+                              aio_ctx.file_offset + buffer_offset);
         if (ret < 0) {
             return ERR_FILE_OPERATION_FAILED;
         }
 
-        buffert_offset += ret;
-        if (buffert_offset != aio_ctx.buffer_size && enable_retry) {
+        buffer_offset += ret;
+        if (buffer_offset != aio_ctx.buffer_size && enable_retry) {
             dwarn_f("write incomplete, request_size={}, write_size={}, will retry delay 10ms",
                     aio_ctx.buffer_size,
                     ret);
@@ -91,7 +91,7 @@ error_code native_linux_aio_provider::write(const aio_context &aio_ctx,
         break;
     }
 
-    *processed_bytes = buffert_offset;
+    *processed_bytes = buffer_offset;
     return ERR_OK;
 }
 
@@ -100,12 +100,12 @@ error_code native_linux_aio_provider::read(const aio_context &aio_ctx,
 {
 
     bool enable_retry = true;
-    uint32_t buffert_offset = 0;
+    uint32_t buffer_offset = 0;
     while (true) {
         ssize_t ret = pread(static_cast<int>((ssize_t)aio_ctx.file),
-                            aio_ctx.buffer + buffert_offset,
-                            aio_ctx.buffer_size - buffert_offset,
-                            aio_ctx.file_offset + buffert_offset);
+                            aio_ctx.buffer + buffer_offset,
+                            aio_ctx.buffer_size - buffer_offset,
+                            aio_ctx.file_offset + buffer_offset);
         if (ret < 0) {
             return ERR_FILE_OPERATION_FAILED;
         }
@@ -113,8 +113,8 @@ error_code native_linux_aio_provider::read(const aio_context &aio_ctx,
             return ERR_HANDLE_EOF;
         }
 
-        buffert_offset += ret;
-        if (buffert_offset != aio_ctx.buffer_size && enable_retry) {
+        buffer_offset += ret;
+        if (buffer_offset != aio_ctx.buffer_size && enable_retry) {
             dwarn_f("read incomplete, request_size={}, read_size={}, will retry delay 10ms",
                     aio_ctx.buffer_size,
                     ret);
@@ -125,7 +125,7 @@ error_code native_linux_aio_provider::read(const aio_context &aio_ctx,
         break;
     }
 
-    *processed_bytes = buffert_offset;
+    *processed_bytes = buffer_offset;
     return ERR_OK;
 }
 
--- a/src/aio/native_linux_aio_provider.cpp
+++ b/src/aio/native_linux_aio_provider.cpp
@@ -71,40 +71,36 @@ error_code native_linux_aio_provider::write(const aio_context &aio_ctx,
                                             /*out*/ uint32_t *processed_bytes)
 {
     uint32_t buffer_offset = 0;
-    uint32_t remaing_size = aio_ctx.buffer_size;
-    while (remaing_size > 0) {
+    do {
         uint32_t ret = pwrite(static_cast<int>((ssize_t)aio_ctx.file),
                               (char *)aio_ctx.buffer + buffer_offset,
-                              remaing_size,
+                              aio_ctx.buffer_size - buffer_offset,
                               aio_ctx.file_offset + buffer_offset);
-        if (ret < 0) {
+        if (dsn_unlikely(ret < 0)) {
             if (errno == EINTR) {
-                dwarn_f("write failed with EINTR, will retry delay 10ms.");
-                usleep(1e4);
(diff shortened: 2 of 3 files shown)


src/aio/native_linux_aio_provider.cpp
```diff
@@ -68,14 +70,35 @@ error_code native_linux_aio_provider::flush(dsn_handle_t fh)
...
+                ret -= 1;
+            }
+        });
+        buffer_offset += ret;
+        if (buffer_offset != aio_ctx.buffer_size && enable_retry) {
+            dwarn_f("write incomplete, request_size={}, write_size={}, will retry delay 10ms",
+                    aio_ctx.buffer_size,
+                    ret);
+            enable_retry = false;
+            usleep(10000);
```
levy5307, line 95: 10000 --> 1e4

src/aio/native_linux_aio_provider.cpp
```diff
@@ -68,14 +70,35 @@ error_code native_linux_aio_provider::flush(dsn_handle_t fh)
...
 {
-    ssize_t ret = pwrite(static_cast<int>((ssize_t)aio_ctx.file),
-                         aio_ctx.buffer,
-                         aio_ctx.buffer_size,
-                         aio_ctx.file_offset);
-    if (ret < 0) {
-        return ERR_FILE_OPERATION_FAILED;
+    bool enable_retry = true;
+    uint32_t buffer_offset = 0;
+    while (true) {
```
levy5307, line 75: while(enable_retry)

src/aio/native_linux_aio_provider.cpp
```diff
@@ -68,14 +70,33 @@ error_code native_linux_aio_provider::flush(dsn_handle_t fh)
...
+        uint32_t ret = pwrite(static_cast<int>((ssize_t)aio_ctx.file),
+                              (char *)aio_ctx.buffer + buffer_offset,
+                              aio_ctx.buffer_size - buffer_offset,
+                              aio_ctx.file_offset + buffer_offset);
+        if (ret < 0) {
+            return ERR_FILE_OPERATION_FAILED;
+        }
+
+        // mock the `ret` to reproduce the `write incomplete` case
+        FAIL_POINT_INJECT_OFF_F("aio_pwrite", [&]() -> void {
```
levy5307, line 85: [&]() --> [retires]()

Shuo-Jia: `ret` is also need

src/aio/native_linux_aio_provider.cpp
```diff
@@ -68,14 +70,33 @@ error_code native_linux_aio_provider::flush(dsn_handle_t fh)
...
-        return ERR_FILE_OPERATION_FAILED;
+    int retries = 2;
+    uint32_t buffer_offset = 0;
+    while (--retries >= 0) {
+        uint32_t ret = pwrite(static_cast<int>((ssize_t)aio_ctx.file),
+                              (char *)aio_ctx.buffer + buffer_offset,
+                              aio_ctx.buffer_size - buffer_offset,
+                              aio_ctx.file_offset + buffer_offset);
+        if (ret < 0) {
+            return ERR_FILE_OPERATION_FAILED;
```
zhangyifan27, line 81: Why don't we try again here?

src/aio/native_linux_aio_provider.cpp
```diff
@@ -68,14 +70,33 @@ error_code native_linux_aio_provider::flush(dsn_handle_t fh)
 error_code native_linux_aio_provider::write(const aio_context &aio_ctx,
                                             /*out*/ uint32_t *processed_bytes)
 {
-    ssize_t ret = pwrite(static_cast<int>((ssize_t)aio_ctx.file),
-                         aio_ctx.buffer,
-                         aio_ctx.buffer_size,
-                         aio_ctx.file_offset);
-    if (ret < 0) {
-        return ERR_FILE_OPERATION_FAILED;
+    int retries = 2;
```
zhangyifan27, line 73: Why we only try it twice here? What if the write operation still fail ?

zhangyifan27 (commented):
I think you could look at the implementation of Kudu: https://github.com/apache/kudu/blob/6983d25c35df8569c1851bf27e2c75e98036c10d/src/kudu/util/env_posix.cc#L490

When errno == EINTR, Kudu would keeps retrying util other errors occur or write successfully, and also, keeps retrying until the number of remaining bytes (unwritten bytes) is 0.

In addition, it's better to test it on a real cluster to see if this issue could be resolved.

src/aio/native_linux_aio_provider.cpp
Shuo-Jia: 1. This pr is not fix for `write failed` but only `write incompletely`
2. This bug is not easy to repeat, and it only decrease the probability but not resolve it completely

src/aio/native_linux_aio_provider.cpp
Shuo-Jia: It will core dump in the following code if  still fail. I don't make sure it will suceess if we let it try forever, so if retry failed, I choose this write is failed.

Shuo-Jia:
> I think you could look at the implementation of Kudu: https://github.com/apache/kudu/blob/6983d25c35df8569c1851bf27e2c75e98036c10d/src/kudu/util/env_posix.cc#L490
> 
> When errno == EINTR, Kudu would keeps retrying util other errors occur or write successfully, and also, keeps retrying until the number of remaining bytes (unwritten bytes) is 0.
> 
> In addition, it's better to test it on a real cluster to see if this issue could be resolved.

Like this ?
```c++
ssize_t
write_with_retry (int fd, const void* buf, size_t size)
{
    ssize_t ret;
    while (size > 0) {
        do
        {
             ret = write(fd, buf, size);
        } while ((ret < 0) && (errno == EINTR));
        if (ret < 0)
            return ret;
        size -= ret;
        buf += ret;
    }
    return 0;
}
```

include/dsn/utility/fail_point.h
```diff
@@ -22,8 +22,8 @@
 
 #include <dsn/utility/string_view.h>
 
-/// The only entry to define a fail point.
-/// When a fail point is defined, it's referenced via the name.
+/// The only entry to define a fail point with `return()` function: lambda function must be
+/// return non-void type;  When a fail point is defined, it's referenced via the name.
```
acelyc111, line 26:
```suggestion
/// return non-void type. When a fail point is defined, it's referenced via the name.
```

include/dsn/utility/fail_point.h
```diff
@@ -35,6 +35,19 @@
         }                                                                                          \
     } while (0)
 
+/// The only entry to define a fail point with `off()` function: lambda  function must be
+/// return void type; When a fail point is defined, it's referenced via the name.
```
acelyc111, line 39:
```suggestion
/// The only entry to define a fail point with `off()` function: lambda function must be
/// return void type. When a fail point is defined, it's referenced via the name.
```

src/aio/native_linux_aio_provider.cpp
```diff
@@ -68,14 +70,43 @@ error_code native_linux_aio_provider::flush(dsn_handle_t fh)
...
 {
-    ssize_t ret = pwrite(static_cast<int>((ssize_t)aio_ctx.file),
-                         aio_ctx.buffer,
-                         aio_ctx.buffer_size,
-                         aio_ctx.file_offset);
-    if (ret < 0) {
-        return ERR_FILE_OPERATION_FAILED;
+    uint32_t buffer_offset = 0;
+    uint32_t remaing_size = aio_ctx.buffer_size;
+    while (remaing_size > 0) {
```
acelyc111, line 75: For performance reason (CPU Branch Prediction), use:
```
do {
   ...
} while (dsn_unlikely(remaing_size > 0));
```
would be better?

src/aio/native_linux_aio_provider.cpp
```diff
@@ -68,14 +70,43 @@ error_code native_linux_aio_provider::flush(dsn_handle_t fh)
...
+                continue;
+            }
+            derror_f("write failed, errno={}, return ERR_FILE_OPERATION_FAILED", strerror(errno));
+            return ERR_FILE_OPERATION_FAILED;
+        }
+
+        // mock the `ret` to reproduce the `write incomplete` case in the first write
+        FAIL_POINT_INJECT_OFF_F("aio_pwrite", [&]() -> void {
+            if (buffer_offset == 0) {
+                ret -= 1;
```
acelyc111, line 93:
```suggestion
                --ret;
```

src/aio/native_linux_aio_provider.cpp
```diff
@@ -68,14 +70,43 @@ error_code native_linux_aio_provider::flush(dsn_handle_t fh)
...
+            }
+        });
+
+        buffer_offset += ret;
+        remaing_size -= ret;
+        if (remaing_size > 0) {
```
acelyc111, line 99:
```suggestion
        if (dsn_unlikely(remaing_size > 0)) {
```

src/aio/native_linux_aio_provider.cpp
```diff
@@ -68,14 +70,43 @@ error_code native_linux_aio_provider::flush(dsn_handle_t fh)
...
+            if (buffer_offset == 0) {
+                --ret;
+            }
+        });
+
+        buffer_offset += ret;
+        remaing_size -= ret;
+        if (remaing_size > 0) {
+            dwarn_f("write incomplete, request_size={}, total_write_size={}, this_write_size={}, "
+                    "will retry delay 10ms",
```
levy5307, line 101: remove `will retry delay 10ms`

src/aio/native_linux_aio_provider.cpp
levy5307, line 75: remove `remaining_size`. use buffer_offset < aio_ctx.buffer_size to see whether it is over or not

src/aio/native_linux_aio_provider.cpp
```diff
@@ -68,14 +70,43 @@ error_code native_linux_aio_provider::flush(dsn_handle_t fh)
...
-    if (ret < 0) {
-        return ERR_FILE_OPERATION_FAILED;
+    uint32_t buffer_offset = 0;
+    uint32_t remaing_size = aio_ctx.buffer_size;
+    while (remaing_size > 0) {
+        uint32_t ret = pwrite(static_cast<int>((ssize_t)aio_ctx.file),
+                              (char *)aio_ctx.buffer + buffer_offset,
+                              remaing_size,
+                              aio_ctx.file_offset + buffer_offset);
+        if (ret < 0) {
```
acelyc111, line 80:
```suggestion
        if (dsn_unlikely(ret < 0)) {
```

src/aio/native_linux_aio_provider.cpp
```diff
@@ -68,14 +70,39 @@ error_code native_linux_aio_provider::flush(dsn_handle_t fh)
...
+        uint32_t ret = pwrite(static_cast<int>((ssize_t)aio_ctx.file),
+                              (char *)aio_ctx.buffer + buffer_offset,
+                              aio_ctx.buffer_size - buffer_offset,
+                              aio_ctx.file_offset + buffer_offset);
+        if (dsn_unlikely(ret < 0)) {
+            if (errno == EINTR) {
+                dwarn_f("write failed with EINTR and will retry it.");
+                continue;
+            }
+            derror_f("write failed, errno={}, return ERR_FILE_OPERATION_FAILED.", strerror(errno));
```
levy5307, line 84:
```
res_err = ERR_FILE_OPERATION_FAILED;
derror_f("write failed, errno={}, return {}.", strerror(errno), res_err);
return res_err;
```

src/aio/native_linux_aio_provider.cpp
```diff
@@ -68,14 +70,39 @@ error_code native_linux_aio_provider::flush(dsn_handle_t fh)
...
-    *processed_bytes = static_cast<uint32_t>(ret);
+    uint32_t buffer_offset = 0;
+    do {
+        uint32_t ret = pwrite(static_cast<int>((ssize_t)aio_ctx.file),
+                              (char *)aio_ctx.buffer + buffer_offset,
+                              aio_ctx.buffer_size - buffer_offset,
+                              aio_ctx.file_offset + buffer_offset);
+        if (dsn_unlikely(ret < 0)) {
+            if (errno == EINTR) {
+                dwarn_f("write failed with EINTR and will retry it.");
```
levy5307, line 81: dwarn_f("write failed with {} and will retry it.", strerror(errno));

levy5307 (dismissed):

src/aio/native_linux_aio_provider.cpp
```diff
@@ -68,15 +70,42 @@ error_code native_linux_aio_provider::flush(dsn_handle_t fh)
...
-                         aio_ctx.file_offset);
-    if (ret < 0) {
-        return ERR_FILE_OPERATION_FAILED;
-    }
-    *processed_bytes = static_cast<uint32_t>(ret);
-    return ERR_OK;
+    dsn::error_code resp = ERR_OK;
+    uint32_t buffer_offset = 0;
+    do {
+        uint32_t ret = pwrite(static_cast<int>((ssize_t)aio_ctx.file),
```
neverchanje, line 76:
```suggestion
        // ret is the written data size
        uint32_t ret = pwrite(static_cast<int>((ssize_t)aio_ctx.file),
```

src/aio/native_linux_aio_provider.cpp
```diff
@@ -68,15 +70,42 @@ error_code native_linux_aio_provider::flush(dsn_handle_t fh)
...
+                dwarn_f("write failed with errno={} and will retry it.", strerror(errno));
+                continue;
+            }
+            resp = ERR_FILE_OPERATION_FAILED;
+            derror_f("write failed with errno={}, return {}.", strerror(errno), resp);
+            return resp;
+        }
+
+        // mock the `ret` to reproduce the `write incomplete` case in the first write
+        FAIL_POINT_INJECT_OFF_F("aio_pwrite", [&]() -> void {
```
neverchanje, line 91:
```suggestion
        FAIL_POINT_INJECT_OFF_F("aio_pwrite_incomplete", [&]() -> void {
```

src/aio/test/aio.cpp
```diff
@@ -37,6 +38,8 @@ DEFINE_TASK_CODE_AIO(LPC_AIO_TEST, TASK_PRIORITY_COMMON, THREAD_POOL_TEST_SERVER
 
 TEST(core, aio)
 {
+    fail::setup();
+    fail::cfg("aio_pwrite", "off()");
```
hycdong, line 42: aio_pwrite_incomplete

Shuo-Jia:
I have test `write incomplete` case  that the log is too large(>=2147479552 bytes)  in test cluster, the result show the `retry` will resolve the problem:
```
// this request log only show once, is also say the retry is sucessful
write incomplete, request_size=2816868364, total_write_size=2147479552, this_write_size=2147479552, and will retry it.
```
however maybe since the log is too large so that the `tcmalloc` report problem:
```
asio write faitcmalloc: large alloc 3501096960 bytes == 0x7570e8000 @  0x7faab64b8e9b 0x7faab64dc8cb 0x4c42c2 0x7faab92ca929 0x7faab92cec18 0x7faab92cee04 0x7faab92ccf39 0x7faab8f25aeb 0x7faab8f3acf9 0x7faab8f436b1 0x7faab8f43edc 0x7faab92cac3b 0x7faab9232581 0x7faab924c4fa 0x7faab924c6fd 0x7faab4fbffd0
led: Connection reset by peer
```  
But this is another problem and not the pr target.

src/aio/native_linux_aio_provider.cpp
```diff
@@ -68,15 +70,43 @@ error_code native_linux_aio_provider::flush(dsn_handle_t fh)
...
+            }
+            resp = ERR_FILE_OPERATION_FAILED;
+            derror_f("write failed with errno={}, return {}.", strerror(errno), resp);
+            return resp;
+        }
+
+        // mock the `ret` to reproduce the `write incomplete` case in the first write
+        FAIL_POINT_INJECT_OFF_F("aio_pwrite_incomplete", [&]() -> void {
+            if (dsn_unlikely(buffer_offset == 0)) {
+                --ret;
```
zhangyifan27, line 94: I'm curious about why the test would pass if `--ret` every loop?

Shuo-Jia: ` --ret` only run at first time (`dsn_unlikely(buffer_offset == 0`))

src/aio/test/aio.cpp
```diff
@@ -143,6 +147,9 @@ TEST(core, aio_share)
 
 TEST(core, operation_failed)
 {
+    fail::setup();
+    fail::cfg("aio_pwrite", "off()");
```
zhangyifan27, line 151: "aio_pwrite_incomplete"

zhangyifan27 (approved):

levy5307 (dismissed):

levy5307 (approved):

zhangyifan27 (approved):

PR #823, opened 2021-04-22
fix(bulk_load): fix remove_local_bulk_load_dir()

zhangyifan27:
Recently our users encountered problems while using bulk_load, their job always failed, and we found remove_directory errors in replica server log:
```
W2021-04-22 11:15:40.633 (1619061340633182759 122434) replica.replica20.0400de080018c494: filesystem.cpp:329:remove_directory(): remove /home/work/ssd3/pegasus/c4tst-bulkload/replica/reps/46.18.pegasus/bulk_load failed, err = Directory not empty
E2021-04-22 11:15:40.633 (1619061340633219926 122434) replica.replica20.0400de080018c494: replica_bulk_loader.cpp:599:remove_local_bulk_load_dir(): [46.18@10.132.5.22:34801] remove bulk_load dir(/home/work/ssd3/pegasus/c4tst-bulkload/replica/reps/46.18.pegasus/bulk_load) failed
```
The reason is that we couldn't remove a directory using whiling writing files in this directory(see the added tests in HDFSClientTest). This patch rename unused bulkload dir to a garbage dir instead of just removing it, then remove garbage dir, disk cleaner would retry to remove it if failed.

### DIFF
--- a/src/replica/bulk_load/replica_bulk_loader.cpp
+++ b/src/replica/bulk_load/replica_bulk_loader.cpp
@@ -392,11 +392,8 @@ error_code replica_bulk_loader::download_sst_files(const std::string &remote_dir
     }
     const std::string local_dir = utils::filesystem::path_combine(
         _replica->_dir, bulk_load_constant::BULK_LOAD_LOCAL_ROOT_DIR);
-    if (utils::filesystem::directory_exists(local_dir)) {
-        // remove old files before downloading sst files.
-        remove_local_bulk_load_dir(local_dir);
-    }
-    if (!utils::filesystem::create_directory(local_dir)) {
+    if (!utils::filesystem::directory_exists(local_dir) &&
+        !utils::filesystem::create_directory(local_dir)) {
         derror_replica("create bulk_load_dir({}) failed", local_dir);
         return ERR_FILE_OPERATION_FAILED;
     }
--- a/src/replica/bulk_load/replica_bulk_loader.cpp
+++ b/src/replica/bulk_load/replica_bulk_loader.cpp
@@ -610,10 +610,13 @@ void replica_bulk_loader::remove_local_bulk_load_dir(const std::string &bulk_loa
                                           bulk_load_constant::BULK_LOAD_LOCAL_ROOT_DIR,
                                           std::to_string(dsn_now_ms()),
                                           kFolderSuffixGar);
-    utils::filesystem::rename_path(bulk_load_dir, garbage_dir);
+    if (!utils::filesystem::rename_path(bulk_load_dir, garbage_dir)) {
+        derror_replica("rename bulk_load dir({}) failed.", bulk_load_dir);
+        return;
+    }
     if (!utils::filesystem::remove_path(garbage_dir)) {
         derror_replica(
-            "remove bulk_load gar dir({}) failed, disk cleaner would retry to remote it.",
+            "remove bulk_load gar dir({}) failed, disk cleaner would retry to remove it.",
             garbage_dir);
     }
 }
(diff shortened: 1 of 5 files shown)


Shuo-Jia:
> before starting to download bulkload

I don't see `remove` before starting? is it not in this pr?

src/replica/bulk_load/replica_bulk_loader.cpp
```diff
@@ -592,27 +593,31 @@ void replica_bulk_loader::handle_bulk_load_finish(bulk_load_status::type new_sta
...
+void replica_bulk_loader::remove_local_bulk_load_dir(const std::string &bulk_load_dir)
 {
-    if (!utils::filesystem::directory_exists(bulk_load_dir) ||
-        !utils::filesystem::remove_path(bulk_load_dir)) {
-        derror_replica("remove bulk_load dir({}) failed", bulk_load_dir);
-        return ERR_FILE_OPERATION_FAILED;
+    if (!utils::filesystem::directory_exists(bulk_load_dir)) {
+        return;
+    }
+    // rename bulk_load_dir to replica_dir.bulkload.timestamp.gar.
```
Shuo-Jia, line 606: I think can add comment to explain why add `rename` before `remove`. otherwise the  `remove` logic may be strange for user

zhangyifan27:
> > before starting to download bulkload
> 
> I don't see `remove` before starting? isn't it in this pr?

Because `do_bulkload()`, `start_bulkload()` would be executed several times during the bulkload progress, so it is not safe to remove the `bulk_load_dir` in the begining of any function. I edited the description.

src/replica/bulk_load/replica_bulk_loader.cpp
```diff
@@ -592,27 +593,33 @@ void replica_bulk_loader::handle_bulk_load_finish(bulk_load_status::type new_sta
...
+    }
+    // Rename bulk_load_dir to replica_dir.bulkload.timestamp.gar before remove it.
+    // Because we download sst files asynchronously and couldn't remove a directory while writing
+    // files in it.
+    std::string garbage_dir = fmt::format("{}.{}.{}.{}",
+                                          _replica->_dir,
+                                          bulk_load_constant::BULK_LOAD_LOCAL_ROOT_DIR,
+                                          std::to_string(dsn_now_ms()),
+                                          kFolderSuffixGar);
+    if (!utils::filesystem::rename_path(bulk_load_dir, garbage_dir)) {
```
Shuo-Jia, line 614: if rename failed, the bug still will be existed? do you test the rename will be failed in `asynchronously wrting`

zhangyifan27: Yes, I ran the test 100 times and it always passed. If rename failed, there seems to be no good solution now.

src/replica/bulk_load/replica_bulk_loader.cpp
```diff
@@ -592,27 +593,33 @@ void replica_bulk_loader::handle_bulk_load_finish(bulk_load_status::type new_sta
...
+void replica_bulk_loader::remove_local_bulk_load_dir(const std::string &bulk_load_dir)
 {
-    if (!utils::filesystem::directory_exists(bulk_load_dir) ||
-        !utils::filesystem::remove_path(bulk_load_dir)) {
-        derror_replica("remove bulk_load dir({}) failed", bulk_load_dir);
-        return ERR_FILE_OPERATION_FAILED;
+    if (!utils::filesystem::directory_exists(bulk_load_dir)) {
+        return;
+    }
+    // Rename bulk_load_dir to replica_dir.bulkload.timestamp.gar before remove it.
```
Shuo-Jia, line 606: `bulkload`->`bulk_load`

Shuo-Jia (approved):

hycdong (approved):

PR #826, opened 2021-04-26
fix(bulk_load): redownload file if local file damaged

hycdong:
In bulk load download stage, replica server will download file from remote file provider, if a same-name file exists,  block service interface `download_file` will return `ERR_PATH_ALREADY_EXIST`. In previous implmentation, if local file doesn't pass verification, it will lead to bulk load failed. This pull request fixs it, if local file doesn't pass verifition, it will remove old file and redownload it.

I test this case in onebox manually in following step:
1. start bulk load
2. pause bulk load
3. modify one file content to simulate downloaded local file is different from remoted file
4. restart bulk load
5. I can find redownload log below
```
W2021-04-26 15:38:04.26 (1619422684026147118 7747) replica.rep_long0.0304000000000420: replica_bulk_loader.cpp:441:operator()(): [2.0@<ip>:<port>] file(<path>/file0.sst) exists, but not verified, try to remove and redownload it
```
6. bulk load succeed

### DIFF
--- a/src/replica/bulk_load/replica_bulk_loader.cpp
+++ b/src/replica/bulk_load/replica_bulk_loader.cpp
@@ -436,12 +436,12 @@ error_code replica_bulk_loader::download_sst_files(const std::string &remote_dir
                         f_size = f_meta.size;
                         verified = true;
                     } else {
-                        dwarn_replica(
-                            "file({}) exists, but not verified, try to remove and redownload it",
+                        derror_replica(
+                            "file({}) exists, but not verified, try to remove local file "
+                            "and redownload it",
                             file_name);
-                        // remove unverified local file and redownload
                         if (!utils::filesystem::remove_path(file_name)) {
-                            derror_f("failed to remove file({})", file_name);
+                            derror_replica("failed to remove file({})", file_name);
                             ec = ERR_FILE_OPERATION_FAILED;
                         } else {
                             ec = _stub->_block_service_manager.download_file(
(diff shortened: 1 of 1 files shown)


src/replica/bulk_load/replica_bulk_loader.cpp
```diff
@@ -428,15 +428,31 @@ error_code replica_bulk_loader::download_sst_files(const std::string &remote_dir
...
+                        derror_replica(
+                            "file({}) exists, but not verified, try to remove local file "
+                            "and redownload it",
+                            file_name);
+                        if (!utils::filesystem::remove_path(file_name)) {
+                            derror_replica("failed to remove file({})", file_name);
+                            ec = ERR_FILE_OPERATION_FAILED;
+                        } else {
+                            ec = _stub->_block_service_manager.download_file(
+                                remote_dir, local_dir, f_meta.name, fs, f_size);
```
levy5307, line 448: remove the `verify_file` below to this line

hycdong: The following cases should verify files:
- ec = ERR_OK
- ec = ERR_PATH_ALREADY_EXIST
- ec = ERR_PATH_ALREADY_EXIST, local file removed, redownload ec = ERR_OK

If `verify_file` move below this line, `ec = ERR_OK` won't be checked

levy5307: But in your code, ec == ERR_OK don't  call the `verify_file` function :
```
                if (ec == ERR_OK && !verified &&
                    !utils::filesystem::verify_file(file_name, f_meta.md5, f_meta.size)) {
                    ec = ERR_CORRUPTION;
                }
```
Only ec == ERR_OK and !verifyed does.

hycdong: Acutually, the possible cases is like below:
1. ec = ERR_OK, verified = false
2. ec = ERR_PATH_ALREADY_EXIST and pass verfication -> ec = ERR_OK, verified = true
3. ec = ERR_PATH_ALREADY_EXIST, local file removed, redownload ec = ERR_OK, verified = false

I add `verified` only to avoid case2 doing verification twice/

neverchanje (approved):

levy5307 (approved):

PR #827, opened 2021-04-28
feat(bulk_load): add bulk load max rollback count

hycdong:
During bulk load process, meta server will control the bulk load status switch: downloading -> downloaded -> ingesting -> succeed, if replica configuration changed or network failed, meta server will rollback status into downloading. In pervious implementation, it is possible for bulk load process rollback for long time. This pull request adds `bulk_load_max_rollback_times`, if the table has been rollback to downloading for too many times, it will turn status into failed to stop the process.

configuration update:
```diff
[meta_server]
+bulk_load_max_rollback_times = 10
```

### DIFF
--- a/src/meta/meta_bulk_load_service.cpp
+++ b/src/meta/meta_bulk_load_service.cpp
@@ -17,11 +17,22 @@
 
 #include <dsn/dist/fmt_logging.h>
 #include <dsn/utility/fail_point.h>
+#include <dsn/utility/flags.h>
 
 #include "meta_bulk_load_service.h"
 
 namespace dsn {
 namespace replication {
+
+DSN_DEFINE_uint32("meta_server",
+                  bulk_load_max_failover_count,
+                  20,
+                  "if bulk load failover count "
+                  "exceed this value, meta won't "
+                  "rollback bulk load process to "
+                  "downloading, but turn it into "
+                  "failed");
+DSN_TAG_VARIABLE(bulk_load_max_failover_count, FT_MUTABLE);
 
 bulk_load_service::bulk_load_service(meta_service *meta_svc, const std::string &bulk_load_dir)
     : _meta_svc(meta_svc), _state(meta_svc->get_server_state()), _bulk_load_root(bulk_load_dir)
@@ -231,6 +242,7 @@ void bulk_load_service::create_app_bulk_load_dir(const std::string &app_name,
                 zauto_write_lock l(_lock);
                 _app_bulk_load_info[ainfo.app_id] = ainfo;
                 _apps_pending_sync_flag[ainfo.app_id] = false;
+                _apps_failover_count[ainfo.app_id] = 0;
             }
             for (int32_t i = 0; i < ainfo.partition_count; ++i) {
                 create_partition_bulk_load_dir(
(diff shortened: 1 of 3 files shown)


src/meta/meta_bulk_load_service.cpp
```diff
@@ -1654,6 +1679,7 @@ void bulk_load_service::do_continue_app_bulk_load(
     {
         zauto_write_lock l(_lock);
         _apps_in_progress_count[app_id] = in_progress_partition_count;
+        _apps_failover_count[app_id] = 0;
```
levy5307, line 1682: Why did you set _apps_failover_count[app_id] to 0 when you continue to bulk load? I think you can reserve the old value.

hycdong: `do_continue_app_bulk_load` is called when meta server crash and restart, find bulk load context on zk and continue to do bulk load. In this case, `_apps_failover_count[app_id]` should be initialized as 0.

levy5307 (dismissed):

Smityz:
LGTM

src/meta/meta_bulk_load_service.cpp
```diff
@@ -17,12 +17,23 @@
...
 #include <dsn/utility/fail_point.h>
+#include <dsn/utility/flags.h>
 
 #include "meta_bulk_load_service.h"
 
 namespace dsn {
 namespace replication {
 
+DSN_DEFINE_uint32("meta_server",
+                  bulk_load_max_failover_count,
```
neverchanje, line 28:
```suggestion
                  bulk_load_max_rollback_times,
```

I prefer to call it retry_times or rollback times. Because in your description you call it "rollback", it's inconsistent to name it "failover".

src/meta/meta_bulk_load_service.cpp
```diff
@@ -17,12 +17,23 @@
...
+                  20,
```
neverchanje, line 29: 20 will still be too many if the job is pretty large and the disk will easily get burst out.
I would recommend 5 or 10.

src/meta/meta_bulk_load_service.h
```diff
@@ -404,6 +406,8 @@ class bulk_load_service
     std::unordered_map<app_id, bool> _apps_cleaning_up;
     // Used for bulk load rolling back to downloading
     std::unordered_map<app_id, bool> _apps_rolling_back;
+    // Used for restrict bulk load rollback count
+    std::unordered_map<app_id, int32_t> _apps_rollback_count;
```
Shuo-Jia, line 410: seem to only need `_apps_rollback_count`,  `_apps_rolling_back` can be replaced by it

hycdong: They have different usage, `_apps_rollback_count` records the rollback count to avoid rollback too many times, `_apps_rolling_back` records if the app is executing rollback to avoid rollback at the same time.

Shuo-Jia: you can also judge if the app is rolling back by   `_apps_rollback_count[app_id]` result

hycdong: `_apps_rollback_count` can't replace `_apps_rolling_back`. Let's consume the situtation, replica server A crash during bulk load, partition0 and partition1 on it. If meta server got partition0 timeout first, meta sever will set the app rollback to downloading, then partition1 will raise it again, always happened when this app is executing rolling back, `_apps_rolling_back` is used to solve this condition. 
Well, `_apps_rollback_count` can't do that. If the current count = 2, count=3 indicating app is rolling back, meta server can't judge the rolling back action through this value.

Shuo-Jia (approved):

levy5307 (approved):

PR #829, opened 2021-04-30
feat(bulk_load): meta server adds bulk load ingestion concurrent count restriction

hycdong:
To shorten all partitions ingestion total time, the pervious implementation didn't do any restrction when meta server sent ingestion request to primaries. However, ingestion will cost replica server lots of cpu, io resource, even affect cluster read/write performance. This pull request adds ingestion concurrent count in meta server, if the count is 4, the fifth ingestion request won't be sent to primary partition unless meta received one ingestion response. Besides, this count can be updated dynamically through http interface.

```diff
[meta_server]
+bulk_load_ingestion_concurrent_count=4
```

### DIFF
--- a/src/meta/meta_bulk_load_service.cpp
+++ b/src/meta/meta_bulk_load_service.cpp
@@ -1004,6 +1004,7 @@ void bulk_load_service::update_app_status_on_remote_storage_reply(const app_bulk
         _app_bulk_load_info[app_id] = ainfo;
         _apps_pending_sync_flag[app_id] = false;
         _apps_in_progress_count[app_id] = partition_count;
+        // when rollback from ingesting, ingesting_count should be reset
         if (old_status == bulk_load_status::BLS_INGESTING &&
             new_status == bulk_load_status::BLS_DOWNLOADING) {
             _apps_ingesting_count[app_id] = 0;
--- a/src/meta/meta_bulk_load_service.cpp
+++ b/src/meta/meta_bulk_load_service.cpp
@@ -1059,6 +1059,15 @@ void bulk_load_service::partition_ingestion(const std::string &app_name, const g
         }
     }
 
+    auto app_status = get_app_bulk_load_status(pid.get_app_id());
+    if (app_status != bulk_load_status::BLS_INGESTING) {
+        dwarn_f("app({}) current status is {}, partition({}), ignore it",
+                app_name,
+                dsn::enum_to_string(app_status),
+                pid);
+        return;
+    }
+
     rpc_address primary_addr;
     {
         zauto_read_lock l(app_lock());
@@ -1071,15 +1080,6 @@ void bulk_load_service::partition_ingestion(const std::string &app_name, const g
             return;
         }
         primary_addr = app->partitions[pid.get_partition_index()].primary;
-    }
-
-    auto app_status = get_app_bulk_load_status(pid.get_app_id());
-    if (app_status != bulk_load_status::BLS_INGESTING) {
-        dwarn_f("app({}) current status is {}, partition({}), ignore it",
-                app_name,
-                dsn::enum_to_string(app_status),
-                pid);
-        return;
     }
 
     if (primary_addr.is_invalid()) {
--- a/src/meta/meta_bulk_load_service.cpp
+++ b/src/meta/meta_bulk_load_service.cpp
@@ -1073,6 +1073,15 @@ void bulk_load_service::partition_ingestion(const std::string &app_name, const g
         primary_addr = app->partitions[pid.get_partition_index()].primary;
     }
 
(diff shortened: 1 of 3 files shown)


src/meta/meta_bulk_load_service.cpp
```diff
@@ -1001,10 +1009,32 @@ void bulk_load_service::update_app_status_on_remote_storage_reply(const app_bulk
...
+                    "wait and try it later",
+                    app_name,
+                    _apps_ingesting_count[pid.get_app_id()],
+                    pid);
+            tasking::enqueue(
+                LPC_META_STATE_NORMAL,
+                _meta_svc->tracker(),
+                std::bind(&bulk_load_service::partition_ingestion, this, app_name, pid),
+                pid.thread_hash(),
+                std::chrono::seconds(1));
```
Shuo-Jia, line 1034: I think can be longer time, 5 or 10s?

src/meta/meta_bulk_load_service.cpp
```diff
@@ -1120,6 +1160,7 @@ void bulk_load_service::on_partition_ingestion_reply(error_code err,
                  pid,
                  resp.err,
                  resp.rocksdb_error);
+        decrease_app_ingestion_count(pid);
```
Shuo-Jia, line 1163: I understand it must `decrease_app_ingestion_count` if `err_code != ok`?  so you just:
```
if != ok {
 decrease_app_ingestion_count
}
```
and put it start line of this method

src/meta/meta_bulk_load_service.cpp
```diff
@@ -995,6 +1004,10 @@ void bulk_load_service::update_app_status_on_remote_storage_reply(const app_bulk
         _app_bulk_load_info[app_id] = ainfo;
         _apps_pending_sync_flag[app_id] = false;
         _apps_in_progress_count[app_id] = partition_count;
+        if (old_status == bulk_load_status::BLS_INGESTING &&
+            new_status == bulk_load_status::BLS_DOWNLOADING) {
+            _apps_ingesting_count[app_id] = 0;
+        }
```
Shuo-Jia, line 1010: is this  `rollback` case? suggest comment, otherwise the people may be confused

src/meta/meta_bulk_load_service.cpp
```diff
@@ -1713,10 +1756,7 @@ void bulk_load_service::do_continue_app_bulk_load(
         gpid pid = gpid(app_id, i);
         partition_bulk_load(ainfo.app_name, pid);
         if (app_status == bulk_load_status::BLS_INGESTING) {
-            tasking::enqueue(
-                LPC_BULK_LOAD_INGESTION,
-                _meta_svc->tracker(),
-                std::bind(&bulk_load_service::partition_ingestion, this, ainfo.app_name, pid));
+            partition_ingestion(ainfo.app_name, pid);
```
Shuo-Jia, line 1768: Will this method be triggred periodically？ like `node balalance check`? if only execute after `update status` or `last response replay`, when one send failed, can we make sure all the replica can receive `ingest request`?

src/meta/meta_bulk_load_service.cpp
```diff
@@ -1024,10 +1032,41 @@ void bulk_load_service::update_app_status_on_remote_storage_reply(const app_bulk
...
+                _meta_svc->tracker(),
+                std::bind(&bulk_load_service::partition_ingestion, this, app_name, pid),
+                pid.thread_hash(),
+                std::chrono::seconds(5));
+            return;
+        }
+    }
+
+    auto app_status = get_app_bulk_load_status(pid.get_app_id());
+    if (app_status != bulk_load_status::BLS_INGESTING) {
```
Shuo-Jia, line 1063: ? isn't = bulk_load_status::BLS_INGESTING, so that we ingnore it?

hycdong: Function `partition_ingestion` should only be called when app bulk load status is INGESTING. As we adds count restriction, and the delayed request will be resent after 5 seconds, app status may be changed.

Shuo-Jia: Oh, however, if one replica has been `INGESTING` status, shouldn't we ingnore it?

Shuo-Jia: Oh, I see, `get_app_bulk_load_status` is getting meta local status?

hycdong: If all partitions' status is INGESTING, app status will be INGESTING. `get_app_bulk_load_status` is to get app status in meta server locally.

Shuo-Jia (approved):

neverchanje (approved):

levy5307 (approved):

PR #831, opened 2021-05-08
refactor: refactor replication_options initialize function

hycdong:
This pull request refactor part of function `replication_options::initialize`. This function will get options from config and store them including replica data directory (data_dirs). To get data_dirs, `initialize` will first parse `config[data_dirs]`, then read black_list file, finally exclude the directories in black list. This pr add three functions to implement them, called: `get_data_dir_and_tag`, `get_data_dirs_in_black_list` and `check_if_in_black_list`, besides, add unit tests for those functions.

### DIFF
--- a/src/common/replication_common.cpp
+++ b/src/common/replication_common.cpp
@@ -580,7 +580,6 @@ replication_options::get_data_dirs_in_black_list(const std::string &fname,
         dirs.push_back(str2);
         count++;
         ddebug_f("black_list[{}] = [{}]", count, str2);
-        str.clear();
     }
 }
 
@@ -588,7 +587,6 @@ replication_options::get_data_dirs_in_black_list(const std::string &fname,
 replication_options::check_if_in_black_list(const std::vector<std::string> &black_list_dir,
                                             const std::string &dir)
 {
-    bool in_black_list = false;
     std::string dir_str = dir;
     if (!black_list_dir.empty()) {
         if (dir_str.back() != '/') {
@@ -596,12 +594,11 @@ replication_options::check_if_in_black_list(const std::vector<std::string> &blac
         }
         for (const std::string &black : black_list_dir) {
             if (dir_str.find(black) == 0) {
-                in_black_list = true;
-                break;
+                return true;
             }
         }
     }
-    return in_black_list;
+    return false;
 }
 
 const std::string backup_restore_constant::FORCE_RESTORE("restore.force_restore");
--- a/src/common/replication_common.cpp
+++ b/src/common/replication_common.cpp
@@ -571,10 +571,12 @@ replication_options::get_data_dirs_in_black_list(const std::string &fname,
     int count = 0;
     while (std::getline(file, str)) {
         std::string str2 = utils::trim_string(const_cast<char *>(str.c_str()));
-        if (str2.empty())
+        if (str2.empty()) {
             continue;
-        if (str2.back() != '/')
+        }
+        if (str2.back() != '/') {
             str2.append("/");
+        }
         dirs.push_back(str2);
         count++;
         ddebug_f("black_list[{}] = [{}]", count, str2);
--- a/src/common/replication_common.cpp
+++ b/src/common/replication_common.cpp
@@ -556,7 +556,7 @@ replication_options::get_data_dir_and_tag(const std::string &config_dirs_str,
 replication_options::get_data_dirs_in_black_list(const std::string &fname,
                                                  /*out*/ std::vector<std::string> &dirs)
 {
-    if (fname.empty() || (!fname.empty() && !utils::filesystem::file_exists(fname))) {
+    if (fname.empty() || !utils::filesystem::file_exists(fname)) {
         ddebug_f("data_dirs_black_list_file[{}] not found, ignore it", fname);
         return;
     }
--- a/src/common/replication_common.cpp
+++ b/src/common/replication_common.cpp
@@ -553,7 +553,7 @@ replication_options::get_data_dir_and_tag(const std::string &config_dirs_str,
     }
 
     for (unsigned i = 0; i < dirs.size(); ++i) {
-        std::string &dir = dirs[i];
+        const std::string &dir = dirs[i];
         ddebug_f("data_dirs[{}] = {}, tag = {}", i + 1, dir, dir_tags[i]);
         data_dirs.push_back(utils::filesystem::path_combine(dir, "reps"));
         data_dir_tags.push_back(dir_tags[i]);
@@ -570,7 +570,7 @@ replication_options::get_data_dirs_in_black_list(const std::string &fname,
         return;
     }
 
-    ddebug_f("data_dirs_black_list_file[{}] found, apply it", fname.c_str());
+    ddebug_f("data_dirs_black_list_file[{}] found, apply it", fname);
     std::ifstream file(fname);
     if (!file) {
         dassert_f(false, "open data_dirs_black_list_file failed: {}", fname);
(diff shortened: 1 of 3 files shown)


src/common/replication_common.cpp
```diff
@@ -569,6 +498,118 @@ void replica_helper::load_meta_servers(/*out*/ std::vector<dsn::rpc_address> &se
...
+/*static*/ void
+replication_options::get_data_dirs_in_black_list(const std::string &fname,
+                                                 /*out*/ std::vector<std::string> &dirs)
+{
+    if (fname.empty() || (!fname.empty() && !utils::filesystem::file_exists(fname))) {
+        ddebug_f("data_dirs_black_list_file[{}] not found, ignore it", fname);
+        return;
+    }
+
+    ddebug_f("data_dirs_black_list_file[{}] found, apply it", fname.c_str());
```
levy5307, line 573: ddebug_f("data_dirs_black_list_file[{}] found, apply it", fname);

src/common/replication_common.cpp
```diff
@@ -569,6 +498,118 @@ void replica_helper::load_meta_servers(/*out*/ std::vector<dsn::rpc_address> &se
...
+    std::ifstream file(fname);
+    if (!file) {
+        dassert_f(false, "open data_dirs_black_list_file failed: {}", fname);
+    }
+
+    std::string str;
+    int count = 0;
+    while (std::getline(file, str)) {
+        std::string str2 = utils::trim_string((char *)str.c_str());
```
levy5307, line 582: static_cast<char*>

src/common/replication_common.cpp
```diff
@@ -569,6 +498,118 @@ void replica_helper::load_meta_servers(/*out*/ std::vector<dsn::rpc_address> &se
...
+                        "dir({}) and dir({}) have same tag({})", dirs[i], dir, tag_and_dir[0]);
+                    return false;
+                }
+            }
+            dir_tags.push_back(tag_and_dir[0]);
+        }
+    }
+
+    for (unsigned i = 0; i < dirs.size(); ++i) {
+        std::string &dir = dirs[i];
```
levy5307, line 556: const std::string&

src/common/replication_common.cpp
hycdong: This pr is for refactor, so I copy old code into the new functions. I will update it.

acelyc111 (dismissed):

Smityz:
LGTM

src/common/replication_common.cpp
```diff
@@ -569,6 +498,118 @@ void replica_helper::load_meta_servers(/*out*/ std::vector<dsn::rpc_address> &se
...
 
+/*static*/ bool
+replication_options::get_data_dir_and_tag(const std::string &config_dirs_str,
+                                          const std::string &default_dir,
+                                          const std::string &app_name,
+                                          /*out*/ std::vector<std::string> &data_dirs,
+                                          /*out*/ std::vector<std::string> &data_dir_tags,
+                                          /*out*/ std::string &err_msg)
+{
+    // - if {config_dirs_str} is empty (return true):
```
zhangyifan27, line 509: I think maybe it will be better to make it a `void` function, if some configurations are invalid, just `assert` and return. That will be easier to understand and no need for complicated comments and conditional judgments.

hycdong: Yes, I agree that make it a `void` function is simple. I used to implement it as a `void` function, but I found that it was hard to write a unit test for this function. Finally, I update my implmentation.

zhangyifan27: Maybe the gtest macro `ASSERT_NO_FATAL_FAILURE` is useful for assertion tests.

hycdong: I tried to use `ASSERT_NO_FATAL_FAILURE`, but it seems not work and still core when `dassert`. I have read the usage of `ASSERT_NO_FATAL_FAILURE`
```
// Macros that execute statement and check that it doesn't generate new fatal failures in the current thread
```
It seems that using `ASSERT_NO_FATAL_FAILURE` can not avoid assert in source code.

levy5307 (dismissed):

src/common/replication_common.cpp
```diff
@@ -569,6 +498,118 @@ void replica_helper::load_meta_servers(/*out*/ std::vector<dsn::rpc_address> &se
...
+        if (dir_str.back() != '/')
+            dir_str.append("/");
+        for (const std::string &black : black_list_dir) {
+            if (dir_str.find(black) == 0) {
+                in_black_list = true;
+                break;
+            }
+        }
+    }
+    return in_black_list;
```
acelyc111, line 601:
```suggestion
    std::string dir_str = dir;
    if (!black_list_dir.empty()) {
        if (dir_str.back() != '/')
            dir_str.append("/");
        for (const std::string &black : black_list_dir) {
            if (dir_str.find(black) == 0) {
                return true;
            }
        }
    }
    return false;
```

src/common/replication_common.cpp
```diff
@@ -569,6 +498,118 @@ void replica_helper::load_meta_servers(/*out*/ std::vector<dsn::rpc_address> &se
...
+        data_dir_tags.push_back(dir_tags[i]);
+    }
+    return true;
+}
+
+/*static*/ void
+replication_options::get_data_dirs_in_black_list(const std::string &fname,
+                                                 /*out*/ std::vector<std::string> &dirs)
+{
+    if (fname.empty() || (!fname.empty() && !utils::filesystem::file_exists(fname))) {
```
acelyc111, line 559:
```suggestion
    if (fname.empty() || !utils::filesystem::file_exists(fname)) {
```

src/common/replication_common.cpp
```diff
@@ -569,6 +498,118 @@ void replica_helper::load_meta_servers(/*out*/ std::vector<dsn::rpc_address> &se
...
+    while (std::getline(file, str)) {
+        std::string str2 = utils::trim_string(const_cast<char *>(str.c_str()));
+        if (str2.empty())
+            continue;
+        if (str2.back() != '/')
+            str2.append("/");
+        dirs.push_back(str2);
+        count++;
+        ddebug_f("black_list[{}] = [{}]", count, str2);
+        str.clear();
```
acelyc111, line 581: Is it needed?

src/common/replication_common.cpp
```diff
@@ -569,6 +495,112 @@ void replica_helper::load_meta_servers(/*out*/ std::vector<dsn::rpc_address> &se
...
+
+/*static*/ bool
+replication_options::check_if_in_black_list(const std::vector<std::string> &black_list_dir,
+                                            const std::string &dir)
+{
+    bool in_black_list = false;
+    std::string dir_str = dir;
+    if (!black_list_dir.empty()) {
+        if (dir_str.back() != '/')
+            dir_str.append("/");
```
acelyc111, line 593:
```suggestion
        if (dir_str.back() != '/') {
            dir_str.append("/");
        }
```

src/common/replication_common.cpp
```diff
@@ -569,6 +495,112 @@ void replica_helper::load_meta_servers(/*out*/ std::vector<dsn::rpc_address> &se
...
+    }
+
+    std::string str;
+    int count = 0;
+    while (std::getline(file, str)) {
+        std::string str2 = utils::trim_string(const_cast<char *>(str.c_str()));
+        if (str2.empty())
+            continue;
+        if (str2.back() != '/')
+            str2.append("/");
```
acelyc111, line 577:
```suggestion
        if (str2.empty()) {
            continue;
        }
        if (str2.back() != '/') {
            str2.append("/");
        }
```

acelyc111 (approved):
LGTM

levy5307 (approved):

PR #833, opened 2021-05-17
feat(disk): reject write if disk space is insufficient

hycdong:
This pull request is about disk protection when space is almost full. If one disk available space is below `disk_min_available_space_ratio`, all replica on this disk will reject user write request. The disk status is updated periodically, when disk space is normal, user write request will be recovered.

This pull request also adds related unit tests and dynamical configuration.
```diff
[replication]
+enable_disk_available_space_check=true
+disk_min_available_space_ratio=10
```


src/client/partition_resolver.cpp
```diff
@@ -53,7 +53,8 @@ void partition_resolver::call_task(const rpc_response_task_ptr &t)
         dsn::error_code err, dsn::message_ex * req, dsn::message_ex * resp)
     {
         if (req->header->gpid.value() != 0 && err != ERR_OK && err != ERR_HANDLER_NOT_FOUND &&
-            err != ERR_APP_NOT_EXIST && err != ERR_OPERATION_DISABLED && err != ERR_BUSY) {
```
levy5307, line 56: you can put all of there error code into a contianer, and see whether it's in the container or not

src/common/fs_manager.cpp
```diff
@@ -75,23 +89,40 @@ unsigned dir_node::remove(const gpid &pid)
     return iter->second.erase(pid);
 }
 
-void dir_node::update_disk_stat()
+void dir_node::update_disk_stat(bool &status_changed)
```
levy5307, line 92: ```bool dir_node::update_disk_stat()```

src/replica/test/replica_disk_test_base.h
```diff
@@ -95,6 +97,32 @@ class replica_disk_test_base : public replica_test_base
         }
     }
 
+    void
+    mock_node_status(int32_t node_index, disk_status::type old_status, disk_status::type new_status)
+    {
+        auto node = get_dir_nodes()[node_index];
+        for (auto &kv : node->holding_replicas) {
```
levy5307, line 104: const auto &

src/common/fs_manager.cpp
hycdong: I used to clear function like your suggestion. If return false, it seems like update disk stat failed, so I pass `status_chaned` instead.

acelyc111 (approved):

levy5307 (approved):

PR #834, opened 2021-05-17
feat(disk): add broken disk check while initialize

hycdong:
This pull request adds disk check while initialize. We choose the way that doris used. When the disk path is found, we will try to write and read data to check if the disk is broken. If the disk is broken, it will be excluded and replica instance can start. In previous implmenetation, if a disk is broken, replica server will core and never start until the broken disk removed.

This pull request also adds related unit tests and configuration.
```diff
[replication]
+ignore_broken_disk=true
```


src/replica/replica_stub.cpp
```diff
@@ -789,6 +775,40 @@ void replica_stub::initialize(const replication_options &opts, bool clear /* = f
...
+    int count = 0;
+    std::vector<std::string> available_dirs;
+    std::vector<std::string> available_dir_tags;
+    for (int i = 0; i < data_dir_tags.size(); ++i) {
+        std::string &dir = data_dirs[i];
+        bool flag = dsn::utils::filesystem::create_directory(dir, cdir, err_msg);
+        if (flag) {
+            flag = dsn::utils::filesystem::check_dir_rw(dir, err_msg);
+        }
+        if (!flag) {
```
acelyc111, line 792:
```suggestion
        if (!utils::filesystem::create_directory(dir, cdir, err_msg)
            || !utils::filesystem::check_dir_rw(dir, err_msg)) {
```

src/utils/filesystem.cpp
```diff
@@ -826,6 +827,78 @@ bool verify_file(const std::string &fname,
...
+    }
+
+    std::string value = "test_value";
+    if (!write_file(fpath, value)) {
+        err_msg = fmt::format("Fail to write file {}.", fpath);
+        return false;
+    }
+
+    std::string buf;
+    if (read_file(fpath, buf) != ERR_OK) {
```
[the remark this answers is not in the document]
hycdong: Done

acelyc111 (dismissed):
LGTM

src/replica/replica_stub.cpp
```diff
@@ -789,6 +775,37 @@ void replica_stub::initialize(const replication_options &opts, bool clear /* = f
...
 }
 
+void replica_stub::initialize_fs_manager(std::vector<std::string> &data_dirs,
+                                         std::vector<std::string> &data_dir_tags)
+{
+    std::string cdir;
+    std::string err_msg;
+    int count = 0;
+    std::vector<std::string> available_dirs;
+    std::vector<std::string> available_dir_tags;
```
Smityz, line 785: maybe you should assert data_dirs.size() == data_dir_tags.size() here

hycdong: data_dirs and data_dirs_tags are `_options.data_dirs` and `_options.data_dir_tags`, you can reference how them set in `replication_common.cpp` function `initialize`, I don't add a link here because it will be refactored by pull request 831.

acelyc111 (dismissed):

src/replica/replica_stub.cpp
```diff
@@ -790,6 +776,37 @@ void replica_stub::initialize(const replication_options &opts, bool clear /* = f
...
+    for (int i = 0; i < data_dir_tags.size(); ++i) {
+        std::string &dir = data_dirs[i];
+        if (!utils::filesystem::create_directory(dir, cdir, err_msg) ||
+            !utils::filesystem::check_dir_rw(dir, err_msg)) {
```
levy5307, line 790: dsn_unlikely

src/utils/filesystem.cpp
```diff
@@ -826,6 +827,80 @@ bool verify_file(const std::string &fname,
...
+    FAIL_POINT_INJECT_F("filesystem_check_dir_rw", [path](string_view str) {
+        // when str contains 'false', and path contains broken_disk_dir, mock check fail(return
+        // false)
+        std::string broken_disk_dir = "disk1";
+        return str.find("false") == string_view::npos ||
+               path.find(broken_disk_dir) == std::string::npos;
+    });
+
+    std::string fname = "read_write_test_file";
+    std::string fpath = path_combine(path, fname);
```
levy5307, line 876: auto cleanup = dsn::defer([&]() { remove_path(fpath) });

src/replica/replica_stub.cpp
```diff
@@ -790,6 +776,37 @@ void replica_stub::initialize(const replication_options &opts, bool clear /* = f
     }
 }
 
+void replica_stub::initialize_fs_manager(std::vector<std::string> &data_dirs,
```
levy5307, line 779: ```const std::vector<std::string> &data_dirs```

src/replica/replica_stub.cpp
```diff
@@ -790,6 +776,37 @@ void replica_stub::initialize(const replication_options &opts, bool clear /* = f
...
+                                         std::vector<std::string> &data_dir_tags)
+{
+    std::string cdir;
+    std::string err_msg;
+    int count = 0;
+    std::vector<std::string> available_dirs;
+    std::vector<std::string> available_dir_tags;
+    for (int i = 0; i < data_dir_tags.size(); ++i) {
+        std::string &dir = data_dirs[i];
+        if (dsn_unlikely(!utils::filesystem::create_directory(dir, cdir, err_msg)) ||
```
levy5307, line 789: I think 
```
if (dsn_unlikely(!utils::filesystem::create_directory(dir, cdir, err_msg) ||  !utils::filesystem::check_dir_rw(dir, err_msg))) {
```
is better

src/replica/replica_stub.cpp
hycdong: In function `initialize_fs_manager`, I should get the value of `data_dirs[i]` and `data_dirs_tag[i]`, using const cann't use index to get the value, so I remain it as original.

levy5307 (approved):

Shuo-Jia:
> If the disk is broken, it will be excluded and replica instance can start.

The case need add perf-counter and let user know it in time

Shuo-Jia (approved):

PR #838, opened 2021-05-27
fix: plog may be lost on going duplicating when replica re-open and then learn

Shuo-Jia:
https://github.com/apache/incubator-pegasus/issues/740 has describe the problem. 

private log from remote node when learning should be protected, which actually was considered in design of https://github.com/XiaoMi/rdsn/pull/355, but the problem is that the `duplicating status` is false when re-open replica,  and status sync(`on_config_sync` rpc)  is lagging to learning,  which result in the code will be ignored: https://github.com/XiaoMi/rdsn/blob/2b4232809262a1e6102a67ae86db2f37bd58e8ce/src/replica/replica_learn.cpp#L1556-L1559

that is to say, the private log is lost since the log of  path `/learn`don't replay to path `/plog`with the `duplicating = false` when replica re-open. for sync the `duplicating status` in time, this pr don't depend the `config sync`:
https://github.com/XiaoMi/rdsn/blob/2b4232809262a1e6102a67ae86db2f37bd58e8ce/src/replica/replica_config.cpp#L1053

but sync the status from `proposal config` when receive `learng request`, which will make sure the `duplicating status` is right before replay plog

### DIFF
--- a/src/replica/replica_learn.cpp
+++ b/src/replica/replica_learn.cpp
@@ -1476,16 +1476,16 @@ void replica::on_learn_completion_notification_reply(error_code err,
 
 void replica::on_add_learner(const group_check_request &request)
 {
-    ddebug("%s: process add learner, primary = %s, ballot = %" PRId64
-           ", status = %s, last_committed_decree = %" PRId64,
-           name(),
-           request.config.primary.to_string(),
-           request.config.ballot,
-           enum_to_string(request.config.status),
-           request.last_committed_decree);
+    dwarn_replica("process add learner, primary = {}, ballot ={}, status ={}, "
+                  "last_committed_decree = {}, duplicating = {}",
+                  request.config.primary.to_string(),
+                  request.config.ballot,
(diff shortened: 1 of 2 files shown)


src/replica/replica_learn.cpp
```diff
@@ -1476,16 +1476,16 @@ void replica::on_learn_completion_notification_reply(error_code err,
...
 void replica::on_add_learner(const group_check_request &request)
 {
-    ddebug("%s: process add learner, primary = %s, ballot = %" PRId64
-           ", status = %s, last_committed_decree = %" PRId64,
-           name(),
-           request.config.primary.to_string(),
-           request.config.ballot,
-           enum_to_string(request.config.status),
-           request.last_committed_decree);
+    dwarn_replica("process add learner, primary = {}, ballot ={}, status ={}, "
```
hycdong, line 1479: ddebug_replica may be better.

hycdong (approved):

levy5307 (approved):

PR #839, opened 2021-06-07
feat(disk): support add disk rpc

hycdong:
Pr #834 adds broken disk check, the broken disk will be ignored while initialize, but if the disk is  repaired, we should stop sever then the disk can be added. This pull request supports a rpc for adding a new disk, we don't need to restart server in this case.
NOTICE: if you add a totally new disk, not in the config files, you should add this disk path and tag in `data_dir` configuration.


src/common/fs_manager.cpp
```diff
@@ -344,5 +344,30 @@ void fs_manager::update_disk_stat()
...
+    ddebug_f("{}: mark data dir({}) as tag({})", dsn_primary_address().to_string(), norm_path, tag);
+}
+
+bool fs_manager::is_dir_node_available(const std::string &data_dir, const std::string &tag) const
+{
+    zauto_read_lock l(_lock);
+    for (const auto &dir_node : _dir_nodes) {
+        std::string norm_path;
+        utils::filesystem::get_normalized_path(data_dir, norm_path);
+        if (dir_node->full_dir == norm_path || dir_node->tag == tag) {
```
Shuo-Jia, line 365: why not &&

src/replica/replica_stub.cpp
```diff
@@ -1136,6 +1135,53 @@ void replica_stub::on_query_app_info(query_app_info_rpc rpc)
...
+
+        std::string cdir;
+        if (dsn_unlikely(!utils::filesystem::create_directory(dir, cdir, err_msg) ||
+                         !utils::filesystem::check_dir_rw(dir, err_msg))) {
+            resp.err = ERR_FILE_OPERATION_FAILED;
+            resp.__set_err_hint(err_msg);
+            return;
+        }
+
+        ddebug_f("Add a new disk in fs_manager, data_dir={}, tag={}", cdir, data_dir_tags[i]);
```
Shuo-Jia, line 1180: cdir->dir?

Shuo-Jia (commented):
if restart, the dir added should be deleted for it isn't be persist in config-file? it has problem if this

src/common/fs_manager.cpp
hycdong: The new dir_node should not has same same path or tag with the current nodes. For example, if the new node dir path=new_path, tag=tag7, if the current dir node(path = new_path, tag=tag6), the new dir node should not be added, and if the current dir node(path = old_path, tag=tag7), the new dir node also should not be added.

src/replica/replica_stub.cpp
hycdong: Both of them is ok, `dir` is from paramter, and `cdir` is the absolute_dir by function `create_directory`.

hycdong:
> if restart, the dir added should be deleted for it isn't be persist in config-file? it has problem if this

Yes, the pull request is to solve the broken disk added into server, whose path has already existed in config files. If adding a new disk, you should also update the config files, I have already add it in description.

src/common/fs_manager.cpp
Shuo-Jia: if (dir_node->full_dir == norm_path || dir_node->tag == tag) {
return false
}
return true

return false only has same case?

src/replica/replica_stub.cpp
Shuo-Jia: has path: /home/xiaomi/

so `create_directory("/home/xiaomi/pegaus", cdir, err_msg)` return cdir="pegasus" or other?

however, in any case,  I think the log should show the full_path, but absolute path,

src/common/fs_manager.cpp
hycdong: Oh, I know your misunderstading. This function name is `is_dir_node_available` to check if dir exist, if it exists, return true, otherwise return false. The caller `on_add_new_disk` will check its return value, if it is true, reject adding this disk, you can reference L1158-L1163 of replica_stub.cpp added in this pr.

src/replica/replica_stub.cpp
hycdong: Acually, the absolute path is full path. 
`create_directory("/home/xiaomi/pegasus", cdir, err_msg)` will get `cdir="/home/xiaomi/pegasus"` 
`create_directory("pegaus", cdir, err_msg)` will also get `cdir="/home/xiaomi/pegasus"` 
The first parameter of `create_directory` is permitted to be relative path.

Shuo-Jia (approved):

src/common/fs_manager.cpp
```diff
@@ -344,5 +344,30 @@ void fs_manager::update_disk_stat()
...
     _counter_max_available_ratio->set(_max_available_ratio);
 }
+
+void fs_manager::add_new_dir_node(const std::string &data_dir, const std::string &tag)
+{
+    zauto_write_lock l(_lock);
+    std::string norm_path;
+    utils::filesystem::get_normalized_path(data_dir, norm_path);
+    dir_node *n = new dir_node(tag, norm_path);
+    _dir_nodes.emplace_back(n);
```
levy5307, line 354: `_dir_nodes.emplace_back(tag, norm_path)`

levy5307 (approved):

PR #840, opened 2021-06-08
fix(backup_policy): ignore current backup when policy is diabled

zhangyifan27:
Before this patch meta will continue last unfinished backup after restarted, however, whether the policy is disabled or not is not considered. This patch fixed it.

### DIFF
--- a/src/meta/meta_backup_service.cpp
+++ b/src/meta/meta_backup_service.cpp
@@ -661,6 +661,19 @@ void policy_context::sync_backup_to_remote_storage_unlocked(const backup_info &b
 
 void policy_context::continue_current_backup_unlocked()
 {
+    if (_policy.is_disable) {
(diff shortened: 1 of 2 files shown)


levy5307 (approved):

src/meta/meta_backup_service.cpp
```diff
@@ -661,6 +661,20 @@ void policy_context::sync_backup_to_remote_storage_unlocked(const backup_info &b
 
 void policy_context::continue_current_backup_unlocked()
 {
+    if (_policy.is_disable) {
```
Shuo-Jia, line 664: policy is abandoned, so why need consider it?

zhangyifan27: I added context in pr discription.

Shuo-Jia (approved):

PR #842, opened 2021-06-10
fix: fix a compilation error on clang

zhangyifan27:
Fix a compilation error: https://github.com/XiaoMi/rdsn/runs/2786707126?check_suite_focus=true, which was introduced by https://github.com/XiaoMi/rdsn/pull/834


levy5307 (approved):

Shuo-Jia:
> which is introduced by #834

Why this pr passed?

Shuo-Jia (approved):

PR #843, opened 2021-06-10
fix(graceful_exit): adjust the destructing order

zhangyifan27:
See https://github.com/apache/incubator-pegasus/issues/734 for coredump stack details.

The stack shows that we still want to enqueue some tasks when `task_worker_pool` is destroyed.

This patch explicitly specified the order in which `service_app` , `rpc_engine` and `task_engine` stop in `~service_node()`, which is the reverese order they start.


neverchanje (commented):
LGTM

neverchanje (approved):

levy5307 (approved):

PR #845, opened 2021-06-11
fix: implement reset_from to fix plog loss when duplicating with learning

Shuo-Jia:
See https://github.com/apache/incubator-pegasus/issues/765


src/replica/mutation_log.cpp
```diff
@@ -916,6 +916,62 @@ int64_t mutation_log::total_size_no_lock() const
...
+{
+    error_code err = ERR_FILE_OPERATION_FAILED;
+
+    // close for flushing current log and be ready to open new log files after reset
+    close();
+
+    // make sure logs in `dir` (such as /learn) are valid.
+    error_s es = log_utils::check_log_files_continuity(dir);
+    if (!es.is_ok()) {
+        derror_f("the log of source dir {} is valid:{}, will remov it.", dir, es);
```
hycdong, line 931: "the log of source dir {} is invalid:{}, will remove it."

src/replica/mutation_log.cpp
```diff
@@ -916,6 +916,62 @@ int64_t mutation_log::total_size_no_lock() const
...
+    }
+    ddebug_f("move {} to {} as our new log directory", dir, _dir);
+
+    // - make sure logs in moved dir(such as /plog) are valid and can be opened successfully.
+    // - re-open new log files  for loading the new log file and register the files into replica,
+    // please make sure the old log files has been closed
+    err = open(replay_error_callback, write_error_callback);
+    if (err != ERR_OK) {
+        derror_f(
+            "the logs of moved dir {} are valid and open failed:{}, will rollback it.", _dir, err);
```
hycdong, line 959: What is this error log meaning? the logs are valid but open failed?

src/replica/mutation_log.h
```diff
@@ -149,13 +149,10 @@ class mutation_log : public ref_counter
...
-    // Resets private-log with log files under `dir`.
-    // The original plog will be removed after this call.
-    // NOTE: private-log should be opened before this method called.
-    virtual error_code reset_from(const std::string &dir, io_failure_callback)
-    {
-        return ERR_NOT_IMPLEMENTED;
-    }
+    // Resets mutation log with log files under `dir`.
+    // The original log will be removed after this call.
+    // NOTE: log should be opened before this method called.
```
hycdong, line 154: I notice that old comments use `private log`, but this pr updates it into mutation log, will shared log use this function?

src/replica/replica_learn.cpp
```diff
@@ -1507,6 +1508,9 @@ void replica::on_add_learner(const group_check_request &request)
 error_code replica::apply_learned_state_from_private_log(learn_state &state)
 {
     bool duplicating = is_duplicating();
+    // step_back means this round of learn must have been stepped back to include all the
+    // unconfirmed when duplicating. default is false
```
hycdong, line 1512: In current code, ```duplicating && state.__isset.learn_start_decree && state.learn_start_decree < _app->last_committed_decree() + 1``` means `step_back = true`, I think you should add more comments to describe it. I still can not know the meaning of `step_back` clearly.

src/replica/replica_learn.cpp
```diff
@@ -1549,21 +1561,23 @@ error_code replica::apply_learned_state_from_private_log(learn_state &state)
...
+                               if (duplicating && !step_back) {
                                    _private_log->append(
                                        mu, LPC_WRITE_REPLICATION_LOG_COMMON, &_tracker, nullptr);
                                }
                            }
                        });
 
     err = mutation_log::replay(state.files,
-                               [&plist](int log_length, mutation_ptr &mu) {
+                               [step_back, duplicating, &plist](int log_length, mutation_ptr &mu) {
```
hycdong, line 1580: It seems that `step_back` and `duplicating` are not used in callback function of replay, why do you add them here?

src/replica/mutation_log.cpp
Shuo-Jia: also should `invalid`

src/replica/mutation_log.h
Shuo-Jia: I only think this should be a common function

src/replica/replica_learn.cpp
Shuo-Jia: deleted

src/replica/replica_learn.cpp
Shuo-Jia: updateed, and maybe you can see all comment: https://github.com/XiaoMi/rdsn/pull/845/files#diff-304228a830e17d7be16721c87b13698c7d767c47e74c3947e2d005fc30f099b5R1511-R1529

hycdong (dismissed):

src/replica/mutation_log.cpp
```diff
@@ -916,6 +916,63 @@ int64_t mutation_log::total_size_no_lock() const
...
+    std::string temp_dir = _dir + '.' + std::to_string(dsn_now_ns());
+    if (!utils::filesystem::rename_path(_dir, temp_dir)) {
+        derror_f("rename {} to {} failed", _dir, temp_dir);
+        return err;
+    }
+    ddebug_f("moved current log dir {}  to tmp_dir {}", _dir, temp_dir);
+
+    // move source dir to target dir
+    if (!utils::filesystem::rename_path(dir, _dir)) {
+        derror_f("rename {} to {} failed", dir, _dir);
```
levy5307, line 948: I think you should remove temp_dir here if the rename_path returns error

Shuo-Jia: `temp_dir` contain the origin `plog` files, we shouldn't remove it rashly. maybe `rollback` is better

Smityz:
LGTM

levy5307 (approved):
LGTM

hycdong (approved):

PR #841, opened 2021-06-08
fix: add a destructor for `join_point`

zhangyifan27:
I met an error when pegasus exit, the progress ouput is:
```
terminate called after throwing an instance of 'std::bad_function_call'
  what():  bad_function_call
```
coredump stack:
```
(gdb) bt
#0  0x00007f23db90d1d7 in raise () from /lib64/libc.so.6
#1  0x00007f23db90e8c8 in abort () from /lib64/libc.so.6
#2  0x00007f23dc23fb3d in __gnu_cxx::__verbose_terminate_handler () at ../../../../libstdc++-v3/libsupc++/vterminate.cc:95
#3  0x00007f23dc23dba6 in __cxxabiv1::__terminate (handler=<optimized out>) at ../../../../libstdc++-v3/libsupc++/eh_terminate.cc:47
#4  0x00007f23dc23cbf9 in __cxa_call_terminate (ue_header=0x1bb0060) at ../../../../libstdc++-v3/libsupc++/eh_call.cc:54
#5  0x00007f23dc23d545 in __cxxabiv1::__gxx_personality_v0 (version=<optimized out>, actions=<optimized out>, exception_class=5138137972254386944, ue_header=<optimized out>, 
    context=0x7f235e50bc30) at ../../../../libstdc++-v3/libsupc++/eh_personality.cc:676
#6  0x00007f23dbca8903 in ?? () from /lib64/libgcc_s.so.1
#7  0x00007f23dbca8c9b in _Unwind_RaiseException () from /lib64/libgcc_s.so.1
#8  0x00007f23dc23ddfb in __cxxabiv1::__cxa_throw (obj=0x1bb0080, tinfo=0x7f23dc5327e0 <typeinfo for std::bad_function_call>, 
    dest=0x7f23dc26acd0 <std::bad_function_call::~bad_function_call()>) at ../../../../libstdc++-v3/libsupc++/eh_throw.cc:82
#9  0x00007f23dc26ac62 in std::__throw_bad_function_call () at ../../../../../libstdc++-v3/src/c++11/functexcept.cc:139
#10 0x00007f23df232278 in operator() (__args#0=dsn::SYS_EXIT_EXCEPTION, this=<optimized out>) at /home/zhangyifan8/local/gcc-5.4.0/include/c++/5.4.0/functional:2266
[...]

PR #846, opened 2021-06-17
build: warn on missing virtual on base class desctructor

neverchanje:
This option is to prevent further issues like https://github.com/XiaoMi/rdsn/pull/841 

Boost 1.69 has the same errors if this option is turned on. If the reviewers have agreed, we can use a newer version of boost, 1.71 for example (which is the default on ubuntu20.04), then this PR can pass.


Smityz:
LGTM

acelyc111 (dismissed):

acelyc111:
@neverchanje need update pre-built third-parties

levy5307 (dismissed):
LGTM

PR #851, opened 2021-07-05
fix(disk): fix disk space insufficient bug

hycdong:
#833 adds the feature that when disk space is insufficient, server will reject write requests. Disk status will be checked and updated periodically, only when disk status changed, the replicas on it will update their status. Here existing a bug that if the disk space is insufficient when the server started, the disk status will always be insufficient, and it won't be updated when we close disk space check. This pull request fixs the bug above, and adds related unit test.

```diff
[replication]
-enable_disk_available_space_check=true
+reject_write_when_disk_insufficient=true
```

### DIFF
--- a/src/common/fs_manager.cpp
+++ b/src/common/fs_manager.cpp
@@ -42,18 +42,13 @@
 namespace dsn {
 namespace replication {
 
-DSN_DEFINE_bool("replication",
-                enable_disk_available_space_check,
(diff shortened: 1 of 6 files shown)


levy5307 (approved):

Smityz:
LGTM

Smityz (approved):

PR #853, opened 2021-07-08
feat: add flag to enable cluster load balance

levy5307:
add remote command to enable cluster load balance. I will add the implementation of cluster load balance in later pull requests.

### DIFF
--- a/src/meta/greedy_load_balancer.cpp
+++ b/src/meta/greedy_load_balancer.cpp
@@ -39,6 +39,7 @@ namespace replication {
 
 greedy_load_balancer::greedy_load_balancer(meta_service *_svc)
     : simple_load_balancer(_svc),
+      _ctrl_balancer_ignored_apps(nullptr),
       _ctrl_balancer_in_turn(nullptr),
       _ctrl_only_primary_balancer(nullptr),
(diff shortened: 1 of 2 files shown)


src/meta/greedy_load_balancer.cpp
```diff
@@ -827,6 +839,20 @@ void greedy_load_balancer::greedy_balancer(const bool balance_checker)
         }
     }
 
+    if (!_balance_cluster) {
+        app_balancer(balance_checker);
+        return;
+    }
+
+    if (!balance_checker) {
```
Smityz, line 847: what is balance_checker?

levy5307: It is used to get the balancer operation count

hycdong (dismissed):

Shuo-Jia:
why add config with gflag to update, but use remote-command?

levy5307:
> why add config with gflag to update, but use remote-command?

I think they are both good. I change change it to using gflag if you think it is more better

Shuo-Jia (approved):

hycdong (approved):

PR #858, opened 2021-07-20
feat:  add balance type for cluster load balance

levy5307:
implement move replicas for cluster load balance

### DIFF
--- a/src/meta/greedy_load_balancer.cpp
+++ b/src/meta/greedy_load_balancer.cpp
@@ -942,7 +942,7 @@ void greedy_load_balancer::cluster_balancer()
     }
 
     migration_list list;
-    bool need_continue = cluster_replica_balance(cluster_balance_type::kTotal, list);
+    bool need_continue = cluster_replica_balance(cluster_balance_type::Total, list);
     if (!need_continue) {
         return;
     }
@@ -958,7 +958,8 @@ bool greedy_load_balancer::cluster_replica_balance(const cluster_balance_type ty
         return false;
     }
     if (!t_migration_result->empty()) {
-        ddebug_f("migration count of copy primary/secondary = {}", t_migration_result->size());
+        ddebug_f(
+            "migration count of copy {} = {}", enum_to_string(type), t_migration_result->size());
         return false;
     }
     return true;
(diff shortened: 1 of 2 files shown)


hycdong (approved):

src/meta/greedy_load_balancer.cpp
```diff
@@ -936,12 +936,37 @@ void greedy_load_balancer::cluster_balancer()
...
+        ddebug_f(
+            "migration count of copy {} = {}", enum_to_string(type), t_migration_result->size());
+        return false;
+    }
+    return true;
+}
+
+bool greedy_load_balancer::do_cluster_replica_balance(const cluster_balance_type type)
+{
+    return true;
```
levy5307, line 968: I will add the implementation of copy partition in later pull requests.

src/meta/greedy_load_balancer.cpp
```diff
@@ -936,12 +936,37 @@ void greedy_load_balancer::cluster_balancer()
...
+        return;
+    }
+
+    // TODO(zlw): copy primary
+}
+
+bool greedy_load_balancer::cluster_replica_balance(const cluster_balance_type type)
+{
+    bool enough_information = do_cluster_replica_balance(type);
+    if (!enough_information) {
```
Smityz, line 955: It's better to add a log here

src/meta/greedy_load_balancer.h
```diff
@@ -140,6 +150,10 @@ class greedy_load_balancer : public simple_load_balancer
 
     void cluster_balancer();
```
Shuo-Jia, line 151: The name of the function is better a verb:
```diff
- greedy_load_balancer::cluster_balancer
+ greedy_load_balancer::balance_cluster
```

src/meta/greedy_load_balancer.cpp
levy5307: OK. I will add it in later pull requests.

src/meta/greedy_load_balancer.cpp
```diff
@@ -936,12 +936,37 @@ void greedy_load_balancer::cluster_balancer()
...
+        return false;
+    }
+    if (!t_migration_result->empty()) {
```
hycdong, line 958: If the migration result is not empty, why it will return false? If this balance type is copy_secondary, and migration result is not empty, it won't copy_primary.

levy5307: The return value means if it need to do next step. There is no need copy primary if the migration result of copy secondary is not null.

Shuo-Jia (approved):

hycdong (approved):

PR #859, opened 2021-07-21
feat: add perf counter to record dropped timeout rpc count

hycdong:
As the issue [pegasus#786](https://github.com/apache/incubator-pegasus/issues/786) shows, if the in-queue time of request has already exceed its client timeout, it should not be executed, because client has already been timeout.
This pull request is add the perf-counter to record the count of those dropped requests.

I test it on onebox with printing perf-counters and set read/write timeout as 10ms, the logs are like:
```
W2021-07-21 16:58:32.311 (1626857912311375216 15077) replica.local_app0.03003aa8001de7b7: rpc_task.cpp:441:exec(): rpc_request_task(RPC_RRDB_RRDB_GET) from(ip:port) stop to execute due to timeout_ms(10) exceed
W2021-07-21 16:58:33.268 (1626857913268678037 15066) replica.replica1.03003aa8001e0587: rpc_task.cpp:441:exec(): rpc_request_task(RPC_RRDB_RRDB_PUT) from(ip:port) stop to execute due to timeout_ms(10) exceed
```

The printed perf-counters are like:
```
[zion*profiler*RPC_RRDB_RRDB_GET.rpc.dropped, NUMBER, 17.00]
[zion*profiler*RPC_RRDB_RRDB_PUT.rpc.dropped, NUMBER, 133.00]
```


Smityz (approved):

src/runtime/profiler.cpp
```diff
@@ -461,6 +475,17 @@ void profiler::install(service_spec &)
                     COUNTER_TYPE_NUMBER_PERCENTILES,
                     "");
             }
+            if (dsn_config_get_value_bool(
```
levy5307, line 478: Why didn't you use flags to define `rpc_request_dropped_before_execution_when_timeout`?

levy5307 (approved):

PR #860, opened 2021-07-21
fix(duplication): reset confirmed_decree to delete secondary private log after remove duplication

Shuo-Jia:
See https://github.com/apache/incubator-pegasus/issues/783, this pr reset confirmed_decree of secondary after duplication is removed

### DIFF
--- a/src/replica/duplication/replica_duplicator_manager.cpp
+++ b/src/replica/duplication/replica_duplicator_manager.cpp
@@ -127,11 +127,13 @@ void replica_duplicator_manager::update_confirmed_decree_if_secondary(decree con
 
     zauto_lock l(_lock);
     remove_all_duplications();
-    if (confirmed >= 0) {
+    if (confirmed >= 0) { // duplication ongoing
         // confirmed decree never decreases
         if (_primary_confirmed_decree < confirmed) {
             _primary_confirmed_decree = confirmed;
         }
+    } else { // duplication add with freeze but no start or no duplication(include removed)
+        _primary_confirmed_decree = confirmed;
     }
 }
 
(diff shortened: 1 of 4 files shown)


Smityz (approved):

src/replica/duplication/replica_duplicator_manager.cpp
```diff
@@ -127,11 +127,13 @@ void replica_duplicator_manager::update_confirmed_decree_if_secondary(decree con
 
     zauto_lock l(_lock);
     remove_all_duplications();
-    if (confirmed >= 0) {
+    if (confirmed >= 0) { // duplication ongoing
```
levy5307, line 130: _primary_confirmed_decree = std::min(_primary_confirmed_decree, confirmed);

src/replica/duplication/replica_duplicator_manager.cpp
```diff
@@ -127,11 +127,13 @@ void replica_duplicator_manager::update_confirmed_decree_if_secondary(decree con
```
hycdong, line 127: If `request.__isset.confirmed_decree = false`, what's value of `confirmed`? is it equal to 0 or below 0?

Shuo-Jia: Oh, I thought the `__isset.confirmed_decree` is `request.confirmed_decree` before. but it actually is `whether set confirmed_decree`:
```c++
void group_check_request::__set_confirmed_decree(const int64_t val) {
  this->confirmed_decree = val;
__isset.confirmed_decree = true;
}
```

however, I find the value is always `true`:  https://github.com/XiaoMi/rdsn/blob/5c8de19867385cfb54d7da22860b85485055d801/src/replica/replica_check.cpp#L104

So I think revert or keep is same

hycdong (approved):

PR #861, opened 2021-07-21
fix(duplication): force wait confirmed_decree sync complete when start duplication

Shuo-Jia:
See https://github.com/apache/incubator-pegasus/issues/784, this PR add wait logic berfore `replay log` of duplication


src/replica/duplication/load_from_private_log.cpp
```diff
@@ -69,6 +70,28 @@ void load_from_private_log::run()
     dassert_replica(_start_decree != invalid_decree, "{}", _start_decree);
     _duplicator->verify_start_decree(_start_decree);
 
+    if (_mutation_batch.last_decree() == invalid_decree) {
```
hycdong, line 73: I recommend adding comments for those condition

hycdong (approved):

levy5307 (approved):

PR #862, opened 2021-07-22
feat(duplication): derror and add perf-counter if mutation loss instead of dassert for duplication

Shuo-Jia:
## Description
In https://github.com/apache/incubator-pegasus/issues/766, we have list many problems related to `mutatiaon loss`:
- [x] XiaoMi/rdsn#838
- [x] XiaoMi/rdsn#845
- [x] XiaoMi/rdsn#860
- [x] XiaoMi/rdsn#861

Whose crush point is the code:
https://github.com/XiaoMi/rdsn/blob/e31d49ca6d98dba7b83b272cae0346edf6026337/src/replica/prepare_list.cpp#L154-L156

This is implementation of prepare_list in `2pc`, which can make sure the request is valid, otherwise, it will crush and need manual intervention.

Since the `crash` has serious impaction for service and it still cause the `duplication failed and mutation loss`, **the `dassert to crash` may be not necessary for `duplication`.**

This PR re-implement the `commit funtion` for `prepare_list` in duplication, which replace `dassert` to `derror`and add `perf-counter`, like:
```diff
- dassert_replica(mu != nullptr && mu->is_logged(), "mutation {} is missing in prepare list", d0);
+ derror_replica..
+ counter_dulication_mutation_loss_count...
```
which still warning user that the duplication failed with mutation loss, user can choose restart or ignore this loss.

`duplication` feature is still unstable, this pr can avoid the follow service crash if has some other problem

## Perf-counter

```diff
+ (todo) support collector
+ replica*eon.replica*dup_recent_mutation_loss_count@{gpid}
```

### DIFF
--- a/src/replica/duplication/mutation_batch.cpp
+++ b/src/replica/duplication/mutation_batch.cpp
@@ -50,11 +50,10 @@ void mutation_buffer::commit(decree d, commit_type ct)
     for (decree d0 = last_committed_decree() + 1; d0 <= d; d0++) {
         mutation_ptr next_commit_mutation = get_mutation_by_decree(d0);
         if (next_commit_mutation == nullptr || !next_commit_mutation->is_logged()) {
-            derror_replica("mutation[decree={}, last_commit_decree={}] is lost: "
+            derror_replica("mutation[{}] is lost: "
                            "prepare_last_commit_decree={}, prepare_min_decree={}, "
                            "prepare_max_decree={}",
-                           next_commit_mutation->data.header.last_committed_decree,
-                           next_commit_mutation->data.header.decree,
+                           d, 
                            last_committed_decree(),
                            min_decree(),
                            max_decree());
--- a/src/replica/duplication/mutation_batch.cpp
+++ b/src/replica/duplication/mutation_batch.cpp
@@ -48,12 +48,12 @@ void mutation_buffer::commit(decree d, commit_type ct)
 
     ballot last_bt = 0;
     for (decree d0 = last_committed_decree() + 1; d0 <= d; d0++) {
-        mutation_ptr next_commit_mutation = get_mutation_by_decree(d0);
-        // The unexpected case as follow: next_commit_decree is out of [start~end]
+        mutation_ptr next_committed_mutation = get_mutation_by_decree(d0);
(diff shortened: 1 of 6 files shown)


src/replica/duplication/mutation_batch.cpp
```diff
@@ -20,13 +20,70 @@
...
+        //                         |                                       |
+        //                        n                                    n+1
+        //
+        //  [min_decree------max_decree]
+        //                |                                |
+        //             n+m(m>1)            n+k(k>=m)
+        //
+        // just derror but not dassert if mutation loss or other problem, it's different from base
+        // class implement. And from the error and perf-counter, we can choose restart duplication
+        // or ignore the loss.
```
hycdong, line 64: In this comments, I suggest `last_commit_decree` -> `last_committed_decree`, `next_commit_decree` -> `next_committed_decree`, `next_commit_mutation` -> `next_committed_mutation`

src/replica/duplication/mutation_batch.cpp
```diff
@@ -20,13 +20,70 @@
...
+        if (next_commit_mutation == nullptr || !next_commit_mutation->is_logged()) {
+            derror_replica("mutation[{}] is lost: "
```
hycdong, line 66: mutation[{}] is lost in prepare_list

hycdong (dismissed):

src/replica/duplication/mutation_batch.cpp
```diff
@@ -20,13 +20,70 @@
...
+        "eon.replica", counter_str.c_str(), COUNTER_TYPE_VOLATILE_NUMBER, counter_str.c_str());
+}
+
+void mutation_buffer::commit(decree d, commit_type ct)
+{
+    if (d <= last_committed_decree())
+        return;
+
+    if (ct != COMMIT_TO_DECREE_HARD) {
+        dassert_replica(false, "invalid commit type %d", (int)ct);
```
levy5307, line 46: dassert_replica(false, "invalid commit type {}", (int)ct);

levy5307 (commented):
LGTM

levy5307 (approved):

hycdong (approved):

PR #863, opened 2021-07-22
feat: throw exception if reads corrupt data

levy5307:
In previous implementation, server will coredump if it reads corrupt data. 
In this pull request, I changed the way to throw exception. So someone can catch the exception if he don't want to make server crash.

### DIFF
--- a/include/dsn/utility/binary_reader.h
+++ b/include/dsn/utility/binary_reader.h
@@ -50,8 +50,8 @@ public:
     int get_remaining_size() const { return _remaining_size; }
 
 protected:
-    int internal_read(blob &blob, int len);
-    int internal_read(char *buffer, int sz);
+    int inner_read(blob &blob, int len);
+    int inner_read(char *buffer, int sz);
 
     blob _blob;
     int _size;
--- a/include/dsn/utility/binary_reader.h
+++ b/include/dsn/utility/binary_reader.h
@@ -39,9 +39,9 @@ public:
     int read(/*out*/ bool &val) { return read_pod(val); }
 
     int read(/*out*/ std::string &s);
-    int read(char *buffer, int sz);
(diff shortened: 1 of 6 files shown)


src/replica/replica_2pc.cpp
```diff
@@ -76,6 +76,15 @@ void replica::on_client_write(dsn::message_ex *request, bool ignore_throttling)
     }
 
     task_spec *spec = task_spec::get(request->rpc_code());
+    if (dsn_unlikely(nullptr == spec || request->rpc_code() == TASK_CODE_INVALID)) {
+        dwarn_f("recv message with unhandled rpc name {} from {}, trace_id = {}",
```
hycdong, line 80: How about updating warning level into error level?

hycdong:
I think `feat` may be better than `refactor` to describe this pull request.

src/replica/replica_2pc.cpp
levy5307: OK

hycdong (approved):

Shuo-Jia (approved):

include/dsn/utility/binary_reader.h
```diff
@@ -39,21 +40,27 @@ class binary_reader
     int read(/*out*/ bool &val) { return read_pod(val); }
 
     int read(/*out*/ std::string &s);
-    int read(char *buffer, int sz);
+    virtual int read(char *buffer, int sz);
```
Smityz, line 43: It's an important path on write, add a virtual function will add a performance cost in c++, is this change necessary？

levy5307: The performance cost introduced by vtable can be negligible. And this pull request is used for dealing with dirty mutation log.

PR #865, opened 2021-07-23
feat: add an env to enable or disable block cache of an app

ZhongChaoqiang:
Sometimes, there are some apps of different business in a cluster.Some apps  want to enable block cache, others my be not.
So we need add an env to enable or disable block cache of an app.

### DIFF
--- a/src/meta/app_env_validator.cpp
+++ b/src/meta/app_env_validator.cpp
@@ -174,7 +174,6 @@ void app_env_validator::register_all_validators()
         {replica_envs::ROCKSDB_USAGE_SCENARIO, nullptr},
         {replica_envs::ROCKSDB_CHECKPOINT_RESERVE_MIN_COUNT, nullptr},
         {replica_envs::ROCKSDB_CHECKPOINT_RESERVE_TIME_SECONDS, nullptr},
-        {replica_envs::ROCKSDB_BLOCK_CACHE_ENABLED, nullptr},
         {replica_envs::MANUAL_COMPACT_DISABLED, nullptr},
         {replica_envs::MANUAL_COMPACT_MAX_CONCURRENT_RUNNING_COUNT, nullptr},
         {replica_envs::MANUAL_COMPACT_ONCE_TRIGGER_TIME, nullptr},
--- a/src/meta/app_env_validator.cpp
+++ b/src/meta/app_env_validator.cpp
@@ -165,7 +165,8 @@ void app_env_validator::register_all_validators()
         {replica_envs::ROCKSDB_ITERATION_THRESHOLD_TIME_MS,
          std::bind(&check_rocksdb_iteration, std::placeholders::_1, std::placeholders::_2)},
         {replica_envs::ROCKSDB_BLOCK_CACHE_ENABLED,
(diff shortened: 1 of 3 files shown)


neverchanje (approved):

neverchanje:
This feature makes sense. But how do you suppose to close a rocksdb's block cache at runtime after it's created?

ZhongChaoqiang:
> This feature makes sense. But how do you suppose to close a rocksdb's block cache at runtime after it's created? Does rocksdb provide such an API?

We don't need to close rocksdb's block ceche, the cache in rocksdb will be removed by LRU.

neverchanje:
Do you have any graphs or testing results proving that? 

> the cache in rocksdb will be removed by LRU. 
What's the mechanism behind?

neverchanje (dismissed):

hycdong:
Thanks for your contribution~ Cloud you please add an issue in [new_issue_in_pegasus](https://github.com/apache/incubator-pegasus/issues/new/choose)? Besides, as @neverchanje comments, could you please add your test process and explain how RocksDB exchange block-cache-disabled table by LRU and how to promise the block_cache of this table is disabled?

ZhongChaoqiang:
> Do you have any graphs or testing results proving that?
> 
> > the cache in rocksdb will be removed by LRU.
> 
> What's the mechanism behind?

We control it by fill_cache option of rocksdb::ReadOptions.
 If fill_cache  is true, when reading data from rocksdb, the data block will be filled into cache.
If fill_cache  change to false,the data block will not be filled into cache.But we do not need to clear the block of cached.When block cache is full, the oldest block will be remove by LRU.
You can see it in PR of pegasus.(https://github.com/apache/incubator-pegasus/pull/792/files)

src/meta/app_env_validator.cpp
```diff
@@ -161,6 +161,7 @@ void app_env_validator::register_all_validators()
         {replica_envs::ROCKSDB_USAGE_SCENARIO, nullptr},
         {replica_envs::ROCKSDB_CHECKPOINT_RESERVE_MIN_COUNT, nullptr},
         {replica_envs::ROCKSDB_CHECKPOINT_RESERVE_TIME_SECONDS, nullptr},
+        {replica_envs::ROCKSDB_BLOCK_CACHE_ENABLED, nullptr},
```
levy5307, line 164: could you add a check function for ROCKSDB_BLOCK_CACHE_ENABLED?

levy5307:
And could you add a new issue for this pr?

src/meta/app_env_validator.cpp
ZhongChaoqiang: > could you add a check function for ROCKSDB_BLOCK_CACHE_ENABLED?

OK，I have added it.

ZhongChaoqiang:
> And could you add a new issue for this pr?

OK! The issue is apache/incubator-pegasus#795

ZhongChaoqiang:
> And could you add a new issue for this pr?

OK! The issue is apache/incubator-pegasus#795

neverchanje (dismissed):

neverchanje (approved):

levy5307 (approved):

PR #874, opened 2021-08-09
feat: make token bucket in fds configurable

levy5307:
### Manual Tests

- Restore

rate limiter rate: 100MB->500MB
img_e5fb934d-7c62-452d-a9cb-be8ce58b9b8l
and the batch size is from 100MB --> 500MB, and it can works well.

- Backup

At rate of 100MB, and change burst size from 100MB->500MB. it can works well

img_468cf4be-380d-4621-929a-f87c4e99f58l

### DIFF
--- a/src/block_service/fds/fds_service.cpp
+++ b/src/block_service/fds/fds_service.cpp
@@ -464,7 +464,7 @@ error_code fds_file_object::get_content_in_batches(uint64_t start,
         uint64_t batch_len = std::min(BATCH_SIZE, start + to_transfer_bytes - pos);
         // burst size should not be less than consume size
         const uint64_t rate = FLAGS_fds_read_limit_rate << 20;
-        _service->_read_token_bucket->consumeWithBorrowAndWait(batch_len, rate, 2 * rate);
+        _service->_read_token_bucket->consumeWithBorrowAndWait(batch_len, rate, rate);
 
         err = get_content(pos, batch_len, os, once_transfered_bytes);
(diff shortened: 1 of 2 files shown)


Shuo-Jia (dismissed):

src/block_service/fds/fds_service.cpp
```diff
@@ -533,11 +533,14 @@ error_code fds_file_object::put_content(/*in-out*/ std::istream &is,
     galaxy::fds::GalaxyFDSClient *c = _service->get_client();
 
     // get tokens from token bucket
-    if (!_service->_write_token_bucket->consumeWithBorrowAndWait(to_transfer_bytes)) {
-        ddebug_f("the transfer count({}) is greater than burst size({}), so it is rejected by "
+    if (!_service->_write_token_bucket->consumeWithBorrowAndWait(to_transfer_bytes,
+                                                                 FLAGS_fds_write_limit_rate << 20,
+                                                                 FLAGS_fds_write_burst_size
+                                                                     << 20)) {
+        ddebug_f("the transfer count({}) is greater than burst size({}MB), so it is rejected by "
```
hycdong, line 540: `the transfer count({})` byte maybe better? Because the burst size has unit(MB)

levy5307: good idea :）

hycdong (approved):

Shuo-Jia (approved):

PR #888, opened 2021-08-26
feat(thirdparty): change galaxy-fds-sdk-cpp download source

Smityz:
It's too slow to use git clone in China, so I want to change it, it will be far faster than before.


thirdparty/CMakeLists.txt
```diff
@@ -199,7 +199,7 @@ ExternalProject_Add(poco
         )
 
 ExternalProject_Add(fds
-        GIT_REPOSITORY https://github.com/XiaoMi/galaxy-fds-sdk-cpp.git
+        GIT_REPOSITORY https://gitee.com/mirrors_XiaoMi/galaxy-fds-sdk-cpp
```
Shuo-Jia, line 202: seem allow multi url, you can test add both them: "url1,url2"

levy5307 (dismissed):

thirdparty/CMakeLists.txt
Smityz: can't work with it

PR #889, opened 2021-08-28
feat: make the block size configurable for mutation logs

empiredan:
The block size used to read the files of mutation logs should be configurable to limit the memory consumed by rDSN, with `log_private_block_bytes` specified for private logs and `log_shared_block_bytes` for shared logs.

The default block size is still 1MB, as it's used before.

### DIFF
--- a/src/common/replication_common.cpp
+++ b/src/common/replication_common.cpp
@@ -95,6 +95,7 @@ replication_options::replication_options()
     log_private_batch_buffer_flush_interval_ms = 10000;
     log_private_reserve_max_size_mb = 0;
     log_private_reserve_max_time_seconds = 0;
+    log_private_block_bytes = 1024 * 1024;
 
     log_shared_file_size_mb = 32;
     log_shared_file_count_limit = 100;
@@ -102,6 +103,7 @@ replication_options::replication_options()
     log_shared_force_flush = false;
     log_shared_pending_size_throttling_threshold_kb = 0;
     log_shared_pending_size_throttling_delay_ms = 0;
+    log_shared_block_bytes = 1024 * 1024;
 
     config_sync_disabled = false;
     config_sync_interval_ms = 30000;
@@ -348,6 +350,10 @@ void replication_options::initialize()
         "log_private_reserve_max_time_seconds",
         log_private_reserve_max_time_seconds,
         "max time in seconds of useless private log to be reserved");
+    log_private_block_bytes = dsn_config_get_value_int64("replication",
+                                                         "log_private_block_bytes",
+                                                         log_private_block_bytes,
+                                                         "block size by bytes for private log");
 
     log_shared_file_size_mb =
         (int)dsn_config_get_value_uint64("replication",
@@ -378,6 +384,10 @@ void replication_options::initialize()
                                          "log_shared_pending_size_throttling_delay_ms",
                                          log_shared_pending_size_throttling_delay_ms,
                                          "log_shared_pending_size_throttling_delay_ms");
+    log_shared_block_bytes = dsn_config_get_value_int64("replication",
+                                                        "log_shared_block_bytes",
+                                                        log_shared_block_bytes,
+                                                        "block size by bytes for shared log");
 
     config_sync_disabled = dsn_config_get_value_bool(
         "replication",
--- a/src/replica/log_file.cpp
+++ b/src/replica/log_file.cpp
@@ -35,7 +35,8 @@ namespace dsn {
 namespace replication {
 
 log_file::~log_file() { close(); }
-/*static */ log_file_ptr log_file::open_read(const char *path, /*out*/ error_code &err)
+/*static */ log_file_ptr
+log_file::open_read(const char *path, /*out*/ error_code &err, size_t block_bytes)
 {
     char splitters[] = {'\\', '/', 0};
     std::string name = utils::get_last_component(std::string(path), splitters);
@@ -86,7 +87,7 @@ log_file::~log_file() { close(); }
         return nullptr;
     }
 
-    auto lf = new log_file(path, hfile, index, start_offset, true);
+    auto lf = new log_file(path, hfile, index, start_offset, true, block_bytes);
     lf->reset_stream();
     blob hdr_blob;
     err = lf->read_next_log_block(hdr_blob);
@@ -125,7 +126,8 @@ log_file::~log_file() { close(); }
     return lf;
 }
 
-/*static*/ log_file_ptr log_file::create_write(const char *dir, int index, int64_t start_offset)
+/*static*/ log_file_ptr
+log_file::create_write(const char *dir, int index, int64_t start_offset, size_t block_bytes)
 {
     char path[512];
     sprintf(path, "%s/log.%d.%" PRId64, dir, index, start_offset);
@@ -141,12 +143,16 @@ log_file::~log_file() { close(); }
         return nullptr;
     }
 
-    return new log_file(path, hfile, index, start_offset, false);
+    return new log_file(path, hfile, index, start_offset, false, block_bytes);
 }
 
-log_file::log_file(
-    const char *path, disk_file *handle, int index, int64_t start_offset, bool is_read)
-    : _is_read(is_read)
+log_file::log_file(const char *path,
+                   disk_file *handle,
+                   int index,
+                   int64_t start_offset,
+                   bool is_read,
+                   size_t block_bytes)
+    : _is_read(is_read), _block_bytes(block_bytes)
 {
     _start_offset = start_offset;
     _end_offset = start_offset;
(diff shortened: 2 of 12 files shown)


src/common/replication_common.cpp
```diff
@@ -348,6 +350,10 @@ void replication_options::initialize()
         "log_private_reserve_max_time_seconds",
         log_private_reserve_max_time_seconds,
         "max time in seconds of useless private log to be reserved");
+    log_private_block_bytes = dsn_config_get_value_int64("replication",
```
acelyc111, line 353: Now a new gflag like config defination is introduced, you can use it instead.
See https://github.com/XiaoMi/rdsn/blob/552169e0b10000d94113d5715be1ee4a1c58352d/include/dsn/utility/flags.h#L56

src/common/replication_common.cpp
```diff
@@ -348,6 +350,10 @@ void replication_options::initialize()
...
+                                                         "log_private_block_bytes",
+                                                         log_private_block_bytes,
+                                                         "block size by bytes for private log");
```
acelyc111, line 356: 'in bytes' ?

src/replica/log_file.cpp
```diff
@@ -35,7 +35,8 @@ namespace dsn {
 namespace replication {
 
 log_file::~log_file() { close(); }
-/*static */ log_file_ptr log_file::open_read(const char *path, /*out*/ error_code &err)
+/*static */ log_file_ptr
+log_file::open_read(const char *path, /*out*/ error_code &err, size_t block_bytes)
```
acelyc111, line 39: If you using the new FLAGS_XXX instead, then these parameters are not needed to pass everywhere.

Smityz:
Can you add some test results about your changes?

hycdong:
Hi, empiredan~ I notice that you would like to make  block size configurable to `limit the memory consumed by rDSN`. I'm really interested in the memory related problems. Could you please adding more comments about it?  Thanks a lot.

empiredan:
> Can you add some test results about your changes? Like the relationship between configuration and memory

@Smityz Yeah, it's necessary to provide some test results that show how much we can limit memory, though we've not make some formal tests. I'll try to make some ones.

empiredan:
> Hi, empiredan~ I notice that you would like to make block size configurable to `limit the memory consumed by rDSN`. I'm really interested in the memory related problems. Could you please adding more comments about it? Thanks a lot.

@hycdong Great, I'm also interested in the memory related problems, we can discuss about this one. 

Since the original description for this PR is too simple, I'll explain some more. Since the memory of our machine is very limited (also deployed with many other services), we'll try our best to save memory.

Actually this PR is added in our internal version more than a year ago. At that time, I observed from the memory profile by tcmalloc that a lot of memory had been consumed by reading plog files into the caches. Therefore, I make the buffer size for plog configurable to limit memory consumed, which is tuned from 1MB to 256KB.

As for this PR, I also make the buffer size for slog configurable. Now there is a problem we can discuss. According to the suggestion from @acelyc111 , we can use FLAGS_XXX to denote the buffer size for slog and plog. However, besides the slog and plog, some other components, such as mutation log tools and unit tests, are operating the underlying log_file class, which will lead to a deep call chain to pass the parameter of block size. 

This will also result in a lot of modifications of source files. To some extent maybe a default block size will alleviate this problem, but it seems not a graceful method. Have some advices about this problem, or this PR ? Thanks !

hycdong:
> > Hi, empiredan~ I notice that you would like to make block size configurable to `limit the memory consumed by rDSN`. I'm really interested in the memory related problems. Could you please adding more comments about it? Thanks a lot.
> 
> @hycdong Great, I'm also interested in the memory related problems, we can discuss about this one.
> 
> Since the original description for this PR is too simple, I'll explain some more. Since the memory of our machine is very limited (also deployed with many other services), we'll try our best to save memory.
> 
> Actually this PR is added in our internal version more than a year ago. At that time, I observed from the memory profile by tcmalloc that a lot of memory had been consumed by reading plog files into the caches. Therefore, I make the buffer size for plog configurable to limit memory consumed, which is tuned from 1MB to 256KB.
> 
> As for this PR, I also make the buffer size for slog configurable. Now there is a problem we can discuss. According to the suggestion from @acelyc111 , we can use FLAGS_XXX to denote the buffer size for slog and plog. However, besides the slog and plog, some other components, such as mutation log tools and unit tests, are operating the underlying log_file class, which will lead to a deep call chain to pass the parameter of block size.
> 
> This will also result in a lot of modifications of source files. To some extent maybe a default block size will alleviate this problem, but it seems not a graceful method. Have some advices about this problem, or this PR ? Thanks !

Thanks for your detailed explanation~ I agree that make the buffer size for slog configurable is a good job~
I have an idea to avoid lots of modification of source files. By default, we use a default block size, such as in unit tests and mutation log tools etc. But if it is really used (in my view, it is really used in `log_file_stream.h` ), you can implement it like:
```
// in log_file::file_streamer construction
auto current_value = default_value
if(current_value != FLAGS_value){
      current_value = FLAGS_value;
}
```

empiredan:
> > > Hi, empiredan~ I notice that you would like to make block size configurable to `limit the memory consumed by rDSN`. I'm really interested in the memory related problems. Could you please adding more comments about it? Thanks a lot.
> > 
> > 
> > @hycdong Great, I'm also interested in the memory related problems, we can discuss about this one.
> > Since the original description for this PR is too simple, I'll explain some more. Since the memory of our machine is very limited (also deployed with many other services), we'll try our best to save memory.
> > Actually this PR is added in our internal version more than a year ago. At that time, I observed from the memory profile by tcmalloc that a lot of memory had been consumed by reading plog files into the caches. Therefore, I make the buffer size for plog configurable to limit memory consumed, which is tuned from 1MB to 256KB.
> > As for this PR, I also make the buffer size for slog configurable. Now there is a problem we can discuss. According to the suggestion from @acelyc111 , we can use FLAGS_XXX to denote the buffer size for slog and plog. However, besides the slog and plog, some other components, such as mutation log tools and unit tests, are operating the underlying log_file class, which will lead to a deep call chain to pass the parameter of block size.
> > This will also result in a lot of modifications of source files. To some extent maybe a default block size will alleviate this problem, but it seems not a graceful method. Have some advices about this problem, or this PR ? Thanks !
> 
> Thanks for your detailed explanation~ I agree that make the buffer size for slog configurable is a good job~
> I have an idea to avoid lots of modification of source files. By default, we use a default block size, such as in unit tests and mutation log tools etc. But if it is really used (in my view, it is really used in `log_file_stream.h` ), you can implement it like:
> 
> ```
> // in log_file::file_streamer construction
> auto current_value = default_value
> if(current_value != FLAGS_value){
>       current_value = FLAGS_value;
> }
> ```

@hycdong Good idea ! Thanks for suggestion ! Does this mean we use a single config `FLAGS_block_bytes` to denote the block size of both slog and plog ?

hycdong:
> > > > Hi, empiredan~ I notice that you would like to make block size configurable to `limit the memory consumed by rDSN`. I'm really interested in the memory related problems. Could you please adding more comments about it? Thanks a lot.
> > > 
> > > 
> > > @hycdong Great, I'm also interested in the memory related problems, we can discuss about this one.
> > > Since the original description for this PR is too simple, I'll explain some more. Since the memory of our machine is very limited (also deployed with many other services), we'll try our best to save memory.
> > > Actually this PR is added in our internal version more than a year ago. At that time, I observed from the memory profile by tcmalloc that a lot of memory had been consumed by reading plog files into the caches. Therefore, I make the buffer size for plog configurable to limit memory consumed, which is tuned from 1MB to 256KB.
> > > As for this PR, I also make the buffer size for slog configurable. Now there is a problem we can discuss. According to the suggestion from @acelyc111 , we can use FLAGS_XXX to denote the buffer size for slog and plog. However, besides the slog and plog, some other components, such as mutation log tools and unit tests, are operating the underlying log_file class, which will lead to a deep call chain to pass the parameter of block size.
> > > This will also result in a lot of modifications of source files. To some extent maybe a default block size will alleviate this problem, but it seems not a graceful method. Have some advices about this problem, or this PR ? Thanks !
> > 
> > 
> > Thanks for your detailed explanation~ I agree that make the buffer size for slog configurable is a good job~
> > I have an idea to avoid lots of modification of source files. By default, we use a default block size, such as in unit tests and mutation log tools etc. But if it is really used (in my view, it is really used in `log_file_stream.h` ), you can implement it like:
> > ```
> > // in log_file::file_streamer construction
> > auto current_value = default_value
> > if(current_value != FLAGS_value){
> >       current_value = FLAGS_value;
> > }
> > ```
> 
> @hycdong Good idea ! Thanks for suggestion ! Does this mean we use a single config `FLAGS_block_bytes` to denote the block size of both slog and plog ?

Sorry for long-time no-reply. Yes, you can use `FLAGS_block_bytes` to denote the block size of both slog and plog, and other method that won't modify too many code~

PR #894, opened 2021-08-30
feat(perf_counter): add a perf_counter of server session

Smityz:
Add a perf counter to monitor the session kept on the server.
```
[server*network*client_session_count, NUMBER, 3.00]
[server*network*client_session_count, NUMBER, 4.00]
[server*network*client_session_count, NUMBER, 4.00]
[server*network*client_session_count, NUMBER, 4.00]
[server*network*client_session_count, NUMBER, 5.00]
[server*network*client_session_count, NUMBER, 14.00]
[server*network*client_session_count, NUMBER, 14.00]
[server*network*client_session_count, NUMBER, 4.00]
```

### DIFF
--- a/include/dsn/tool-api/network.h
+++ b/include/dsn/tool-api/network.h
@@ -32,6 +32,7 @@
 #include <dsn/tool-api/rpc_address.h>
 #include <dsn/utility/exp_delay.h>
 #include <dsn/utility/dlib.h>
+#include <dsn/perf_counter/perf_counter_wrapper.h>
 #include <atomic>
 
 namespace dsn {
@@ -40,6 +41,8 @@ class rpc_engine;
 class service_node;
 class task_worker_pool;
 class task_queue;
+class perf_counter_wrapper;
+
 /*!
 @addtogroup tool-api-providers
 @{
@@ -186,6 +189,7 @@ protected:
     utils::rw_lock_nr _servers_lock;
 
     uint32_t _cfg_conn_threshold_per_ip;
+    perf_counter_wrapper _client_session_count;
 };
 
 /*!
--- a/include/dsn/tool-api/network.h
+++ b/include/dsn/tool-api/network.h
@@ -41,7 +41,6 @@ class rpc_engine;
 class service_node;
 class task_worker_pool;
 class task_queue;
-class perf_counter_wrapper;
 
 /*!
 @addtogroup tool-api-providers
--- a/src/runtime/rpc/network.cpp
+++ b/src/runtime/rpc/network.cpp
@@ -705,11 +705,12 @@ void connection_oriented_network::on_server_session_accepted(rpc_session_ptr &s)
 
 void connection_oriented_network::on_server_session_disconnected(rpc_session_ptr &s)
 {
-    int ip_count =
-        0; // how many unique client(the same ip:port is considered to be a unique client)
-    int ip_conn_count =
-        0; // one unique client may remain more than one connection on the server, which
-           // is a unexpected behavior of client, we should record it in logs.
+    // how many unique client(the same ip:port is considered to be a unique client)
+    int ip_count = 0;
+    // one unique client may remain more than one connection on the server, which
+    // is an unexpected behavior of client, we should record it in logs.
+    int ip_conn_count = 0;
+
     bool session_removed = false;
     {
         utils::auto_write_lock l(_servers_lock);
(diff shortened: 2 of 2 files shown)


include/dsn/tool-api/network.h
```diff
@@ -40,6 +41,8 @@ class rpc_engine;
 class service_node;
 class task_worker_pool;
 class task_queue;
+class perf_counter_wrapper;
```
levy5307, line 44: I think this line is useless

src/runtime/rpc/network.cpp
```diff
@@ -590,6 +590,11 @@ connection_oriented_network::connection_oriented_network(rpc_engine *srv, networ
     : network(srv, inner_provider)
 {
     _cfg_conn_threshold_per_ip = 0;
+    _client_session_count.init_global_counter("replica",
+                                              "network",
+                                              "server_session_count",
+                                              COUNTER_TYPE_NUMBER,
+                                              "current session count on server");
```
hycdong, line 597: I think `server.network.client_session_count` maybe better. 
Firstly, both replica server and meta server will share this counter, `server` maybe more suitable.  Besides, `client_session_count` or `session_count` is better.

src/runtime/rpc/network.cpp
```diff
@@ -663,8 +669,8 @@ rpc_session_ptr connection_oriented_network::get_server_session(::dsn::rpc_addre
 
 void connection_oriented_network::on_server_session_accepted(rpc_session_ptr &s)
 {
-    int scount = 0;
-    int ecount = 1;
+    int ip_count = 0;
+    int ip_conn_count = 1;
```
hycdong, line 673: Could you please adding some comments to explain the difference between `ip_count` and `ip_conn_count`?

hycdong:
In my view, to record the server session count, when came a new client session,  `_client_session_count++`, when a client disconnect, `_client_session_count++`. However, I only saw `_client_session_count->set(ip_count);` in this patch to set the counter value. Could you please explain how to record the right session count?

src/runtime/rpc/network.cpp
Smityz: https://github.com/XiaoMi/rdsn/blob/a1e831b6a2c8cefe7e24eb5259d146818d4d2ef9/src/runtime/rpc/network.cpp#L708
The description of these names already existed in this file

Smityz:
> In my view, to record the server session count, when came a new client session, `_client_session_count++`, when a client disconnect, `_client_session_count++`. However, I only saw `_client_session_count->set(ip_count);` in this patch to set the counter value. Could you please explain how to record the right session count?

I just add a perf counter below the original log, its correctness is guaranteed by the log.

Shuo-Jia (approved):

src/runtime/rpc/network.cpp
hycdong: Well, the meaning of `ip_count` is clear. I just don't know the meaning of `ip_conn_count`, especially with the comment below.
https://github.com/XiaoMi/rdsn/blob/a1e831b6a2c8cefe7e24eb5259d146818d4d2ef9/src/runtime/rpc/network.cpp#L709

hycdong:
> > In my view, to record the server session count, when came a new client session, `_client_session_count++`, when a client disconnect, `_client_session_count++`. However, I only saw `_client_session_count->set(ip_count);` in this patch to set the counter value. Could you please explain how to record the right session count?
> 
> I just add a perf counter below the original log, its correctness is guaranteed by the log.

I just don't figure out the original code logic :-( , and I wonder if you can explain it to me ^_^

Smityz:
> > > In my view, to record the server session count, when came a new client session, `_client_session_count++`, when a client disconnect, `_client_session_count++`. However, I only saw `_client_session_count->set(ip_count);` in this patch to set the counter value. Could you please explain how to record the right session count?
> > 
> > 
> > I just add a perf counter below the original log, its correctness is guaranteed by the log.
> 
> I just don't figure out the original code logic, :-( , and I wonder if you can explain it to me ^_^.

I add more comments here https://github.com/XiaoMi/rdsn/pull/894/files#diff-8e070378de2a59fce981099050b87a61f246697846a65c568ca5b3a2798d3bedL700
Hope that can help you

Smityz:
> In my view, to record the server session count, when came a new client session, `_client_session_count++`, when a client disconnect, `_client_session_count++`. However, I only saw `_client_session_count->set(ip_count);` in this patch to set the counter value. Could you please explain how to record the right session count?

add count:
https://github.com/XiaoMi/rdsn/blob/fcd899333a0f3eee526d32d58e3cff1f48b1baab/src/runtime/rpc/network.cpp#L677
minus count:
https://github.com/XiaoMi/rdsn/blob/fcd899333a0f3eee526d32d58e3cff1f48b1baab/src/runtime/rpc/network.cpp#L719

all sessions are kept in a map, so we can get the total count according to the map size

hycdong (approved):

Shuo-Jia (approved):

PR #901, opened 2021-09-03
feat: support nfs server rate limit

Shuo-Jia:
In some test, we found that the influence usually caused by nfs_server side under `learning` copy checkpoint. this pr support the nfs server send rate limit.

## Note
- limit server send rate may casuse the rpc timeout,so this pr update `nfs rpc timeout` from `10s` to `100s`
- update `max_send_rate` via admin-cli [server-config ](https://github.com/pegasus-kv/admin-cli/blob/main/cmd/server_config.go) may be block for a long time, this pr update the task priority
- for compatibility with `remote-command`, this pr support it too

### DIFF
--- a/src/nfs/nfs_client_impl.cpp
+++ b/src/nfs/nfs_client_impl.cpp
@@ -75,7 +75,7 @@ DSN_DEFINE_int32("nfs",
                  "maximum retry count when copy failed");
 DSN_DEFINE_int32("nfs",
                  rpc_timeout_ms,
-                 200000,
+                 100000,
                  "rpc timeout in milliseconds for nfs copy, "
                  "0 means use default timeout of rpc engine");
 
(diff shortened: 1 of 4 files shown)


src/nfs/nfs_server_impl.cpp
```diff
@@ -36,12 +36,16 @@
 #include <sys/stat.h>
 #include <dsn/utility/filesystem.h>
 #include <dsn/tool-api/async_calls.h>
+#include <dsn/tool-api/command_manager.h>
```
levy5307, line 39: remove this line

src/nfs/nfs_client_impl.cpp
```diff
@@ -75,7 +75,7 @@ DSN_DEFINE_int32("nfs",
                  "maximum retry count when copy failed");
 DSN_DEFINE_int32("nfs",
                  rpc_timeout_ms,
-                 10000,
+                 100000,
```
levy5307, line 78: 1e6

levy5307 (approved):

hycdong (approved):

PR #903, opened 2021-09-06
refactor: implement execute_balance to refactor greedy_load_balancer

levy5307:


src/meta/greedy_load_balancer.cpp
```diff
@@ -903,102 +903,96 @@ void greedy_load_balancer::greedy_balancer(const bool balance_checker)
...
-            return;
-        }
-        if (!t_migration_result->empty()) {
-            ddebug_f("migration count of move primary = {}", t_migration_result->size());
-            return;
-        }
+    if (!execute_balance(apps,
+                         false,
+                         true,
+                         true,
```
hycdong, line 990: How about adding comment for those boolean params? such as:
```
execute_balance(apps,
                false, /*balance_checker*/
                true, /*balance_in_turn*/
                true, /*only_move_primary*/
                ......
)
```

src/meta/greedy_load_balancer.cpp
```diff
@@ -453,7 +453,7 @@ bool greedy_load_balancer::copy_primary_per_app(const std::shared_ptr<app_state>
     return true;
 }
 
-bool greedy_load_balancer::copy_secondary_per_app(const std::shared_ptr<app_state> &app)
+bool greedy_load_balancer::copy_secondary(const std::shared_ptr<app_state> &app, bool place_holder)
```
hycdong, line 456: What is the `place_holder` used for?

levy5307: The callback in `execute_balance` need two parameters. So this function needs two parameters also, and the secondary one is useless

hycdong (approved):

src/meta/greedy_load_balancer.cpp
```diff
@@ -903,102 +903,95 @@ void greedy_load_balancer::greedy_balancer(const bool balance_checker)
...
     }
 
     // we seperate the primary/secondary balancer for 2 reasons:
     // 1. globally primary balancer may make secondary unbalanced
     // 2. in one-by-one mode, a secondary balance decision for an app may be prior than
     // another app's primary balancer if not seperated.
+    execute_balance(apps,
+                    balance_checker,
+                    _balancer_in_turn,
+                    _only_move_primary,
```
Shuo-Jia, line 928: if `_balancer_in_turn`,` _only_move_primary` are global, so why need pass it? why `_only_primary_balancer` don't like this?

src/meta/greedy_load_balancer.cpp
```diff
@@ -903,102 +903,95 @@ void greedy_load_balancer::greedy_balancer(const bool balance_checker)
...
-            ddebug_f("skip to do secondary balance for the ignored app[{}]", app->get_logname());
+            ddebug_f("skip to do balance for the ignored app[{}]", app->get_logname());
             continue;
         }
-
         if (app->status != app_status::AS_AVAILABLE || app->is_bulk_loading || app->splitting())
             continue;
 
-        bool enough_information = copy_secondary_per_app(app);
+        bool enough_information = callback(app, only_move_primary);
```
Shuo-Jia, line 964: name `callback` is too general, name it `balance`?

src/meta/greedy_load_balancer.cpp
levy5307: cluster load balance will pass value different with global value.

src/meta/greedy_load_balancer.cpp
levy5307: OK, I will change it in next pull request

src/meta/greedy_load_balancer.cpp
Shuo-Jia: why need different? https://github.com/XiaoMi/rdsn/pull/903/files/9b3d0611015a31ae4d9bec75c137e9c71f5cb1e9#diff-ea29ba8c2a8fee984249d3c2431770e76a92a6239971ebb98e21c057017592c5

Shuo-Jia (approved):

PR #908, opened 2021-09-08
refactor: add class ford_fulkerson

levy5307:
add class `ford_fulkerson` used for primary load balance

### DIFF
--- a/src/meta/greedy_load_balancer.cpp
+++ b/src/meta/greedy_load_balancer.cpp
@@ -650,6 +650,100 @@ bool greedy_load_balancer::calc_disk_load(app_id id,
     }
 }
 
+struct flow_path
+{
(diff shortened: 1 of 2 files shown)


src/meta/greedy_load_balancer.cpp
```diff
@@ -650,6 +650,101 @@ bool greedy_load_balancer::calc_disk_load(app_id id,
     }
 }
 
+struct flow_path
```
hycdong, line 653: Why would you define the struct and class in `.cpp` file, not in `.h`?

levy5307: Because other files don't need them

Shuo-Jia (approved):

hycdong (approved):

PR #916, opened 2021-09-22
feat(load-balance): update selection strategy for max load disk

hycdong:
In cluster load balance, we will try to migrate partition from `max_node` who has the max replica count to `min_node`, then we would like to pick up a `max_load_disk` who has the max replica count in `max_node`. 
The old pickup strategy, if different disks hold same replica count, it will chosse the first one.
```
max_node:                   |               min_node:
ssd1: 2.1, 2.2              |                2.1, 2.2, 2.5
ssd2: 2.3, 2.4              |
ssd3: 2.6                   |
...
```
Just as the example shows, `ssd1` will be choosen. However, 2.1 and 2.2 have already exsited in min_node, so there won't be any partition to be moved.

This pr fixs the problem above. We use `get_max_load_disk_set` to get a disk set. `ssd1` and `ssd2` will be included in the set. Then `pick_up_move` will choose one of them randomly.

### DIFF
--- a/src/meta/greedy_load_balancer.cpp
+++ b/src/meta/greedy_load_balancer.cpp
@@ -102,6 +102,14 @@ void get_min_max_set(const std::map<rpc_address, uint32_t> &node_count_map,
     }
 }
 
+template <typename S>
+auto select_random(const S &s, size_t n)
+{
+    auto it = std::begin(s);
+    std::advance(it, n);
+    return it;
+}
+
 greedy_load_balancer::greedy_load_balancer(meta_service *_svc)
     : server_load_balancer(_svc),
       _ctrl_balancer_ignored_apps(nullptr),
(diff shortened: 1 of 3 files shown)


src/meta/greedy_load_balancer.cpp
```diff
@@ -1244,59 +1253,69 @@ bool greedy_load_balancer::pick_up_move(const cluster_migration_info &cluster_in
...
-    get_max_load_disk(
-        cluster_info, max_nodes, app_id, max_load_node, max_load_disk, max_load_partitions);
-
+    std::set<app_disk_info> max_load_disk_set;
+    get_max_load_disk_set(cluster_info, max_nodes, app_id, max_load_disk_set);
+    if (max_load_disk_set.empty()) {
+        return false;
+    }
+    auto index = rand() % max_load_disk_set.size();
+    auto max_load_disk = *select_random(max_load_disk_set, index);
```
levy5307, line 1262: what about choose in sequence to find one which can move?

levy5307 (approved):

Shuo-Jia (approved):

PR #926, opened 2021-09-27
refactor: refactor move_primary in greedy_load_balancer

levy5307:


Smityz (approved):

src/meta/greedy_load_balancer.cpp
```diff
@@ -784,105 +784,112 @@ void ford_fulkerson::update_flow(int pos,
...
     }
-    return true;
+
+    dassert_f(selected != potential_moving.end(),
+              "can't find gpid to move from({}) to({})",
+              from.to_string(),
+              to.to_string());
+    auto res = *selected;
+    potential_moving.erase(selected);
+    return res;
```
Shuo-Jia, line 892: directly return `*selected`?

levy5307: We can't do it like this, because selected will be removed by `potential_moving.erase(selected);`

Shuo-Jia: So the remove should't put here, the function only `select`, better add remove logic, better put out

levy5307: Yes, you are right. But in set of cpp, it doesn't support to move *selected from potential_moving. It only support remove iter from it.

Shuo-Jia (approved):

PR #932, opened 2021-10-08
feat: support single replica

empiredan:
As is well known, it's not reliable enough for most replication protocols that a table has only one replica, or just two replicas. However, considering the cost, some of our customers choose to set single replica, or double replicas for a table.

In fact, [PacificA Protocol](https://www.microsoft.com/en-us/research/wp-content/uploads/2008/02/tr-2008-25.pdf) can support single replica or double replicas well. Therefore, to support single replica, we can just eliminate the restriction while creating a table.

On the other hand, as described in https://github.com/apache/incubator-pegasus/issues/820, now `greedy_load_balancer` cannot support single replica or double replicas well, since if a cluster has only one replica server, or two replica servers, meta server will core dump. By this PR,  meta server will run normally when a cluster a cluster has one or two replica servers, especially when `min_live_node_count_for_unfreeze` is set to 1.

### DIFF
--- a/src/client/replication_ddl_client.cpp
+++ b/src/client/replication_ddl_client.cpp
@@ -128,8 +128,8 @@ dsn::error_code replication_ddl_client::create_app(const std::string &app_name,
         return ERR_INVALID_PARAMETERS;
     }
 
-    if (replica_count < 2) {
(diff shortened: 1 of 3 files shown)


levy5307:
LGTM

PR #938, opened 2021-10-19
feat: support read throttling by size

Smityz:
Ref: https://github.com/apache/incubator-pegasus/issues/830


hycdong:
I suggest splitting this pull request into 3 parts:
1.  update with `rpc_holder`, and add unit test for it.
2.  add `token_bucket_throttling_controller`
3.  add other logic related with read size throttling

Besides, I notice that you implement `token_bucket_throttling_controller` under `utils` folder, but `token_bucket_throttling_controller` contains logic related with read size, such as function `validate`, variety `_partition_count` etc.
If the `token_bucket_throttling_controller` is a util class, it is supposed NOT contain content about its upper logic, otherwise, it is not suitable under `util` folder.

Smityz:
has merged by split prs

PR #950, opened 2021-11-01
refactor: add an interface for checking if the bucket is empty

Smityz:
I add an interface to check if the bucket is empty, so we don't need to use `get_bucket(1)` to do these things.
BTW, I add a group of unit tests to ensure the rate is always > 0

Ref: https://github.com/apache/incubator-pegasus/pull/829#discussion_r740053407


Shuo-Jia (approved):

src/utils/test/token_bucket_throttling_controller_test.cpp
```diff
@@ -109,7 +110,8 @@ class token_bucket_throttling_controller_test : public ::testing::Test
         for (int i = 0; i < 100000; i++) {
             token_bucket->consumeWithBorrowAndWait(
                 1, throttle_limit / partition_count * 0.8, throttle_limit / partition_count * 1.0);
-            if (!cntl->get_token(1)) {
+            cntl->get_token(1);
```
levy5307, line 113: why did you add `cntl->get_token(1)` here?

Smityz: I just split one function to two functions

levy5307: I know, but didn't the function name change to cosume_token?

Smityz: Fix later

include/dsn/utils/token_bucket_throttling_controller.h
```diff
@@ -45,7 +45,10 @@ class token_bucket_throttling_controller
     // return ture means you can get token
     // return false means the bucket is already empty, but the token is borrowed from future.
     // non-blocking
-    bool get_token(int32_t request_units);
+    bool consume_token(int32_t request_units);
+
+    // if the bucket has no tokens, return false
+    bool available();
```
levy5307, line 51:
```
bool available() const;
```

levy5307 (approved):

Shuo-Jia (approved):

PR #957, opened 2021-11-08
fix: partitions are unbalanced after the table is created

ZhongChaoqiang:
What problem does this PR solve?
partitions are unbalanced after the table is created
For details, see apache/incubator-pegasus#832

Check List
Tests

Manual test (add detailed scripts or steps below)
1. start a onebox with 5 repicaserver: ./run start_onebox -r 5
2. create an app with 100 partitions : create test100 -p 100
3. to view the balancer of the test100 app: app test100 -d

### DIFF
--- a/src/meta/partition_guardian.cpp
+++ b/src/meta/partition_guardian.cpp
@@ -658,7 +658,8 @@ void partition_guardian::finish_cure_proposal(meta_view &view,
             np->newly_remove_primary(gpid.get_app_id(), false);
         } else if (act.type == config_type::CT_UPGRADE_TO_PRIMARY) {
             np->newly_remove_primary(gpid.get_app_id(), true);
-        } else if (act.type == config_type::CT_UPGRADE_TO_SECONDARY) {
+        } else if (act.type == config_type::CT_UPGRADE_TO_SECONDARY ||
+                   act.type == config_type::CT_ADD_SECONDARY) {
             np->newly_remove_partition(gpid.get_app_id());
         }
     }


src/meta/partition_guardian.cpp
```diff
@@ -658,7 +658,8 @@ void partition_guardian::finish_cure_proposal(meta_view &view,
             np->newly_remove_primary(gpid.get_app_id(), false);
         } else if (act.type == config_type::CT_UPGRADE_TO_PRIMARY) {
             np->newly_remove_primary(gpid.get_app_id(), true);
-        } else if (act.type == config_type::CT_UPGRADE_TO_SECONDARY) {
+        } else if (act.type == config_type::CT_UPGRADE_TO_SECONDARY ||
+                   act.type == config_type::CT_ADD_SECONDARY) {
```
Smityz, line 662:
```suggestion
        } else if (act.type == config_type::CT_UPGRADE_TO_SECONDARY){
            np->newly_remove_partition(gpid.get_app_id());
        } else if (act.type == config_type::CT_ADD_SECONDARY){
            np->newly_remove_partition(gpid.get_app_id());
        }
```
It's more clearly

levy5307: I don't think so. It doesn't make sense

levy5307 (approved):

src/meta/partition_guardian.cpp
Smityz: ref:
https://github.com/XiaoMi/rdsn/blob/3dedd1e98816d639821ccc9b0139976230bc670c/src/meta/partition_guardian.cpp#L657-L663

you can refactor it using `switch` or keep the original style.

Smityz: okay

Smityz (approved):

PR #963, opened 2021-11-10
feat: restrict the replication factor while creating app

empiredan:
While creating app on the side of the meta server, currently the replication factor(i.e. replica count) is simply checked to just keep it positive. However, this is not enough since a wrong replication factor will result in later failures.

For example, as we've supported single-replica in https://github.com/XiaoMi/rdsn/pull/932, creating an app with 3 replicas by shell in a cluster that just includes a single node will be blocked for long since there're not enough nodes for the extra replicas. 

Another example will be creating an app with a single replica in a cluster that includes 3 nodes. Later then we decide to upgrade this cluster by HA(i.e. high-availability, upgraded without stopping service of the cluster) to a higher version. The upgrade process will be blocked in the phase of moving the only replica of that app out to another node, since the node that deployed the only replica is stopped.

Also, If there are 2 alive replica servers (with 1 dead), creating an app with 3 replicas will end up with failure. Creating an app in a 3-node cluster with 5 replicas is nonsense. An excessive replication factor is unreasonable for a storage system, too.

Based on the problems illustrated above, I think the replica factor should be restricted further by the following rules.

### Replication factor should not exceed the number of alive replica servers
It's obvious that there should be enough replica servers to put the replicas. Once there're extra replicas that cannot be placed anywhere, creating app will fail.

###  Replication factor has a minimum
Actually creating an app with a single replica in a 3-node cluster is eccentric. Therefore, A minimum value should be set for the replication factor. For example, the minimum for a 3-node cluster can be set to 3. This minimum replication factor will be a configuration for the cluster which is named as `min_allowed_replica_count`.

###  Replication factor has a maximum
Similarly, an excessive replication factor is invalid for a cluster. It'll be restricted by another configuration, named `max_allowed_replica_count`. Any attempt that sets an app with a replication factor exceeding `max_allowed_replica_count` will be rejected. Typically, this value is set to 3, since 4 or more replicas are unusual.


src/meta/server_state.cpp
```diff
@@ -2864,5 +2883,79 @@ void server_state::clear_app_envs(const app_env_rpc &env_rpc)
...
+    bool is_valid = validate_target_max_replica_count_internal(
+        max_replica_count, alive_node_count, &hint_message);
+    if (!is_valid) {
+        dwarn_f("target max replica count is invalid: message={}", hint_message);
+    }
+
+    return is_valid;
+}
+
+int32_t server_state::get_alive_node_count_unlocked()
```
hycdong, line 2949: meta_service has a local variety called `_alive_set`, you can get alive node count through it.
https://github.com/XiaoMi/rdsn/blob/289eb4609ae781e3bcc6bbc6bfeae64dfd8fa785/src/meta/meta_service.h#L269

src/meta/server_state.cpp
```diff
@@ -2864,5 +2883,79 @@ void server_state::clear_app_envs(const app_env_rpc &env_rpc)
...
+bool validate_target_max_replica_count_internal(int32_t max_replica_count,
+                                                int32_t alive_node_count,
+                                                std::string *hint_message)
+{
+    dassert_f(FLAGS_min_allowed_replica_count <= FLAGS_max_allowed_replica_count,
+              "min allowed replica count({}) should be <= max allowed replica count({})",
+              FLAGS_min_allowed_replica_count,
+              FLAGS_max_allowed_replica_count);
+
+    if (alive_node_count < 1) {
```
hycdong, line 2898: Meta server has a state called `freeze`, if alive node is not enough, meta server will freeze itself, in this situation, `create_app` will be rejected. 
https://github.com/XiaoMi/rdsn/blob/289eb4609ae781e3bcc6bbc6bfeae64dfd8fa785/src/meta/meta_service.cpp#L97
You can add meta_function_level on the top of `create_app`, like:
https://github.com/XiaoMi/rdsn/blob/289eb4609ae781e3bcc6bbc6bfeae64dfd8fa785/src/meta/meta_service.cpp#L724-L725

hycdong:
Thanks for your pull request for adding replica count check while creating table~
In this pull request, you adds three rules for replica count:
1. replica count should not exceed alive node count
2. replica count should not below min_replica_count
3. replica count should not exceed max_replica_count
I have a different option with your third rule, I think system should not restrict its table max replica count for scalability. Besides, I think it's unnecessary to make `min_replica_count` as an option, in my view, `min_replica_count` will always be 0.

src/meta/server_state.cpp
```diff
@@ -56,6 +56,24 @@ using namespace dsn;
 namespace dsn {
 namespace replication {
 
+DSN_DEFINE_int32("meta_server",
```
levy5307, line 59:
```
DSN_DEFINE_uint32("meta_server",
                 max_allowed_replica_count,
                 3,
                 "max allowed replica count for arbitrary number of nodes in a cluster");

DSN_DEFINE_validator(max_allowed_replica_count, [](int32_t allowed_replica_count) -> bool {
    return allowed_replica_count <= 3;
});
```

src/meta/server_state.cpp
```diff
@@ -2864,5 +2883,79 @@ void server_state::clear_app_envs(const app_env_rpc &env_rpc)
                    new_envs.c_str());
         });
 }
+
+namespace {
+
+bool validate_target_max_replica_count_internal(int32_t max_replica_count,
+                                                int32_t alive_node_count,
+                                                std::string *hint_message)
```
hycdong, line 2891: I recommend using `std::string &hint_message` instead.

src/meta/server_state.cpp
```diff
@@ -2864,5 +2883,79 @@ void server_state::clear_app_envs(const app_env_rpc &env_rpc)
...
+{
+    dassert_f(FLAGS_min_allowed_replica_count <= FLAGS_max_allowed_replica_count,
```
Shuo-Jia, line 2893: just return error, but not dassert, make meta crush is not good

empiredan:
> Thanks for your pull request for adding replica count check while creating table~ In this pull request, you adds three rules for replica count:
> 
> 1. replica count should not exceed alive node count
> 2. replica count should not below min_replica_count
> 3. replica count should not exceed max_replica_count
> 
> I have a different option with your third rule, I think system should not restrict its table max replica count for scalability. Besides, I think it's unnecessary to make `min_replica_count` as an option, in my view, `min_replica_count` will always be 0. Besides, I add some comments for this pr.

Yeah, I agree that we should provide scalability and flexibility for the system. The purpose of introducing `max_allowed_replica_count` is to avoid unreasonable replication factor. A big replication factor may lead to poor behavior of the system. We can increase this maximum, but I think we'd better give an upper limit to prevent users from creating an app with 10 replicas. 

Introducing `min_allowed_replica_count` is motivated from the example illustrated in the description above:
```
Another example will be creating an app with a single replica in a cluster that includes 3 nodes. 

Later then, we decide to make a high-availability upgrade (without stopping service of the cluster) for this cluster. 

The upgrade process will be blocked in the phase of moving the only replica of that app out to another node, since the node that deployed the only replica is stopped.
```

Considering this example, I think we can set the minimum replication factor as 3 for a cluster that has 3 or more replica servers.

src/meta/server_state.cpp
empiredan: Could we keep the type of `FLAGS_ max_allowed_replica_count` consistent with `max_replica_count` whose type is `int32` ? So we can compare them without explicitly/implicitly conversion between `int32` and `uint32`.

On the other hand, could we reserve `allowed_replica_count  > 0` ? Since zero replica count is invalid.

src/meta/server_state.cpp
empiredan: Actually this check has been done during validator, so I think we can just remove it ?
```C++
DSN_DEFINE_validator(min_allowed_replica_count, [](int32_t allowed_replica_count) -> bool {
    return allowed_replica_count > 0 && allowed_replica_count <= FLAGS_max_allowed_replica_count;
});
```

hycdong:
> > Thanks for your pull request for adding replica count check while creating table~ In this pull request, you adds three rules for replica count:
> > 
> > 1. replica count should not exceed alive node count
> > 2. replica count should not below min_replica_count
> > 3. replica count should not exceed max_replica_count
> > 
> > I have a different option with your third rule, I think system should not restrict its table max replica count for scalability. Besides, I think it's unnecessary to make `min_replica_count` as an option, in my view, `min_replica_count` will always be 0. Besides, I add some comments for this pr.
> 
> Yeah, I agree that we should provide scalability and flexibility for the system. The purpose of introducing `max_allowed_replica_count` is to avoid unreasonable replication factor. A big replication factor may lead to poor behavior of the system. We can increase this maximum, but I think we'd better give an upper limit to prevent users from creating an app with 10 replicas.
> 
> Introducing `min_allowed_replica_count` is motivated from the example illustrated in the description above:
> 
> ```
> Another example will be creating an app with a single replica in a cluster that includes 3 nodes. 
> 
> Later then, we decide to make a high-availability upgrade (without stopping service of the cluster) for this cluster. 
> 
> The upgrade process will be blocked in the phase of moving the only replica of that app out to another node, since the node that deployed the only replica is stopped.
> ```
> 
> Considering this example, I think we can set the minimum replication factor as 3 for a cluster that has 3 or more replica servers.

Okay, your comment persuades me~ `min_replica_count` is used to set the cluster table min replica count, `max_replica_count` is used to avoid user unreasonable replica count. How about make `max_replica_count` as a const value, not a configurable option. For one hand, this value won't be updated in normal cases, for other hands, it will simply the `DSN_DEFINE_validator` check for `min_replica_count`.

empiredan:
> > > Thanks for your pull request for adding replica count check while creating table~ In this pull request, you adds three rules for replica count:
> > > 
> > > 1. replica count should not exceed alive node count
> > > 2. replica count should not below min_replica_count
> > > 3. replica count should not exceed max_replica_count
> > > 
> > > I have a different option with your third rule, I think system should not restrict its table max replica count for scalability. Besides, I think it's unnecessary to make `min_replica_count` as an option, in my view, `min_replica_count` will always be 0. Besides, I add some comments for this pr.
> > 
> > 
> > Yeah, I agree that we should provide scalability and flexibility for the system. The purpose of introducing `max_allowed_replica_count` is to avoid unreasonable replication factor. A big replication factor may lead to poor behavior of the system. We can increase this maximum, but I think we'd better give an upper limit to prevent users from creating an app with 10 replicas.
> > Introducing `min_allowed_replica_count` is motivated from the example illustrated in the description above:
> > ```
> > Another example will be creating an app with a single replica in a cluster that includes 3 nodes. 
> > 
> > Later then, we decide to make a high-availability upgrade (without stopping service of the cluster) for this cluster. 
> > 
> > The upgrade process will be blocked in the phase of moving the only replica of that app out to another node, since the node that deployed the only replica is stopped.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Considering this example, I think we can set the minimum replication factor as 3 for a cluster that has 3 or more replica servers.
> 
> Okay, your comment persuades me~ `min_replica_count` is used to set the cluster table min replica count, `max_replica_count` is used to avoid user unreasonable replica count. How about make `max_replica_count` as a const value, not a configurable option. For one hand, this value won't be updated in normal cases, for other hands, it will simply the `DSN_DEFINE_validator` check for `min_replica_count`.

A good idea ! How about set `max_allowed_replica_count` as 3 ? Can this value satisfy all of your scenarios ?

hycdong:
> > > > Thanks for your pull request for adding replica count check while creating table~ In this pull request, you adds three rules for replica count:
> > > > 
> > > > 1. replica count should not exceed alive node count
> > > > 2. replica count should not below min_replica_count
> > > > 3. replica count should not exceed max_replica_count
> > > > 
> > > > I have a different option with your third rule, I think system should not restrict its table max replica count for scalability. Besides, I think it's unnecessary to make `min_replica_count` as an option, in my view, `min_replica_count` will always be 0. Besides, I add some comments for this pr.
> > > 
> > > 
> > > Yeah, I agree that we should provide scalability and flexibility for the system. The purpose of introducing `max_allowed_replica_count` is to avoid unreasonable replication factor. A big replication factor may lead to poor behavior of the system. We can increase this maximum, but I think we'd better give an upper limit to prevent users from creating an app with 10 replicas.
> > > Introducing `min_allowed_replica_count` is motivated from the example illustrated in the description above:
> > > ```
> > > Another example will be creating an app with a single replica in a cluster that includes 3 nodes. 
> > > 
> > > Later then, we decide to make a high-availability upgrade (without stopping service of the cluster) for this cluster. 
> > > 
> > > The upgrade process will be blocked in the phase of moving the only replica of that app out to another node, since the node that deployed the only replica is stopped.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Considering this example, I think we can set the minimum replication factor as 3 for a cluster that has 3 or more replica servers.
> > 
> > 
> > Okay, your comment persuades me~ `min_replica_count` is used to set the cluster table min replica count, `max_replica_count` is used to avoid user unreasonable replica count. How about make `max_replica_count` as a const value, not a configurable option. For one hand, this value won't be updated in normal cases, for other hands, it will simply the `DSN_DEFINE_validator` check for `min_replica_count`.
> 
> A good idea ! How about set `max_allowed_replica_count` as 3 ? Can this value satisfy all of your scenarios ?

In most our scenarios, replica_count is 3. As our discussion above, `max_allowed_replica_count` should be greater than the normal replica_count. If we set `max_allowed_replica_count` as 3, then if we would like to provide 5 replica, which is also a reasonable replica count, we should update code. As a result, I suggest seting `max_allowed_replica_count` as 10, or even greater than 10.

empiredan:
> > > > > Thanks for your pull request for adding replica count check while creating table~ In this pull request, you adds three rules for replica count:
> > > > > 
> > > > > 1. replica count should not exceed alive node count
> > > > > 2. replica count should not below min_replica_count
> > > > > 3. replica count should not exceed max_replica_count
> > > > > 
> > > > > I have a different option with your third rule, I think system should not restrict its table max replica count for scalability. Besides, I think it's unnecessary to make `min_replica_count` as an option, in my view, `min_replica_count` will always be 0. Besides, I add some comments for this pr.
> > > > 
> > > > 
> > > > Yeah, I agree that we should provide scalability and flexibility for the system. The purpose of introducing `max_allowed_replica_count` is to avoid unreasonable replication factor. A big replication factor may lead to poor behavior of the system. We can increase this maximum, but I think we'd better give an upper limit to prevent users from creating an app with 10 replicas.
> > > > Introducing `min_allowed_replica_count` is motivated from the example illustrated in the description above:
> > > > ```
> > > > Another example will be creating an app with a single replica in a cluster that includes 3 nodes. 
> > > > 
> > > > Later then, we decide to make a high-availability upgrade (without stopping service of the cluster) for this cluster. 
> > > > 
> > > > The upgrade process will be blocked in the phase of moving the only replica of that app out to another node, since the node that deployed the only replica is stopped.
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > Considering this example, I think we can set the minimum replication factor as 3 for a cluster that has 3 or more replica servers.
> > > 
> > > 
> > > Okay, your comment persuades me~ `min_replica_count` is used to set the cluster table min replica count, `max_replica_count` is used to avoid user unreasonable replica count. How about make `max_replica_count` as a const value, not a configurable option. For one hand, this value won't be updated in normal cases, for other hands, it will simply the `DSN_DEFINE_validator` check for `min_replica_count`.
> > 
> > 
> > A good idea ! How about set `max_allowed_replica_count` as 3 ? Can this value satisfy all of your scenarios ?
> 
> In most our scenarios, replica_count is 3. As our discussion above, `max_allowed_replica_count` should be greater than the normal replica_count. If we set `max_allowed_replica_count` as 3, then if we would like to provide 5 replica, which is also a reasonable replica count, we should update code. As a result, I suggest seting `max_allowed_replica_count` as 10, or even greater than 10.

Ok, how about setting `max_allowed_replica_count` as 15 (an odd number, though there's no particular reason) ?

hycdong:
> > > > > > Thanks for your pull request for adding replica count check while creating table~ In this pull request, you adds three rules for replica count:
> > > > > > 
> > > > > > 1. replica count should not exceed alive node count
> > > > > > 2. replica count should not below min_replica_count
> > > > > > 3. replica count should not exceed max_replica_count
> > > > > > 
> > > > > > I have a different option with your third rule, I think system should not restrict its table max replica count for scalability. Besides, I think it's unnecessary to make `min_replica_count` as an option, in my view, `min_replica_count` will always be 0. Besides, I add some comments for this pr.
> > > > > 
> > > > > 
> > > > > Yeah, I agree that we should provide scalability and flexibility for the system. The purpose of introducing `max_allowed_replica_count` is to avoid unreasonable replication factor. A big replication factor may lead to poor behavior of the system. We can increase this maximum, but I think we'd better give an upper limit to prevent users from creating an app with 10 replicas.
> > > > > Introducing `min_allowed_replica_count` is motivated from the example illustrated in the description above:
> > > > > ```
> > > > > Another example will be creating an app with a single replica in a cluster that includes 3 nodes. 
> > > > > 
> > > > > Later then, we decide to make a high-availability upgrade (without stopping service of the cluster) for this cluster. 
> > > > > 
> > > > > The upgrade process will be blocked in the phase of moving the only replica of that app out to another node, since the node that deployed the only replica is stopped.
> > > > > ```
> > > > > 
> > > > > 
> > > > >     
> > > > >       
> > > > >     
> > > > > 
> > > > >       
> > > > >     
> > > > > 
> > > > >     
> > > > >   
> > > > > Considering this example, I think we can set the minimum replication factor as 3 for a cluster that has 3 or more replica servers.
> > > > 
> > > > 
> > > > Okay, your comment persuades me~ `min_replica_count` is used to set the cluster table min replica count, `max_replica_count` is used to avoid user unreasonable replica count. How about make `max_replica_count` as a const value, not a configurable option. For one hand, this value won't be updated in normal cases, for other hands, it will simply the `DSN_DEFINE_validator` check for `min_replica_count`.
> > > 
> > > 
> > > A good idea ! How about set `max_allowed_replica_count` as 3 ? Can this value satisfy all of your scenarios ?
> > 
> > 
> > In most our scenarios, replica_count is 3. As our discussion above, `max_allowed_replica_count` should be greater than the normal replica_count. If we set `max_allowed_replica_count` as 3, then if we would like to provide 5 replica, which is also a reasonable replica count, we should update code. As a result, I suggest seting `max_allowed_replica_count` as 10, or even greater than 10.
> 
> Ok, how about setting `max_allowed_replica_count` as 15 (an odd number, though there's no particular reason) ?

I think it's okay to set  `max_allowed_replica_count` as 15~

src/meta/test/meta_app_operation_test.cpp
```diff
@@ -114,32 +119,71 @@ TEST_F(meta_app_operation_test, create_app)
...
+                 {APP_NAME, 4, 3, 2, 0, 1, false, app_status::AS_DROPPED, ERR_OK},
+                 {APP_NAME, 4, 3, 2, 0, 1, true, app_status::AS_INVALID, ERR_OK}};
+
+    clear_nodes();
+    std::vector<rpc_address> nodes = ensure_enough_alive_nodes(3);
+
+    // the meta function level will become freezed once
+    // alive_nodes * 100 < total_nodes * node_live_percentage_threshold_for_update
+    // even if alive_nodes >= min_live_node_count_for_unfreeze
+    set_node_live_percentage_threshold_for_update(30);
```
levy5307, line 164: auto reserved_min_allowed_replica_count = FLAGS_min_allowed_replica_count;

src/meta/test/meta_app_operation_test.cpp
```diff
@@ -114,32 +119,71 @@ TEST_F(meta_app_operation_test, create_app)
...
         }
-        auto err = create_app_test(test.partition_count, test.replica_count, test.success_if_exist);
+        auto err = create_app_test(
+            test.partition_count, test.replica_count, test.success_if_exist, test.app_name);
         ASSERT_EQ(err, test.expected_err);
+
+        _ms->set_node_state(nodes, true);
     }
+
+    FLAGS_min_allowed_replica_count = 1;
```
levy5307, line 186: FLAGS_min_allowed_replica_count = reserved_min_allowed_replica_count;

levy5307 (dismissed):

Smityz:
LGTM

src/meta/server_state.cpp
```diff
@@ -2864,5 +2881,55 @@ void server_state::clear_app_envs(const app_env_rpc &env_rpc)
...
+
+bool server_state::validate_target_max_replica_count(int32_t max_replica_count)
+{
+    auto alive_node_count = static_cast<int32_t>(_meta_svc->get_alive_node_count());
+
+    std::string hint_message;
+    bool is_valid = validate_target_max_replica_count_internal(
+        max_replica_count, alive_node_count, hint_message);
+    if (!is_valid) {
+        dwarn_f("target max replica count is invalid: message={}", hint_message);
```
hycdong, line 2928: I suggest using derror_f instead.

src/meta/server_state.cpp
```diff
@@ -1072,7 +1083,13 @@ void server_state::create_app(dsn::message_ex *msg)
                opt.replica_count == exist_app.max_replica_count;
     };
 
-    if (request.options.partition_count <= 0 || request.options.replica_count <= 0) {
+    auto level = _meta_svc->get_function_level();
+    if (level <= meta_function_level::fl_freezed) {
+        dwarn_f("current meta function level is freezed since there are too few alive nodes");
```
hycdong, line 1088: I suggest using derror_f instead.

src/meta/test/meta_app_operation_test.cpp
```diff
@@ -114,32 +119,73 @@ TEST_F(meta_app_operation_test, create_app)
     // - create succeed with success_if_exist=true
```
hycdong, line 119: It will be better to add the unit cases this pr adds in the case comments.

empiredan: Yes, I think this is necessary since this PR introduces lots of cases.

hycdong (approved):
LGTM~ By the way, I suggest using 'feat' rather than 'fix' to describe this pull request, because this pull request actually provides feature~

empiredan:
> LGTM~ By the way, I suggest using 'feat' rather than 'fix' to describe this pull request, because this pull request actually provides feature~

Sure, I'll change the type

src/meta/server_state.cpp
```diff
@@ -2864,5 +2881,55 @@ void server_state::clear_app_envs(const app_env_rpc &env_rpc)
...
+    }
+
+    if (max_replica_count < FLAGS_min_allowed_replica_count) {
+        hint_message = fmt::format("requested replica count({}) is less than the min "
+                                   "allowed replica count({})",
+                                   max_replica_count,
+                                   FLAGS_min_allowed_replica_count);
+        return false;
+    }
+
```
Shuo-Jia, line 2906: you may don't need to split two `if` to  log the `hit_message`, consider the following whether more simple:
```c++
 if (max_replica_count > max_allowed_replica_count || max_replica_count < FLAGS_min_allowed_replica_count) {
        hint_message = fmt::format("requested replica count({}) must be range within [min={} max={}] ",
                                   max_replica_count,
                                   FLAGS_min_allowed_replica_count
                                   max_allowed_replica_count);
        return false;
    }
```

src/meta/server_state.cpp
```diff
@@ -56,6 +56,17 @@ using namespace dsn;
 namespace dsn {
 namespace replication {
 
+const int32_t max_allowed_replica_count = 15;
```
Shuo-Jia, line 59: Why the `max` is not set `dynamic variable` like `min`?

empiredan: Initially the `max_allowed_replica_count ` was set configurable. Then as discussed with @hycdong above, `max_allowed_replica_count ` has been changed as a const value, not a configurable one, since in normal cases it will not be updated.

Smityz: By the way, the const int is named like `k...`  refer to [google code style](https://google.github.io/styleguide/cppguide.html#Constant_Names)

empiredan: Agreed. I'd considered following google code style such as the camel-case naming of constant which begins with 'k'. On the other hand, however I'd found many constants are named in upper or lower snake case before I turned to follow the existing custom. I think we should gradually unify the naming style.

src/meta/server_state.cpp
empiredan: Ok, I'll merge the both conditions. A range seems more specific.

acelyc111:
> > > > > > > Thanks for your pull request for adding replica count check while creating table~ In this pull request, you adds three rules for replica count:
> > > > > > > 
> > > > > > > 1. replica count should not exceed alive node count
> > > > > > > 2. replica count should not below min_replica_count
> > > > > > > 3. replica count should not exceed max_replica_count
> > > > > > > 
> > > > > > > I have a different option with your third rule, I think system should not restrict its table max replica count for scalability. Besides, I think it's unnecessary to make `min_replica_count` as an option, in my view, `min_replica_count` will always be 0. Besides, I add some comments for this pr.
> > > > > > 
> > > > > > 
> > > > > > Yeah, I agree that we should provide scalability and flexibility for the system. The purpose of introducing `max_allowed_replica_count` is to avoid unreasonable replication factor. A big replication factor may lead to poor behavior of the system. We can increase this maximum, but I think we'd better give an upper limit to prevent users from creating an app with 10 replicas.
> > > > > > Introducing `min_allowed_replica_count` is motivated from the example illustrated in the description above:
> > > > > > ```
> > > > > > Another example will be creating an app with a single replica in a cluster that includes 3 nodes. 
> > > > > > 
> > > > > > Later then, we decide to make a high-availability upgrade (without stopping service of the cluster) for this cluster. 
> > > > > > 
> > > > > > The upgrade process will be blocked in the phase of moving the only replica of that app out to another node, since the node that deployed the only replica is stopped.
> > > > > > ```
> > > > > > 
> > > > > > 
> > > > > >     
> > > > > >       
> > > > > >     
> > > > > > 
> > > > > >       
> > > > > >     
> > > > > > 
> > > > > >     
> > > > > >   
> > > > > > Considering this example, I think we can set the minimum replication factor as 3 for a cluster that has 3 or more replica servers.
> > > > > 
> > > > > 
> > > > > Okay, your comment persuades me~ `min_replica_count` is used to set the cluster table min replica count, `max_replica_count` is used to avoid user unreasonable replica count. How about make `max_replica_count` as a const value, not a configurable option. For one hand, this value won't be updated in normal cases, for other hands, it will simply the `DSN_DEFINE_validator` check for `min_replica_count`.
> > > > 
> > > > 
> > > > A good idea ! How about set `max_allowed_replica_count` as 3 ? Can this value satisfy all of your scenarios ?
> > > 
> > > 
> > > In most our scenarios, replica_count is 3. As our discussion above, `max_allowed_replica_count` should be greater than the normal replica_count. If we set `max_allowed_replica_count` as 3, then if we would like to provide 5 replica, which is also a reasonable replica count, we should update code. As a result, I suggest seting `max_allowed_replica_count` as 10, or even greater than 10.
> > 
> > 
> > Ok, how about setting `max_allowed_replica_count` as 15 (an odd number, though there's no particular reason) ?
> 
> I think it's okay to set `max_allowed_replica_count` as 15~

I do consider a const and unoptional value of 15 is too large, any new user could create a table with 15 replicas at most without any limitation, it's harmful for the system. If most cases are 3 replicas, why not make it optional, in the rare case if any body want to create a table with some larger replication factor, they can set a larger RF and reboot the metaserver.

hycdong:
> > > > > > > > Thanks for your pull request for adding replica count check while creating table~ In this pull request, you adds three rules for replica count:
> > > > > > > > 
> > > > > > > > 1. replica count should not exceed alive node count
> > > > > > > > 2. replica count should not below min_replica_count
> > > > > > > > 3. replica count should not exceed max_replica_count
> > > > > > > > 
> > > > > > > > I have a different option with your third rule, I think system should not restrict its table max replica count for scalability. Besides, I think it's unnecessary to make `min_replica_count` as an option, in my view, `min_replica_count` will always be 0. Besides, I add some comments for this pr.
> > > > > > > 
> > > > > > > 
> > > > > > > Yeah, I agree that we should provide scalability and flexibility for the system. The purpose of introducing `max_allowed_replica_count` is to avoid unreasonable replication factor. A big replication factor may lead to poor behavior of the system. We can increase this maximum, but I think we'd better give an upper limit to prevent users from creating an app with 10 replicas.
> > > > > > > Introducing `min_allowed_replica_count` is motivated from the example illustrated in the description above:
> > > > > > > ```
> > > > > > > Another example will be creating an app with a single replica in a cluster that includes 3 nodes. 
> > > > > > > 
> > > > > > > Later then, we decide to make a high-availability upgrade (without stopping service of the cluster) for this cluster. 
> > > > > > > 
> > > > > > > The upgrade process will be blocked in the phase of moving the only replica of that app out to another node, since the node that deployed the only replica is stopped.
> > > > > > > ```
> > > > > > > 
> > > > > > > 
> > > > > > >     
> > > > > > >       
> > > > > > >     
> > > > > > > 
> > > > > > >       
> > > > > > >     
> > > > > > > 
> > > > > > >     
> > > > > > >   
> > > > > > > Considering this example, I think we can set the minimum replication factor as 3 for a cluster that has 3 or more replica servers.
> > > > > > 
> > > > > > 
> > > > > > Okay, your comment persuades me~ `min_replica_count` is used to set the cluster table min replica count, `max_replica_count` is used to avoid user unreasonable replica count. How about make `max_replica_count` as a const value, not a configurable option. For one hand, this value won't be updated in normal cases, for other hands, it will simply the `DSN_DEFINE_validator` check for `min_replica_count`.
> > > > > 
> > > > > 
> > > > > A good idea ! How about set `max_allowed_replica_count` as 3 ? Can this value satisfy all of your scenarios ?
> > > > 
> > > > 
> > > > In most our scenarios, replica_count is 3. As our discussion above, `max_allowed_replica_count` should be greater than the normal replica_count. If we set `max_allowed_replica_count` as 3, then if we would like to provide 5 replica, which is also a reasonable replica count, we should update code. As a result, I suggest seting `max_allowed_replica_count` as 10, or even greater than 10.
> > > 
> > > 
> > > Ok, how about setting `max_allowed_replica_count` as 15 (an odd number, though there's no particular reason) ?
> > 
> > 
> > I think it's okay to set `max_allowed_replica_count` as 15~
> 
> @hycdong @empiredan I do consider a const and unoptional value of 15 is too large, any new user could create a table with 15 replicas at most without any limitation, it's harmful for the system. If most cases are 3 replicas, why not make it optional, in the rare case if any body want to create a table with some larger replication factor, they can set a larger RF and reboot the metaserver.

As I mentioned before, I think `system should not restrict its table max replica count for scalability`, but as empiredan's option, it  can avoid user unreasonable replica count. If we set `max_allowed_replica_count` as 3, it somehow affects scalability. It is okay for user to create table with 4 replicas and 5 replicas and so on. Besides, 4 and 5 are not user unreasonable replica count. So I prefer seting it as a bigger value. 
And there are two reasons I recommend it as a const. Firstly, it is easiler to check validation `max_allowed_replica_count` and `min_allowed_replica_count` in validation if one of them is a const value. Secondly, as the  `max_allowed_replica_count` is used to avoid unreasonable replica count, we will hardly update it, it is okay to be a const value, because its value is unreasonable enough.

empiredan:
> > > > > > > > > Thanks for your pull request for adding replica count check while creating table~ In this pull request, you adds three rules for replica count:
> > > > > > > > > 
> > > > > > > > > 1. replica count should not exceed alive node count
> > > > > > > > > 2. replica count should not below min_replica_count
> > > > > > > > > 3. replica count should not exceed max_replica_count
> > > > > > > > > 
> > > > > > > > > I have a different option with your third rule, I think system should not restrict its table max replica count for scalability. Besides, I think it's unnecessary to make `min_replica_count` as an option, in my view, `min_replica_count` will always be 0. Besides, I add some comments for this pr.
> > > > > > > > 
> > > > > > > > 
> > > > > > > > Yeah, I agree that we should provide scalability and flexibility for the system. The purpose of introducing `max_allowed_replica_count` is to avoid unreasonable replication factor. A big replication factor may lead to poor behavior of the system. We can increase this maximum, but I think we'd better give an upper limit to prevent users from creating an app with 10 replicas.
> > > > > > > > Introducing `min_allowed_replica_count` is motivated from the example illustrated in the description above:
> > > > > > > > ```
> > > > > > > > Another example will be creating an app with a single replica in a cluster that includes 3 nodes. 
> > > > > > > > 
> > > > > > > > Later then, we decide to make a high-availability upgrade (without stopping service of the cluster) for this cluster. 
> > > > > > > > 
> > > > > > > > The upgrade process will be blocked in the phase of moving the only replica of that app out to another node, since the node that deployed the only replica is stopped.
> > > > > > > > ```
> > > > > > > > 
> > > > > > > > 
> > > > > > > >     
> > > > > > > >       
> > > > > > > >     
> > > > > > > > 
> > > > > > > >       
> > > > > > > >     
> > > > > > > > 
> > > > > > > >     
> > > > > > > >   
> > > > > > > > Considering this example, I think we can set the minimum replication factor as 3 for a cluster that has 3 or more replica servers.
> > > > > > > 
> > > > > > > 
> > > > > > > Okay, your comment persuades me~ `min_replica_count` is used to set the cluster table min replica count, `max_replica_count` is used to avoid user unreasonable replica count. How about make `max_replica_count` as a const value, not a configurable option. For one hand, this value won't be updated in normal cases, for other hands, it will simply the `DSN_DEFINE_validator` check for `min_replica_count`.
> > > > > > 
> > > > > > 
> > > > > > A good idea ! How about set `max_allowed_replica_count` as 3 ? Can this value satisfy all of your scenarios ?
> > > > > 
> > > > > 
> > > > > In most our scenarios, replica_count is 3. As our discussion above, `max_allowed_replica_count` should be greater than the normal replica_count. If we set `max_allowed_replica_count` as 3, then if we would like to provide 5 replica, which is also a reasonable replica count, we should update code. As a result, I suggest seting `max_allowed_replica_count` as 10, or even greater than 10.
> > > > 
> > > > 
> > > > Ok, how about setting `max_allowed_replica_count` as 15 (an odd number, though there's no particular reason) ?
> > > 
> > > 
> > > I think it's okay to set `max_allowed_replica_count` as 15~
> > 
> > 
> > @hycdong @empiredan I do consider a const and unoptional value of 15 is too large, any new user could create a table with 15 replicas at most without any limitation, it's harmful for the system. If most cases are 3 replicas, why not make it optional, in the rare case if any body want to create a table with some larger replication factor, they can set a larger RF and reboot the metaserver.
> 
> As I mentioned before, I think `system should not restrict its table max replica count for scalability`, but as empiredan's option, it can `avoid user unreasonable replica count`. If we set `max_allowed_replica_count` as 3, it somehow affects scalability. It is okay for user to create table with 4 replicas and 5 replicas and so on. Besides, 4 and 5 are not user unreasonable replica count. So I prefer seting it as a bigger value. And there are two reasons I recommend it as a const. Firstly, it is easiler to check validation `max_allowed_replica_count` and `min_allowed_replica_count` in validation if one of them is a const value. Secondly, as the `max_allowed_replica_count` is used to avoid unreasonable replica count, we will hardly update it, it is okay to be a const value, because its value is unreasonable enough. That is my opinion above, I think we can disucuss the usage of `max_allowed_replica_count` to determine seting it as an optional value or a const value.

@empiredan @acelyc111  Or, can we turn to refer to the implementation of the system that has the similar problem with us ?

In https://github.com/apache/kudu/commit/b39491de611eb7a36be46816c66fa3a85590a656, Todd has explained why kudu uses `max_num_replicas` to restrict the replication factor. Now `max_num_replicas` is set to 7, since higher RF hasn't been proved to be effective by some test cases.  `A well-intentioned user may try to set replication factor to a very high number to get a table replicated onto every server in their cluster`.

Therefore, should we make `max_allowed_replica_count` configurable and its default value a little higher than 3, say, 5 ?

And also, RF is restricted by `min_num_repilcas` . See https://github.com/apache/kudu/commit/6cb15486cdbeacb722e69b81d563b6f05d840650 for details.

As for the related code, please see the following quotes:
```c++
DEFINE_int32(max_num_replicas, 7,
             "Maximum number of replicas that may be specified for a table.");
// Tag as unsafe since we have done very limited testing of higher than 5 replicas.
TAG_FLAG(max_num_replicas, unsafe);
TAG_FLAG(max_num_replicas, runtime);

DEFINE_int32(min_num_replicas, 1,
             "Minimum number of replicas that may be specified when creating "
             "a table: this is to enforce the minimum replication factor for "
             "tables created in a Kudu cluster. For example, setting this flag "
             "to 3 enforces every new table to have at least 3 replicas for "
             "each of its tablets, so there cannot be a data loss when a "
             "single tablet server fails irrecoverably.");
TAG_FLAG(min_num_replicas, advanced);
TAG_FLAG(min_num_replicas, runtime);

...

template<typename RespClass>
Status CatalogManager::ValidateNumberReplicas(const std::string& normalized_table_name,
                                              RespClass* resp, ValidateType type,
                                              const boost::optional<int>& partitions_count,
                                              int num_replicas) {
  if (num_replicas > FLAGS_max_num_replicas) {
    return SetupError(Status::InvalidArgument(
        Substitute("illegal replication factor $0: maximum allowed replication "
                   "factor is $1 (controlled by --max_num_replicas)",
                   num_replicas, FLAGS_max_num_replicas)),
        resp, MasterErrorPB::REPLICATION_FACTOR_TOO_HIGH);
  }
  if (num_replicas < FLAGS_min_num_replicas) {
    return SetupError(Status::InvalidArgument(
        Substitute("illegal replication factor $0: minimum allowed replication "
                   "factor is $1 (controlled by --min_num_replicas)",
            num_replicas, FLAGS_min_num_replicas)),
        resp, MasterErrorPB::ILLEGAL_REPLICATION_FACTOR);
  }
  // Reject create/alter table with even replication factors, unless master flag
  // allow_unsafe_replication_factor is on.
  if (num_replicas % 2 == 0 && !FLAGS_allow_unsafe_replication_factor) {
    return SetupError(Status::InvalidArgument(
        Substitute("illegal replication factor $0: replication factor must be odd",
                   num_replicas)),
        resp, MasterErrorPB::EVEN_REPLICATION_FACTOR);
  }
...
}
```

hycdong:
> > > > > > > > > > Thanks for your pull request for adding replica count check while creating table~ In this pull request, you adds three rules for replica count:
> > > > > > > > > > 
> > > > > > > > > > 1. replica count should not exceed alive node count
> > > > > > > > > > 2. replica count should not below min_replica_count
> > > > > > > > > > 3. replica count should not exceed max_replica_count
> > > > > > > > > > 
> > > > > > > > > > I have a different option with your third rule, I think system should not restrict its table max replica count for scalability. Besides, I think it's unnecessary to make `min_replica_count` as an option, in my view, `min_replica_count` will always be 0. Besides, I add some comments for this pr.
> > > > > > > > > 
> > > > > > > > > 
> > > > > > > > > Yeah, I agree that we should provide scalability and flexibility for the system. The purpose of introducing `max_allowed_replica_count` is to avoid unreasonable replication factor. A big replication factor may lead to poor behavior of the system. We can increase this maximum, but I think we'd better give an upper limit to prevent users from creating an app with 10 replicas.
> > > > > > > > > Introducing `min_allowed_replica_count` is motivated from the example illustrated in the description above:
> > > > > > > > > ```
> > > > > > > > > Another example will be creating an app with a single replica in a cluster that includes 3 nodes. 
> > > > > > > > > 
> > > > > > > > > Later then, we decide to make a high-availability upgrade (without stopping service of the cluster) for this cluster. 
> > > > > > > > > 
> > > > > > > > > The upgrade process will be blocked in the phase of moving the only replica of that app out to another node, since the node that deployed the only replica is stopped.
> > > > > > > > > ```
> > > > > > > > > 
> > > > > > > > > 
> > > > > > > > >     
> > > > > > > > >       
> > > > > > > > >     
> > > > > > > > > 
> > > > > > > > >       
> > > > > > > > >     
> > > > > > > > > 
> > > > > > > > >     
> > > > > > > > >   
> > > > > > > > > Considering this example, I think we can set the minimum replication factor as 3 for a cluster that has 3 or more replica servers.
> > > > > > > > 
> > > > > > > > 
> > > > > > > > Okay, your comment persuades me~ `min_replica_count` is used to set the cluster table min replica count, `max_replica_count` is used to avoid user unreasonable replica count. How about make `max_replica_count` as a const value, not a configurable option. For one hand, this value won't be updated in normal cases, for other hands, it will simply the `DSN_DEFINE_validator` check for `min_replica_count`.
> > > > > > > 
> > > > > > > 
> > > > > > > A good idea ! How about set `max_allowed_replica_count` as 3 ? Can this value satisfy all of your scenarios ?
> > > > > > 
> > > > > > 
> > > > > > In most our scenarios, replica_count is 3. As our discussion above, `max_allowed_replica_count` should be greater than the normal replica_count. If we set `max_allowed_replica_count` as 3, then if we would like to provide 5 replica, which is also a reasonable replica count, we should update code. As a result, I suggest seting `max_allowed_replica_count` as 10, or even greater than 10.
> > > > > 
> > > > > 
> > > > > Ok, how about setting `max_allowed_replica_count` as 15 (an odd number, though there's no particular reason) ?
> > > > 
> > > > 
> > > > I think it's okay to set `max_allowed_replica_count` as 15~
> > > 
> > > 
> > > @hycdong @empiredan I do consider a const and unoptional value of 15 is too large, any new user could create a table with 15 replicas at most without any limitation, it's harmful for the system. If most cases are 3 replicas, why not make it optional, in the rare case if any body want to create a table with some larger replication factor, they can set a larger RF and reboot the metaserver.
> > 
> > 
> > As I mentioned before, I think `system should not restrict its table max replica count for scalability`, but as empiredan's option, it can `avoid user unreasonable replica count`. If we set `max_allowed_replica_count` as 3, it somehow affects scalability. It is okay for user to create table with 4 replicas and 5 replicas and so on. Besides, 4 and 5 are not user unreasonable replica count. So I prefer seting it as a bigger value. And there are two reasons I recommend it as a const. Firstly, it is easiler to check validation `max_allowed_replica_count` and `min_allowed_replica_count` in validation if one of them is a const value. Secondly, as the `max_allowed_replica_count` is used to avoid unreasonable replica count, we will hardly update it, it is okay to be a const value, because its value is unreasonable enough. That is my opinion above, I think we can disucuss the usage of `max_allowed_replica_count` to determine seting it as an optional value or a const value.
> 
> @hycdong @acelyc111 Or, can we turn to refer to the implementation of the system that has the similar problem with us ?
> 
> In [apache/kudu@b39491d](https://github.com/apache/kudu/commit/b39491de611eb7a36be46816c66fa3a85590a656), Todd has explained why kudu uses `max_num_replicas` to restrict the replication factor. Now `max_num_replicas` is set to 7, since higher RF hasn't been proved to be effective by some test cases. `A well-intentioned user may try to set replication factor to a very high number to get a table replicated onto every server in their cluster`.
> 
> Therefore, should we make `max_allowed_replica_count` configurable and its default value a little higher than 3, say, 5 ?
> 
> And also, RF is restricted by `min_num_repilcas` . See [apache/kudu@6cb1548](https://github.com/apache/kudu/commit/6cb15486cdbeacb722e69b81d563b6f05d840650) for details.
> 
> As for the related code, please see the following quotes:
> 
> ```c++
> DEFINE_int32(max_num_replicas, 7,
>              "Maximum number of replicas that may be specified for a table.");
> // Tag as unsafe since we have done very limited testing of higher than 5 replicas.
> TAG_FLAG(max_num_replicas, unsafe);
> TAG_FLAG(max_num_replicas, runtime);
> 
> DEFINE_int32(min_num_replicas, 1,
>              "Minimum number of replicas that may be specified when creating "
>              "a table: this is to enforce the minimum replication factor for "
>              "tables created in a Kudu cluster. For example, setting this flag "
>              "to 3 enforces every new table to have at least 3 replicas for "
>              "each of its tablets, so there cannot be a data loss when a "
>              "single tablet server fails irrecoverably.");
> TAG_FLAG(min_num_replicas, advanced);
> TAG_FLAG(min_num_replicas, runtime);
> 
> ...
> 
> template<typename RespClass>
> Status CatalogManager::ValidateNumberReplicas(const std::string& normalized_table_name,
>                                               RespClass* resp, ValidateType type,
>                                               const boost::optional<int>& partitions_count,
>                                               int num_replicas) {
>   if (num_replicas > FLAGS_max_num_replicas) {
>     return SetupError(Status::InvalidArgument(
>         Substitute("illegal replication factor $0: maximum allowed replication "
>                    "factor is $1 (controlled by --max_num_replicas)",
>                    num_replicas, FLAGS_max_num_replicas)),
>         resp, MasterErrorPB::REPLICATION_FACTOR_TOO_HIGH);
>   }
>   if (num_replicas < FLAGS_min_num_replicas) {
>     return SetupError(Status::InvalidArgument(
>         Substitute("illegal replication factor $0: minimum allowed replication "
>                    "factor is $1 (controlled by --min_num_replicas)",
>             num_replicas, FLAGS_min_num_replicas)),
>         resp, MasterErrorPB::ILLEGAL_REPLICATION_FACTOR);
>   }
>   // Reject create/alter table with even replication factors, unless master flag
>   // allow_unsafe_replication_factor is on.
>   if (num_replicas % 2 == 0 && !FLAGS_allow_unsafe_replication_factor) {
>     return SetupError(Status::InvalidArgument(
>         Substitute("illegal replication factor $0: replication factor must be odd",
>                    num_replicas)),
>         resp, MasterErrorPB::EVEN_REPLICATION_FACTOR);
>   }
> ...
> }
> ```

Okay, I think if we set max_num_replica as 5, we should add an option for it, user can update it through config. In this case, it is the unreasonable replica count, it restirct the max replica count through a cluster-level configuration. If user would like to create table with higher replica_count, it should update config.

acelyc111:
> config.

+1

acelyc111:
> > > > > > > > > > > Thanks for your pull request for adding replica count check while creating table~ In this pull request, you adds three rules for replica count:
> > > > > > > > > > > 
> > > > > > > > > > > 1. replica count should not exceed alive node count
> > > > > > > > > > > 2. replica count should not below min_replica_count
> > > > > > > > > > > 3. replica count should not exceed max_replica_count
> > > > > > > > > > > 
> > > > > > > > > > > I have a different option with your third rule, I think system should not restrict its table max replica count for scalability. Besides, I think it's unnecessary to make `min_replica_count` as an option, in my view, `min_replica_count` will always be 0. Besides, I add some comments for this pr.
> > > > > > > > > > 
> > > > > > > > > > 
> > > > > > > > > > Yeah, I agree that we should provide scalability and flexibility for the system. The purpose of introducing `max_allowed_replica_count` is to avoid unreasonable replication factor. A big replication factor may lead to poor behavior of the system. We can increase this maximum, but I think we'd better give an upper limit to prevent users from creating an app with 10 replicas.
> > > > > > > > > > Introducing `min_allowed_replica_count` is motivated from the example illustrated in the description above:
> > > > > > > > > > ```
> > > > > > > > > > Another example will be creating an app with a single replica in a cluster that includes 3 nodes. 
> > > > > > > > > > 
> > > > > > > > > > Later then, we decide to make a high-availability upgrade (without stopping service of the cluster) for this cluster. 
> > > > > > > > > > 
> > > > > > > > > > The upgrade process will be blocked in the phase of moving the only replica of that app out to another node, since the node that deployed the only replica is stopped.
> > > > > > > > > > ```
> > > > > > > > > > 
> > > > > > > > > > 
> > > > > > > > > >     
> > > > > > > > > >       
> > > > > > > > > >     
> > > > > > > > > > 
> > > > > > > > > >       
> > > > > > > > > >     
> > > > > > > > > > 
> > > > > > > > > >     
> > > > > > > > > >   
> > > > > > > > > > Considering this example, I think we can set the minimum replication factor as 3 for a cluster that has 3 or more replica servers.
> > > > > > > > > 
> > > > > > > > > 
> > > > > > > > > Okay, your comment persuades me~ `min_replica_count` is used to set the cluster table min replica count, `max_replica_count` is used to avoid user unreasonable replica count. How about make `max_replica_count` as a const value, not a configurable option. For one hand, this value won't be updated in normal cases, for other hands, it will simply the `DSN_DEFINE_validator` check for `min_replica_count`.
> > > > > > > > 
> > > > > > > > 
> > > > > > > > A good idea ! How about set `max_allowed_replica_count` as 3 ? Can this value satisfy all of your scenarios ?
> > > > > > > 
> > > > > > > 
> > > > > > > In most our scenarios, replica_count is 3. As our discussion above, `max_allowed_replica_count` should be greater than the normal replica_count. If we set `max_allowed_replica_count` as 3, then if we would like to provide 5 replica, which is also a reasonable replica count, we should update code. As a result, I suggest seting `max_allowed_replica_count` as 10, or even greater than 10.
> > > > > > 
> > > > > > 
> > > > > > Ok, how about setting `max_allowed_replica_count` as 15 (an odd number, though there's no particular reason) ?
> > > > > 
> > > > > 
> > > > > I think it's okay to set `max_allowed_replica_count` as 15~
> > > > 
> > > > 
> > > > @hycdong @empiredan I do consider a const and unoptional value of 15 is too large, any new user could create a table with 15 replicas at most without any limitation, it's harmful for the system. If most cases are 3 replicas, why not make it optional, in the rare case if any body want to create a table with some larger replication factor, they can set a larger RF and reboot the metaserver.
> > > 
> > > 
> > > As I mentioned before, I think `system should not restrict its table max replica count for scalability`, but as empiredan's option, it can `avoid user unreasonable replica count`. If we set `max_allowed_replica_count` as 3, it somehow affects scalability. It is okay for user to create table with 4 replicas and 5 replicas and so on. Besides, 4 and 5 are not user unreasonable replica count. So I prefer seting it as a bigger value. And there are two reasons I recommend it as a const. Firstly, it is easiler to check validation `max_allowed_replica_count` and `min_allowed_replica_count` in validation if one of them is a const value. Secondly, as the `max_allowed_replica_count` is used to avoid unreasonable replica count, we will hardly update it, it is okay to be a const value, because its value is unreasonable enough. That is my opinion above, I think we can disucuss the usage of `max_allowed_replica_count` to determine seting it as an optional value or a const value.
> > 
> > 
> > @hycdong @acelyc111 Or, can we turn to refer to the implementation of the system that has the similar problem with us ?
> > In [apache/kudu@b39491d](https://github.com/apache/kudu/commit/b39491de611eb7a36be46816c66fa3a85590a656), Todd has explained why kudu uses `max_num_replicas` to restrict the replication factor. Now `max_num_replicas` is set to 7, since higher RF hasn't been proved to be effective by some test cases. `A well-intentioned user may try to set replication factor to a very high number to get a table replicated onto every server in their cluster`.
> > Therefore, should we make `max_allowed_replica_count` configurable and its default value a little higher than 3, say, 5 ?
> > And also, RF is restricted by `min_num_repilcas` . See [apache/kudu@6cb1548](https://github.com/apache/kudu/commit/6cb15486cdbeacb722e69b81d563b6f05d840650) for details.
> > As for the related code, please see the following quotes:
> > ```c++
> > DEFINE_int32(max_num_replicas, 7,
> >              "Maximum number of replicas that may be specified for a table.");
> > // Tag as unsafe since we have done very limited testing of higher than 5 replicas.
> > TAG_FLAG(max_num_replicas, unsafe);
> > TAG_FLAG(max_num_replicas, runtime);
> > 
> > DEFINE_int32(min_num_replicas, 1,
> >              "Minimum number of replicas that may be specified when creating "
> >              "a table: this is to enforce the minimum replication factor for "
> >              "tables created in a Kudu cluster. For example, setting this flag "
> >              "to 3 enforces every new table to have at least 3 replicas for "
> >              "each of its tablets, so there cannot be a data loss when a "
> >              "single tablet server fails irrecoverably.");
> > TAG_FLAG(min_num_replicas, advanced);
> > TAG_FLAG(min_num_replicas, runtime);
> > 
> > ...
> > 
> > template<typename RespClass>
> > Status CatalogManager::ValidateNumberReplicas(const std::string& normalized_table_name,
> >                                               RespClass* resp, ValidateType type,
> >                                               const boost::optional<int>& partitions_count,
> >                                               int num_replicas) {
> >   if (num_replicas > FLAGS_max_num_replicas) {
> >     return SetupError(Status::InvalidArgument(
> >         Substitute("illegal replication factor $0: maximum allowed replication "
> >                    "factor is $1 (controlled by --max_num_replicas)",
> >                    num_replicas, FLAGS_max_num_replicas)),
> >         resp, MasterErrorPB::REPLICATION_FACTOR_TOO_HIGH);
> >   }
> >   if (num_replicas < FLAGS_min_num_replicas) {
> >     return SetupError(Status::InvalidArgument(
> >         Substitute("illegal replication factor $0: minimum allowed replication "
> >                    "factor is $1 (controlled by --min_num_replicas)",
> >             num_replicas, FLAGS_min_num_replicas)),
> >         resp, MasterErrorPB::ILLEGAL_REPLICATION_FACTOR);
> >   }
> >   // Reject create/alter table with even replication factors, unless master flag
> >   // allow_unsafe_replication_factor is on.
> >   if (num_replicas % 2 == 0 && !FLAGS_allow_unsafe_replication_factor) {
> >     return SetupError(Status::InvalidArgument(
> >         Substitute("illegal replication factor $0: replication factor must be odd",
> >                    num_replicas)),
> >         resp, MasterErrorPB::EVEN_REPLICATION_FACTOR);
> >   }
> > ...
> > }
> > ```
> 
> Okay, I think if we set max_num_replica as 5, we should add an option for it, user can update it through config. In this case, it is the unreasonable replica count, it restirct the max replica count through a cluster-level configuration. If user would like to create table with higher replica_count, it should update config.

+1

empiredan:
> > > > > > > > > > > > Thanks for your pull request for adding replica count check while creating table~ In this pull request, you adds three rules for replica count:
> > > > > > > > > > > > 
> > > > > > > > > > > > 1. replica count should not exceed alive node count
> > > > > > > > > > > > 2. replica count should not below min_replica_count
> > > > > > > > > > > > 3. replica count should not exceed max_replica_count
> > > > > > > > > > > > 
> > > > > > > > > > > > I have a different option with your third rule, I think system should not restrict its table max replica count for scalability. Besides, I think it's unnecessary to make `min_replica_count` as an option, in my view, `min_replica_count` will always be 0. Besides, I add some comments for this pr.
> > > > > > > > > > > 
> > > > > > > > > > > 
> > > > > > > > > > > Yeah, I agree that we should provide scalability and flexibility for the system. The purpose of introducing `max_allowed_replica_count` is to avoid unreasonable replication factor. A big replication factor may lead to poor behavior of the system. We can increase this maximum, but I think we'd better give an upper limit to prevent users from creating an app with 10 replicas.
> > > > > > > > > > > Introducing `min_allowed_replica_count` is motivated from the example illustrated in the description above:
> > > > > > > > > > > ```
> > > > > > > > > > > Another example will be creating an app with a single replica in a cluster that includes 3 nodes. 
> > > > > > > > > > > 
> > > > > > > > > > > Later then, we decide to make a high-availability upgrade (without stopping service of the cluster) for this cluster. 
> > > > > > > > > > > 
> > > > > > > > > > > The upgrade process will be blocked in the phase of moving the only replica of that app out to another node, since the node that deployed the only replica is stopped.
> > > > > > > > > > > ```
> > > > > > > > > > > 
> > > > > > > > > > > 
> > > > > > > > > > >     
> > > > > > > > > > >       
> > > > > > > > > > >     
> > > > > > > > > > > 
> > > > > > > > > > >       
> > > > > > > > > > >     
> > > > > > > > > > > 
> > > > > > > > > > >     
> > > > > > > > > > >   
> > > > > > > > > > > Considering this example, I think we can set the minimum replication factor as 3 for a cluster that has 3 or more replica servers.
> > > > > > > > > > 
> > > > > > > > > > 
> > > > > > > > > > Okay, your comment persuades me~ `min_replica_count` is used to set the cluster table min replica count, `max_replica_count` is used to avoid user unreasonable replica count. How about make `max_replica_count` as a const value, not a configurable option. For one hand, this value won't be updated in normal cases, for other hands, it will simply the `DSN_DEFINE_validator` check for `min_replica_count`.
> > > > > > > > > 
> > > > > > > > > 
> > > > > > > > > A good idea ! How about set `max_allowed_replica_count` as 3 ? Can this value satisfy all of your scenarios ?
> > > > > > > > 
> > > > > > > > 
> > > > > > > > In most our scenarios, replica_count is 3. As our discussion above, `max_allowed_replica_count` should be greater than the normal replica_count. If we set `max_allowed_replica_count` as 3, then if we would like to provide 5 replica, which is also a reasonable replica count, we should update code. As a result, I suggest seting `max_allowed_replica_count` as 10, or even greater than 10.
> > > > > > > 
> > > > > > > 
> > > > > > > Ok, how about setting `max_allowed_replica_count` as 15 (an odd number, though there's no particular reason) ?
> > > > > > 
> > > > > > 
> > > > > > I think it's okay to set `max_allowed_replica_count` as 15~
> > > > > 
> > > > > 
> > > > > @hycdong @empiredan I do consider a const and unoptional value of 15 is too large, any new user could create a table with 15 replicas at most without any limitation, it's harmful for the system. If most cases are 3 replicas, why not make it optional, in the rare case if any body want to create a table with some larger replication factor, they can set a larger RF and reboot the metaserver.
> > > > 
> > > > 
> > > > As I mentioned before, I think `system should not restrict its table max replica count for scalability`, but as empiredan's option, it can `avoid user unreasonable replica count`. If we set `max_allowed_replica_count` as 3, it somehow affects scalability. It is okay for user to create table with 4 replicas and 5 replicas and so on. Besides, 4 and 5 are not user unreasonable replica count. So I prefer seting it as a bigger value. And there are two reasons I recommend it as a const. Firstly, it is easiler to check validation `max_allowed_replica_count` and `min_allowed_replica_count` in validation if one of them is a const value. Secondly, as the `max_allowed_replica_count` is used to avoid unreasonable replica count, we will hardly update it, it is okay to be a const value, because its value is unreasonable enough. That is my opinion above, I think we can disucuss the usage of `max_allowed_replica_count` to determine seting it as an optional value or a const value.
> > > 
> > > 
> > > @hycdong @acelyc111 Or, can we turn to refer to the implementation of the system that has the similar problem with us ?
> > > In [apache/kudu@b39491d](https://github.com/apache/kudu/commit/b39491de611eb7a36be46816c66fa3a85590a656), Todd has explained why kudu uses `max_num_replicas` to restrict the replication factor. Now `max_num_replicas` is set to 7, since higher RF hasn't been proved to be effective by some test cases. `A well-intentioned user may try to set replication factor to a very high number to get a table replicated onto every server in their cluster`.
> > > Therefore, should we make `max_allowed_replica_count` configurable and its default value a little higher than 3, say, 5 ?
> > > And also, RF is restricted by `min_num_repilcas` . See [apache/kudu@6cb1548](https://github.com/apache/kudu/commit/6cb15486cdbeacb722e69b81d563b6f05d840650) for details.
> > > As for the related code, please see the following quotes:
> > > ```c++
> > > DEFINE_int32(max_num_replicas, 7,
> > >              "Maximum number of replicas that may be specified for a table.");
> > > // Tag as unsafe since we have done very limited testing of higher than 5 replicas.
> > > TAG_FLAG(max_num_replicas, unsafe);
> > > TAG_FLAG(max_num_replicas, runtime);
> > > 
> > > DEFINE_int32(min_num_replicas, 1,
> > >              "Minimum number of replicas that may be specified when creating "
> > >              "a table: this is to enforce the minimum replication factor for "
> > >              "tables created in a Kudu cluster. For example, setting this flag "
> > >              "to 3 enforces every new table to have at least 3 replicas for "
> > >              "each of its tablets, so there cannot be a data loss when a "
> > >              "single tablet server fails irrecoverably.");
> > > TAG_FLAG(min_num_replicas, advanced);
> > > TAG_FLAG(min_num_replicas, runtime);
> > > 
> > > ...
> > > 
> > > template<typename RespClass>
> > > Status CatalogManager::ValidateNumberReplicas(const std::string& normalized_table_name,
> > >                                               RespClass* resp, ValidateType type,
> > >                                               const boost::optional<int>& partitions_count,
> > >                                               int num_replicas) {
> > >   if (num_replicas > FLAGS_max_num_replicas) {
> > >     return SetupError(Status::InvalidArgument(
> > >         Substitute("illegal replication factor $0: maximum allowed replication "
> > >                    "factor is $1 (controlled by --max_num_replicas)",
> > >                    num_replicas, FLAGS_max_num_replicas)),
> > >         resp, MasterErrorPB::REPLICATION_FACTOR_TOO_HIGH);
> > >   }
> > >   if (num_replicas < FLAGS_min_num_replicas) {
> > >     return SetupError(Status::InvalidArgument(
> > >         Substitute("illegal replication factor $0: minimum allowed replication "
> > >                    "factor is $1 (controlled by --min_num_replicas)",
> > >             num_replicas, FLAGS_min_num_replicas)),
> > >         resp, MasterErrorPB::ILLEGAL_REPLICATION_FACTOR);
> > >   }
> > >   // Reject create/alter table with even replication factors, unless master flag
> > >   // allow_unsafe_replication_factor is on.
> > >   if (num_replicas % 2 == 0 && !FLAGS_allow_unsafe_replication_factor) {
> > >     return SetupError(Status::InvalidArgument(
> > >         Substitute("illegal replication factor $0: replication factor must be odd",
> > >                    num_replicas)),
> > >         resp, MasterErrorPB::EVEN_REPLICATION_FACTOR);
> > >   }
> > > ...
> > > }
> > > ```
> > 
> > 
> > Okay, I think if we set max_num_replica as 5, we should add an option for it, user can update it through config. In this case, it is the unreasonable replica count, it restirct the max replica count through a cluster-level configuration. If user would like to create table with higher replica_count, it should update config.
> 
> +1

Ok, I'll make `max_allowed_replica_count` configurable, with default value set as 5 ~

empiredan:
As has been mentioned by @hycdong , it's difficult to guarantee that `FLAGS_min_allowed_replica_count` 
 should never be greater than `FLAGS_max_allowed_replica_count` in `flag_validator` for the purpose of validation for individual flag. Thus I've committed https://github.com/XiaoMi/rdsn/pull/978 to introduce grouped flag validator to detect the inconsistency between 2 or more flags. Therefore, https://github.com/XiaoMi/rdsn/pull/978 is the prerequisite for this PR.

src/meta/server_state.h
```diff
@@ -292,6 +292,9 @@ class server_state
     void process_one_partition(std::shared_ptr<app_state> &app);
     void transition_staging_state(std::shared_ptr<app_state> &app);
 
+    // check whether a max replica count is valid especially for a new app
+    bool validate_target_max_replica_count(int32_t max_replica_count);
```
acelyc111, line 296: private is OK?

src/meta/test/meta_app_operation_test.cpp
```diff
@@ -24,17 +25,24 @@
 
 namespace dsn {
 namespace replication {
+
+DSN_DECLARE_int32(min_allowed_replica_count);
+
```
acelyc111, line 30: No need to left blank lines.

src/meta/server_state.cpp
```diff
@@ -56,6 +56,39 @@ using namespace dsn;
 namespace dsn {
 namespace replication {
 
+DSN_DEFINE_int32("meta_server",
+                 max_allowed_replica_count,
+                 5,
+                 "max replica count allowed for any app of a cluster");
+
```
acelyc111, line 63: No need to left blank lines for related DSN_TAG_VARIABLE and DSN_DEFINE_validator of a config.

src/meta/server_state.cpp
```diff
@@ -56,6 +56,39 @@ using namespace dsn;
...
+DSN_TAG_VARIABLE(min_allowed_replica_count, FT_MUTABLE);
+
+DSN_DEFINE_validator(min_allowed_replica_count, [](int32_t allowed_replica_count) -> bool {
+    return allowed_replica_count > 0;
+});
+
+DSN_DEFINE_group_validator(min_max_allowed_replica_count, [](std::string &message) -> bool {
+    if (FLAGS_min_allowed_replica_count > FLAGS_max_allowed_replica_count) {
+        message = fmt::format(
+            "FLAGS_min_allowed_replica_count({}) should be <= FLAGS_max_allowed_replica_count({})",
```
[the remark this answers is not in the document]
empiredan: Sure

src/meta/server_state.h
empiredan: This means `validate_target_max_replica_count` should be defined as a private method ? Actually it is now

src/meta/test/meta_app_operation_test.cpp
empiredan: Ok

src/meta/server_state.cpp
empiredan: Ok

src/meta/server_state.h
acelyc111: OK, I missed that.

acelyc111 (approved):
LGTM

levy5307 (approved):

PR #964, opened 2021-11-12
feat(bulk_load): add unhealthy partition check

hycdong:
This pull request adds a function called `check_partition_status`, will be called by `partition_bulk_load` and `partition_ingestion`, checking if primary address is valid and if partition has unhealthy partition. This pull request also adds related unit test.

### DIFF
--- a/src/meta/meta_bulk_load_service.cpp
+++ b/src/meta/meta_bulk_load_service.cpp
@@ -283,41 +283,92 @@ void bulk_load_service::create_partition_bulk_load_dir(const std::string &app_na
 }
 
 // ThreadPool: THREAD_POOL_META_STATE
-void bulk_load_service::partition_bulk_load(const std::string &app_name, const gpid &pid)
+bool bulk_load_service::check_partition_status(
+    const std::string &app_name,
+    const gpid &pid,
+    bool always_unhealthy_check,
+    const std::function<void(const std::string &, const gpid &)> &retry_function,
+    /*out*/ partition_configuration &pconfig)
 {
-    FAIL_POINT_INJECT_F("meta_bulk_load_partition_bulk_load", [](dsn::string_view) {});
-
     std::shared_ptr<app_state> app = get_app(pid.get_app_id());
     if (app == nullptr || app->status != app_status::AS_AVAILABLE) {
         dwarn_f(
             "app(name={}, id={}) is not existed, set bulk load failed", app_name, pid.get_app_id());
         handle_app_unavailable(pid.get_app_id(), app_name);
-        return;
+        return false;
     }
 
-    rpc_address primary_addr = app->partitions[pid.get_partition_index()].primary;
-    if (primary_addr.is_invalid()) {
+    pconfig = app->partitions[pid.get_partition_index()];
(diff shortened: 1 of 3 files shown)


src/meta/meta_bulk_load_service.cpp
```diff
@@ -283,41 +283,92 @@ void bulk_load_service::create_partition_bulk_load_dir(const std::string &app_na
...
-    req->pid = pid;
-    req->app_name = app_name;
-    req->primary_addr = primary_addr;
-    req->remote_provider_name = ainfo.file_provider_type;
-    req->cluster_name = ainfo.cluster_name;
-    req->meta_bulk_load_status = get_partition_bulk_load_status_unlocked(pid);
-    req->ballot = app->partitions[pid.get_partition_index()].ballot;
-    req->query_bulk_load_metadata = is_partition_metadata_not_updated_unlocked(pid);
-    req->remote_root_path = ainfo.remote_root_path;
+    {
```
levy5307, line 359: I think there is no need to add lock for `req`

hycdong: The `_lock` is not for the req, but is used for following:
```
const app_bulk_load_info &ainfo = _app_bulk_load_info[pid.get_app_id()];
req->meta_bulk_load_status = get_partition_bulk_load_status_unlocked(pid);
req->query_bulk_load_metadata = is_partition_metadata_not_updated_unlocked(pid);
```
The lock is added for `_app_bulk_load_info` and two functions, all of them is used to set parameter for `req`. To make code clear, I add the lock as pull request shows.

levy5307 (approved):

src/meta/meta_bulk_load_service.cpp
Smityz: Maybe it's better to wrap a thread-safe map here, it's easier to read the codes. Just keep these codes in this PR

src/meta/meta_bulk_load_service.cpp
```diff
@@ -283,41 +283,92 @@ void bulk_load_service::create_partition_bulk_load_dir(const std::string &app_na
...
 
-    rpc_address primary_addr = app->partitions[pid.get_partition_index()].primary;
-    if (primary_addr.is_invalid()) {
+    pconfig = app->partitions[pid.get_partition_index()];
+    if (pconfig.primary.is_invalid()) {
         dwarn_f("app({}) partition({}) primary is invalid, try it later", app_name, pid);
         tasking::enqueue(LPC_META_STATE_NORMAL,
                          _meta_svc->tracker(),
-                         std::bind(&bulk_load_service::partition_bulk_load, this, app_name, pid),
+                         [this, retry_function, app_name, pid]() { retry_function(app_name, pid); },
```
Smityz, line 306: why capture `this`

hycdong: Well, it is not necessary in this final version. How about delete it in my next pull request?

Smityz: ok

Smityz (approved):

PR #966, opened 2021-11-17
feat(bulk_load): add bulk load manager to meta server

GiantKing:
1. Bulk load is a time-consuming operation, and it will make influence on online service while it proceeding. So we have to add a cluster operation lock, serialized schedul these operations. 
2. Move bulk load manager into meta server to reduce the count of Pegasus components.


src/meta/meta_service.cpp
```diff
@@ -257,6 +257,24 @@ void meta_service::get_node_state(/*out*/ std::map<rpc_address, bool> &all_nodes
 
 void meta_service::balancer_run() { _state->check_all_partitions(); }
 
+bool meta_service::try_lock_meta_op_status(ENUM_META_OP_STATUS op_status)
+{
+    ENUM_META_OP_STATUS expected = ENUM_META_OP_STATUS::FREE;
+    if (!_meta_op_status.compare_exchange_strong(expected, op_status)) {
+        ddebug_f("LOCK meta op status failed, meta server is busy, current op status is {}", (uint8_t)expected);
```
hycdong, line 264: I recommend using error log here, not debug log.

src/meta/meta_service.cpp
```diff
@@ -257,6 +257,24 @@ void meta_service::get_node_state(/*out*/ std::map<rpc_address, bool> &all_nodes
...
+        return false;
+    }
+
+    ddebug_f("LOCK meta op status to {}", (uint8_t)op_status);
```
hycdong, line 268: I suggest print string for op_status, not uint8_t

src/meta/meta_bulk_load_service.cpp
```diff
@@ -104,6 +104,17 @@ void bulk_load_service::on_start_bulk_load(start_bulk_load_rpc rpc)
     // avoid possible load balancing
     _meta_svc->set_function_level(meta_function_level::fl_steady);
 
+    if (!_meta_svc->try_lock_meta_op_status(ENUM_META_OP_STATUS::BULKLOAD)) {
```
hycdong, line 107: I recommend moving `try_lock_meta_op_status` check on the top before L69.

src/meta/meta_bulk_load_service.cpp
```diff
@@ -1408,10 +1466,11 @@ void bulk_load_service::on_query_bulk_load_status(query_bulk_load_rpc rpc)
     }
 
     if (!app->is_bulk_loading) {
-        auto hint_msg = fmt::format("app({}) is not during bulk load", app_name);
+        auto hint_msg = fmt::format("app({}) is not during bulk load, return last time result", app_name);
         derror_f("{}", hint_msg);
-        response.err = ERR_INVALID_STATE;
+        response.err = ERR_OK;
         response.__set_hint_msg(hint_msg);
+        response.__set_last_time_result(app->last_time_result);
```
hycdong, line 1473: query_bulk_load_response doesn't have a filed called `last_time_result`, and this pull request doesn't adds the filed for this struct.

src/meta/meta_bulk_load_service.cpp
```diff
@@ -1460,6 +1519,7 @@ void bulk_load_service::sync_apps_from_remote_storage()
         std::move(path), [this](bool flag, const std::vector<std::string> &children) {
             if (flag && children.size() > 0) {
                 ddebug_f("There are {} apps need to sync bulk load status", children.size());
+                dassert(children.size() == 1, "fatal, more than one app are processing bulkload");
```
hycdong, line 1522: I suggest removing `dassert` check for more than one app is bulk load. In normal cases, it won't happen, but if it happens, meta server can not work totally for this assert. Meta server is important for system, concurrent bulk load process should not affert normal functions.
The same as the following assert in function `try_to_continue_bulk_load`, `do_continue_app_bulk_load`.

src/meta/meta_bulk_load_service.h
```diff
@@ -440,6 +459,8 @@ class bulk_load_service
     std::unordered_map<gpid, bool> _partitions_cleaned_up;
     // Used for bulk load failed and app unavailable to avoid duplicated clean up
     std::unordered_map<app_id, bool> _apps_cleaning_up;
+    // indicate that have the app ever get into ingesting status
+    std::unordered_map<app_id, bool> _apps_ever_ingesting;
```
hycdong, line 463: I think `_apps_ever_ingesting` should be store on zk, not in meta server memory. If an app ever get into ingesting, and meta server crashed, this value will be lost, then it may get wrong bulk_load_result. I suggest adding this filed into `app_bulk_load_info`, which will store onto zk during bulk load process.

src/meta/meta_bulk_load_service.cpp
```diff
@@ -719,6 +745,20 @@ void bulk_load_service::handle_bulk_load_finish(const bulk_load_response &respon
...
+
+            bulk_load_status::type app_status = get_app_bulk_load_status(response.pid.get_app_id());
+            app_state *as = _state->get_app(app_name).get();
+            if (bulk_load_status::BLS_SUCCEED == app_status) {
+                update_app_bulk_load_result(as, bulk_load_result::BLR_SUCCEED);
+            } else if (bulk_load_status::BLS_CANCELED == app_status) {
+                update_app_bulk_load_result(as, bulk_load_result::BLR_CANCELED);
+            } else if (bulk_load_status::BLS_FAILED == app_status) {
+                // last_time_result has already been set before
+                dassert(bulk_load_result::BLR_ERR_FS_INTERNAL == as->last_time_result ||
```
hycdong, line 757: Why would you add assert here?

src/meta/meta_bulk_load_service.cpp
```diff
@@ -797,6 +837,15 @@ void bulk_load_service::try_rollback_to_downloading(const std::string &app_name,
             "app({}) has been rollback to downloading for {} times, make bulk load process failed",
             app_name,
             _apps_rollback_count[pid.get_app_id()]);
+
+        bulk_load_result::type ret = bulk_load_result::BLR_INVALID;
+        if (get_app_ever_ingesting(pid.get_app_id())) {
+            ret = bulk_load_result::BLR_ERR_INGESTED_FAILED;
+        } else {
+            ret = bulk_load_result::BLR_ERR_FS_INTERNAL;    // TODO or BLR_ERR_CORRUPTION?
```
hycdong, line 845: It is not `BLR_ERR_FS_INTERNAL` nor `BLR_ERR_CORRUPTION`, just failed for too many rollback. I think bulk_load_result should add a new filed for it.

hycdong:
1. I notice that this pull request adds a filed called `bulk_load_result` in `app_info` struct indicating the last bulk load result. I think it doesn't provide enough information to users, users only know the result of it, lack of other information such as the start time, end time and the remote file path etc, which are helpful for user find problems if bulk load process failed. We can discuss it together.
2. This pull request contains lots of function update, but has no unit test, please add unit tests.
3. This pull request may update structure on zk, but has no design document nor issue with it, please raise an issue or add more in pull request description.

src/meta/meta_service.cpp
GiantKing: Good idea

src/meta/meta_service.cpp
GiantKing: OK

src/meta/meta_bulk_load_service.cpp
GiantKing: Added in  src/common/bulk_load.thrift

src/meta/meta_bulk_load_service.cpp
GiantKing: Add a new result status: BLR_ERR_TRY_EXHAUST

src/meta/meta_bulk_load_service.cpp
GiantKing: Assert is useful in debug mode，let me aware of some logic mistake. I think we should support NDEBUG in the near future.

src/meta/meta_bulk_load_service.cpp
```diff
@@ -1025,6 +1065,7 @@ void bulk_load_service::update_app_status_on_remote_storage_reply(const app_bulk
                                                                   bulk_load_status::type new_status,
                                                                   bool should_send_request)
 {
+    ddebug_f("DEBUG:update_app_status_on_remote_storage_reply 4");
```
levy5307, line 1068: Does this log useful?

include/dsn/dist/replication/replication_enums.h
```diff
@@ -13,6 +13,7 @@ ENUM_REG(app_status::AS_DROPPED)
 ENUM_REG(app_status::AS_RECALLING)
 ENUM_END2(app_status::type, app_status)
 
+
```
levy5307, line 16: remove this line

src/common/bulk_load.thrift
```diff
@@ -29,7 +29,12 @@ enum bulk_load_status
     BLS_DOWNLOADED,
     BLS_INGESTING,
     BLS_SUCCEED,
-    BLS_FAILED,
+    // BLS_FAILED,
```
hycdong, line 32: Please remove this line.

src/meta/meta_bulk_load_service.cpp
```diff
@@ -101,6 +112,9 @@ void bulk_load_service::on_start_bulk_load(start_bulk_load_rpc rpc)
              request.file_provider_type,
              request.remote_root_path);
 
+    remove_bulk_load_dir_on_remote_storage(app, false);
```
hycdong, line 115: As we decide to not remove app bulk load directory when the last bulk load finished, there are two following solutions:
1. remove potential app when start bulk load, then create directory immediately
2. update original create directory logic, use not-exist then create, exist then update

In current implemetation, you choose solution 1, and I prefer solution 2. Solution 2 can avoid sleep for waiting removing succeed and combine remote storage operation together. You can update function `do_start_app_bulk_load` to implement solution2.

src/meta/meta_bulk_load_service.cpp
```diff
@@ -1408,11 +1462,10 @@ void bulk_load_service::on_query_bulk_load_status(query_bulk_load_rpc rpc)
     }
 
     if (!app->is_bulk_loading) {
-        auto hint_msg = fmt::format("app({}) is not during bulk load", app_name);
+        auto hint_msg = fmt::format("app({}) is not during bulk load, return last time result", app_name);
```
hycdong, line 1465: There are two cases:
1. table finish last round bulk load, it will return last time result
2. table has never executed bulk load before, it don't have last time result, it should return some error, at least should be different from case1

I suggest updating code to consider case 2 .

src/meta/meta_bulk_load_service.cpp
```diff
@@ -1523,6 +1576,11 @@ void bulk_load_service::try_to_continue_bulk_load()
 {
     FAIL_POINT_INJECT_F("meta_try_to_continue_bulk_load", [](dsn::string_view) {});
     for (const auto app_id : _bulk_load_app_id) {
+        // TODO
+        // std::shared_ptr<app_state> app = get_app(app_id);
+        // if (!app || app->status != app_status::AS_AVAILABLE || !app->is_bulk_loading) {
+        //     continue;
+        // }
```
hycdong, line 1583: Please remove this debug log.

src/meta/meta_bulk_load_service.cpp
```diff
@@ -1710,6 +1768,9 @@ void bulk_load_service::do_continue_app_bulk_load(
     const int32_t same_count = pinfo_map.size() - different_count;
     const int32_t invalid_count = partition_count - pinfo_map.size();
 
+    if (!_meta_svc->try_lock_meta_op_status(meta_op_status::type::BULKLOAD)) {
+        derror_f("fatal, the op status of meta server must be meta_op_status::type::FREE");
```
hycdong, line 1772: I think there should add error handling here, if bulk load can not acquire lock, it still moves on without any handling.

src/meta/meta_bulk_load_service.h
```diff
@@ -192,7 +194,7 @@ class bulk_load_service
                                       const gpid &pid,
                                       const rpc_address &primary_addr);
 
-    void reset_local_bulk_load_states(int32_t app_id, const std::string &app_name);
+    void reset_local_bulk_load_states(int32_t app_id, const std::string &app_name, bool is_reset_all);
```
hycdong, line 197: Please adding comments for param `is_reset_all`.

src/runtime/dsn.layer2_types.cpp
```diff
@@ -13,1088 +13,1066 @@
...
-                                   "AS_CREATING",
-                                   "AS_CREATE_FAILED",
-                                   "AS_DROPPING",
-                                   "AS_DROP_FAILED",
-                                   "AS_DROPPED",
-                                   "AS_RECALLING"};
-const std::map<int, const char *> _app_status_VALUES_TO_NAMES(
-    ::apache::thrift::TEnumIterator(8, _kapp_statusValues, _kapp_statusNames),
-    ::apache::thrift::TEnumIterator(-1, NULL, NULL));
+int _kapp_statusValues[] = {
```
Shuo-Jia, line 16: I remember this `Autogenerated` diff has been not upload? @hycdong

hycdong:
I notice that this pull request doesn't format files, and it will complie tests failed because of removing `FAILED` status but doesn't update related unit tests. Expecting your completed pull request~

src/runtime/dsn.layer2_types.cpp
hycdong: Actually, not all `thrift` files are generate `.h` and `.cpp` files during compile process. This file `dsn.layers.thrift` will auto-generate offline and upload `.h` and `.cpp` files.
By the way, this pull request doesn't update it, but it used to, those files are not formated.

PR #968, opened 2021-11-17
feat: add alive nodes perf-counter

empiredan:
Currently we've counted unalive nodes which can be used to warn that some node is dead. Since we've managed lots of clusters, one of the first things that we tend to learn is how many replica servers this cluster has. However, now we cannot read this from a graph. We have to connect to the cluster and check it by commands.

Therefore, I added a counter for alive nodes. Then, taking Prometheus for example,  we can read the number of alive nodes by `meta_eon_meta_service_alive_nodes`, and also the total number of replica servers by `meta_eon_meta_service_alive_nodes + meta_eon_meta_service_unalive_nodes`.

The name and type of the counter is listed as below:
```
meta*eon.meta_service*alive_nodes - NUMBER
```

### DIFF
--- a/src/meta/meta_service.cpp
+++ b/src/meta/meta_service.cpp
@@ -77,6 +77,8 @@ meta_service::meta_service()
         "replica server disconnect count in the recent period");
     _unalive_nodes_count.init_app_counter(
         "eon.meta_service", "unalive_nodes", COUNTER_TYPE_NUMBER, "current count of unalive nodes");
+    _alive_nodes_count.init_app_counter(
+        "eon.meta_service", "alive_nodes", COUNTER_TYPE_NUMBER, "current count of alive nodes");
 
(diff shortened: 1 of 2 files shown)


hycdong (approved):
Please add the counter name and type in your pr description, like https://github.com/XiaoMi/rdsn/pull/891. Besides I recommend that you could add suitable tag for your further pull request for convenience, such as `type/perf-counter` tag indicating this pull request update perf-counters.

empiredan:
> Please add the counter name and type in your pr description, like #891. Besides I recommend that you could add suitable tag for your further pull request for convenience, such as `type/perf-counter` tag indicating this pull request update perf-counters.

Ok, thanks for the suggestion ! I'll add counter name and type in the description.

levy5307 (approved):

Shuo-Jia (approved):

PR #976, opened 2021-12-06
fix: add app_info duplicating initial value

Arsenal821:


Arsenal821:
from：https://github.com/apache/incubator-pegasus/issues/827

Smityz:
It's strange give a optional parameter an initial value, can you fix it by another way?

Shuo-Jia:
> from：[apache/incubator-pegasus#827](https://github.com/apache/incubator-pegasus/issues/827)

The reason is the thrift variable is `optional` so as to it won't be serialized into file.

Arsenal821:
> It's strange to give an optional parameter an initial value, can you fix it in another way?

I don't think ignore this field different when we diff new app_info between old is better. And follow this field, is_bulk_loading also have an initial value. I have no idea for more.

hycdong (approved):

Shuo-Jia (approved):

levy5307 (approved):

PR #978, opened 2021-12-06
feat: implement grouped flag validators

empiredan:
## Motivation
We've implemented `flag_validator` to validate the value of individual flag. However, there are scenarios where inconsistency should be detected and avoided between 2 or more flags.

For example, `FLAGS_a` and `FLAGS_b` are mutually exclusive: they cannot both be true. Therefore, the related validator may be something like:
```c++
bool validate() {
    return !FLAGS_a || !FLAGS_b;
}
```

Another example is that FLAGS_c must be less than FLAGS_d. As for this example, the related validator can be implemented as:
```c++
bool validate() {
    return FLAGS_c < FLAGS_d;
}
```

A recent scenario is from https://github.com/XiaoMi/rdsn/pull/963 where we should guarantee that  `FLAGS_min_allowed_replica_count` will never be greater than `FLAGS_max_allowed_replica_count`.

Unfortunately, `flag_validator` is used to validate an individual flag without involving others. Once another flag is involved in a `flag_validator`, perhaps the validation is ineffective since that flag may not have been loaded from the configuration file.

## Grouped flag validator
Inspired from https://github.com/apache/kudu/commit/e7334c2e6607ac0fa778499eda2df3f4cfcd3fe3, we can use grouped flag validator to detect the inconsistency between 2 or more flags.

In contrast with `flag_validator` for individual flag, `group_flag_validator` has a guarantee that it will be run after all flags have been loaded from the configuration file.

Also, after a flag is updated dynamically with some new value, all grouped validators will be run to find potential inconsistency.

### Implement a validator
The grouped flag validator must be the type of `std::function<bool(std::string &)>`. It does not receive any input argument, but return true if the validation passed otherwise false, with a hint message set as the output argument `std::string &`, if any.

### Register a validator
The macro `DSN_DEFINE_group_validator` is used to register the validator. Similar with `DSN_DEFINE_validator`, a name should be given for each grouped flag validator.

### Example
Take the mutually exclusive flags illustrated above for example, the validator can be implemented and registered as:
```c++
DSN_DEFINE_group_validator(inconsistent_flags, [](std::string &message) -> bool {
    if (FLAGS_a && FLAGS_b) {
        message = fmt::format("a({}) is conflicted with b({})", FLAGS_a, FLAGS_b);
        return false;
    }
    return true;
});
```

For more detailed examples, please see the test cases in `src/utils/test/flag_test.cpp`.

### DIFF
--- a/src/utils/flags.cpp
+++ b/src/utils/flags.cpp
@@ -53,6 +53,8 @@ ENUM_REG(FV_DOUBLE)
 ENUM_REG(FV_STRING)
 ENUM_END(value_type)
 
+using group_validators_runner_fn = std::function<bool(std::string &)>;
+
 class flag_data
 {
 public:
@@ -124,7 +126,7 @@ public:
     }
 
     void set_validator(validator_fn &validator) { _validator = std::move(validator); }
-    void set_group_validators_runner(group_validator_fn &&runner)
+    void set_group_validators_runner(group_validators_runner_fn &&runner)
     {
         _group_validators_runner = std::move(runner);
     }
@@ -193,7 +195,7 @@ private:
     const char *_name;
     const char *_desc;
     validator_fn _validator;
-    group_validator_fn _group_validators_runner;
+    group_validators_runner_fn _group_validators_runner;
     std::unordered_set<flag_tag> _tags;
 };
 
@@ -229,9 +231,12 @@ public:
                        validate_messages.end(),
                        std::back_inserter(messages),
                        [](const std::pair<std::string, std::string> &message) {
-                           return fmt::format("group validator \"{}\" failed: \"{}\"",
-                                              message.first,
-                                              message.second);
+                           std::string base(
+                               fmt::format("group validator \"{}\" failed", message.first));
+                           if (message.second.empty()) {
+                               return base;
+                           }
+                           return fmt::format("{}: \"{}\"", base, message.second);
                        });
 
         total_message = boost::join(messages, "; ");
@@ -332,7 +337,7 @@ private:
 
 private:
     std::map<std::string, flag_data> _flags;
-    std::map<std::string, group_validator_fn> _group_flag_validators;
+    std::map<std::string, group_validators_runner_fn> _group_flag_validators;
 };
 
 #define FLAG_REG_CONSTRUCTOR(type, type_enum)                                                      \
--- a/src/utils/flags.cpp
+++ b/src/utils/flags.cpp
@@ -15,6 +15,8 @@
 // specific language governing permissions and limitations
 // under the License.
 
+#include <boost/algorithm/string/join.hpp>
+
 #include <dsn/utility/flags.h>
 #include <dsn/utility/config_api.h>
 #include <dsn/utility/singleton.h>
@@ -62,6 +64,7 @@ public:
         }                                                                                          \
         break
 
+// we should run all group validators to find potential invalidation
 #define FLAG_DATA_UPDATE_CASE(type, type_enum, suffix)                                             \
     case type_enum: {                                                                              \
         type old_val = value<type>(), tmpval_##type_enum;                                          \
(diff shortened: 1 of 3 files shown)


src/utils/flags.cpp
```diff
@@ -62,6 +66,7 @@ class flag_data
         }                                                                                          \
         break
 
+// we should run all group validators to find potential invalidation
 #define FLAG_DATA_UPDATE_CASE(type, type_enum, suffix)                                             \
```
levy5307, line 70: So we can only group validator in update case?

src/utils/flags.cpp
```diff
@@ -73,6 +78,11 @@ class flag_data
             value<type>() = old_val;                                                               \
             return error_s::make(ERR_INVALID_PARAMETERS, "value validation failed");               \
         }                                                                                          \
+        std::string total_message;                                                                 \
```
levy5307, line 81:
```
        std::string total_message;                                                                 \
        if (_group_validators_runner && !_group_validators_runner(total_message)) {                \
            value<type>() = old_val;                                                               \
            return error_s::make(ERR_INVALID_PARAMETERS, total_message.c_str());                   \
        }                  
```
I think it's better to move group validator to `flag_registry::update_flag`

empiredan: Initially I put the detection of group validators in `flag_registry::update_flag`. However, the old value has to be read and saved in a temporary variable, since later once the detection failed, the old value will be used to set back to  `flag_data` again. We will read the old value by template, which has been implemented well in `FLAG_DATA_UPDATE_CASE`. So can we reuse the template in `FLAG_DATA_UPDATE_CASE` ?

levy5307: OK.

levy5307: But I think you can pass the pointer of flag_registry to flag_data, so you can use flag_registry::run_group_validator here

empiredan: If we want to use the pointer of `flag_registry` and flag_registry::run_group_validators in `flag_data`, we have to declare class `flag_registry` before `flag_data`.  Actually flag_registry also depends on `flag_data`. Thus, the declaration of `flag_data` and `flag_registry` should be moved to the header file. This will lead to a big change to current structure.

src/utils/flags.cpp
```diff
@@ -181,13 +195,62 @@ class flag_data
...
+            failed |= invalid;
+            if (invalid) {
+                validate_messages[validator.first] = message;
+            }
+        }
+
+        return !failed;
+    }
+
+    bool run_group_validators(std::string &total_message)
```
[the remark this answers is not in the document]
empiredan: Does this mean boost::optional<std::string> run_group_validators() ? In `auto res = run_group_validators();`, `!res` will means that validation succeeds, is this counterintuitive ?

src/utils/flags.cpp
levy5307: I think it's OK to change the current structure. What we should do is split `flag_data` and `flag_registry` into seperate files.
The reason why I want to do this is that I think group flag validators belongs to `flag_registry`. It's better to not 
 to affect `flag_data`

empiredan: Yeah I agree that it's better not to affect `flag_data`. All flags and grouped validators are saved in `flag_registry`. So at first I put everything about grouped validators in `flag_registry`. `flag_data` knows nothing about grouped validators. Even it knows nothing about `flag_registry`.

However,  the value that grouped validators detect belong to each `flag_data`, just as the individual validator. In this respect, there must be some "relation" between `flag_data` and grouped validators. I try to think other alternative patterns, but only `flag_data` knows how to parse to the templated value. Also the error message is defined in the returned `error_s`. So I think running grouped validators should be done by `update` function.

I think an extra advantage of taking `flag_registry` pointer in `flag_data` is that if later there's another "relation" found, `flag_registry` pointer can be reused.

empiredan: Suggested by @levy5307 , we will use `join_point` to trigger group validations while `flag_data` is updating the value.

levy5307 (dismissed):
LGTM

levy5307 (approved):

Smityz (approved):

PR #979, opened 2021-12-07
feat: zookeeper client support kerberos

Geha-fighting:
from: https://github.com/apache/incubator-pegasus/issues/847


scripts/linux/start_zk.sh
```diff
@@ -27,29 +27,34 @@ fi
 
 cd $INSTALL_DIR
 
-ZOOKEEPER_PKG=${PROJECT_DIR}/thirdparty/build/Download/zookeeper/zookeeper-3.4.10.tar.gz
-if [ ! -f ${ZOOKEEPER_PKG} ]; then
-    echo "no such file \"${ZOOKEEPER_PKG}\""
-    echo "please install third-parties first"
-    exit 1
+ZOOKEEPER_ROOT=apache-zookeeper-3.7.0-bin
+ZOOKEEPER_TAR_NAME=${ZOOKEEPER_ROOT}.tar.gz
```
Shuo-Jia, line 31: suggest keep consistent with the original package management: use cmake https://github.com/XiaoMi/rdsn/blob/v2.3.0/thirdparty/CMakeLists.txt#L155

Smityz:
Is it forward compatible?

Smityz:
you need use `./scripts/linux/run-clang-format.sh` to format your codes, so that you can pass the CI

thirdparty/CMakeLists.txt
```diff
@@ -153,14 +153,15 @@ if (COMPILER_SUPPORTS_FORMAT_OVERFLOW)
...
-        URL_MD5 e4cf1b1593ca870bf1c7a75188f09678
-        CONFIGURE_COMMAND cd src/c && CFLAGS=${ZOOKEEPER_CFLAGS} ./configure --enable-static=yes --enable-shared=no --prefix=${TP_OUTPUT} --with-pic=yes
-        BUILD_COMMAND cd src/c && make
-        INSTALL_COMMAND cd src/c && make install
+        URL https://dlcdn.apache.org/zookeeper/zookeeper-3.7.0/apache-zookeeper-3.7.0.tar.gz
+        https://dlcdn.apache.org/zookeeper/zookeeper-3.7.0/apache-zookeeper-3.7.0.tar.gz
+        URL_MD5 44c2a33e01931aed94ef7f3d39d0963e
+        CONFIGURE_COMMAND cd zookeeper-jute && mvn compile && cd ../zookeeper-client/zookeeper-client-c && autoreconf -if && CFLAGS=${ZOOKEEPER_CFLAGS} ./configure --disable-shared --prefix=${TP_OUTPUT} --enable-debug --with-pic=yes --with-sasl=${TP_OUTPUT}
+        BUILD_COMMAND cd zookeeper-client/zookeeper-client-c && make
+        INSTALL_COMMAND cd zookeeper-client/zookeeper-client-c && make install
```
Shuo-Jia, line 161: via cmake to compile zk, you should use it directly, but I find you still download public binary package to use: https://github.com/XiaoMi/rdsn/pull/979/files#diff-9035c089f6e615a77e99d6c4f8b2ea40df5933d3b5f6fa97ae22cc157e4968b6R35

scripts/linux/start_zk.sh
Arsenal821: zookeeper-3.4.10 just have one package for us. (https://zookeeper.apache.org/releases.html) But on zookeeper-3.7.0 it provided two packages, one just has bin what we can use it to start zookeeper cluster, another one both bin and source code.  I think we just need bin when we need start zookeeper cluster.

Arsenal821:
> compatible if you means zookeeper server 3.4.10，I did it

scripts/linux/start_zk.sh
Smityz: Using Cmake is good for package management, and also we can use OSS to speed up download

src/runtime/security/init.cpp
```diff
@@ -68,5 +68,22 @@ bool init(bool is_server)
...
 }
 
+bool init_for_zookeeper_client()
+{
+    error_s err = run_kinit();
+    if (!err.is_ok()) {
+        derror_f("initialize kerberos failed, with err = {}", err.description());
+        return false;
+    }
+    ddebug("initialize kerberos for zookeeper client succeed");
```
Smityz, line 78:
```suggestion
    dinfo("initialize kerberos for zookeeper client succeed");
```

src/runtime/security/init.cpp
```diff
@@ -68,5 +68,22 @@ bool init(bool is_server)
...
+
+    err = init_sasl(false);
+    if (!err.is_ok()) {
+        derror_f("initialize sasl failed, with err = {}", err.description());
+        return false;
+    }
+    ddebug("initialize sasl for zookeeper client succeed");
```
Smityz, line 85:
```suggestion
    dinfo("initialize sasl for zookeeper client succeed");
```

src/zookeeper/zookeeper_session.cpp
```diff
@@ -141,12 +151,26 @@ int zookeeper_session::attach(void *callback_owner, const state_callback &cb)
 {
     utils::auto_write_lock l(_watcher_lock);
     if (nullptr == _handle) {
-        _handle = zookeeper_init(zookeeper_session_mgr::instance().zoo_hosts(),
-                                 global_watcher,
-                                 zookeeper_session_mgr::instance().timeout(),
-                                 nullptr,
-                                 this,
-                                 0);
+        if (dsn::security::FLAGS_enable_zookeeper_kerberos) {
```
Smityz, line 154: what will happen when `FLAGS_enable_zookeeper_kerberos=true` but ZK does not support this function?

src/runtime/security/init.cpp
levy5307: I think it's ok to use ddebug here

Arsenal821:
How to change ut docker file and command? I need add mvn and netcat, and rebuild thirdparty libs.

src/runtime/security/init.cpp
acelyc111: I think we can make a plan to reverse info-level and debug-level, it's confused for most developers :)

src/runtime/security/init.cpp
```diff
@@ -68,5 +68,22 @@ bool init(bool is_server)
...
+bool init_for_zookeeper_client()
+{
+    error_s err = run_kinit();
+    if (!err.is_ok()) {
+        derror_f("initialize kerberos failed, with err = {}", err.description());
+        return false;
+    }
+    ddebug("initialize kerberos for zookeeper client succeed");
+
+    err = init_sasl(false);
```
acelyc111, line 80: Better to make a simple refactor on the boolean parameter, boolean is meanless. For example add an enum like:
```
enum class KerberosRole {
    kClient = 0,
    kServer = 1
};
```

src/runtime/security/negotiation.cpp
```diff
@@ -32,6 +32,10 @@ const std::set<std::string> supported_mechanisms{"GSSAPI"};
 
 DSN_DEFINE_bool("security", enable_auth, false, "whether open auth or not");
 DSN_DEFINE_bool("security", mandatory_auth, false, "wheter to do authertication mandatorily");
+DSN_DEFINE_bool("security",
```
acelyc111, line 35: Better to keep these configs in lexicographical order, it would be friendly for reading.

src/zookeeper/zookeeper_session.cpp
acelyc111: Emm, after this patch, zk client lib has been upgraded to a higher version which support this function already.

scripts/linux/start_zk.sh
```diff
@@ -27,29 +27,34 @@ fi
...
+
+if [ ! -f $ZOOKEEPER_TAR_NAME ]; then
+    echo "Downloading zookeeper..."
+    download_url="https://downloads.apache.org/zookeeper/zookeeper-3.7.0/apache-zookeeper-3.7.0-bin.tar.gz"
+    if ! wget -T 5 -t 1 $download_url; then
+        echo "ERROR: download zookeeper failed"
```
acelyc111, line 37: Better to check the MD5 after downloaded.

thirdparty/CMakeLists.txt
```diff
@@ -153,14 +153,15 @@ if (COMPILER_SUPPORTS_FORMAT_OVERFLOW)
     set(ZOOKEEPER_CFLAGS -Wno-error=format-overflow)
 endif ()
 ExternalProject_Add(zookeeper
-        URL ${OSS_URL_PREFIX}/zookeeper-3.4.10.tar.gz
-        http://ftp.jaist.ac.jp/pub/apache/zookeeper/zookeeper-3.4.10/zookeeper-3.4.10.tar.gz
-        URL_MD5 e4cf1b1593ca870bf1c7a75188f09678
-        CONFIGURE_COMMAND cd src/c && CFLAGS=${ZOOKEEPER_CFLAGS} ./configure --enable-static=yes --enable-shared=no --prefix=${TP_OUTPUT} --with-pic=yes
-        BUILD_COMMAND cd src/c && make
-        INSTALL_COMMAND cd src/c && make install
+        URL https://dlcdn.apache.org/zookeeper/zookeeper-3.7.0/apache-zookeeper-3.7.0.tar.gz
```
acelyc111, line 156: How about download from https://github.com/apache/zookeeper/archive/refs/tags/release-3.7.0.tar.gz ? It would be much faster than download from apache.org

src/runtime/security/init.cpp
levy5307: +1

thirdparty/CMakeLists.txt
```diff
@@ -108,6 +108,7 @@ ExternalProject_Add(gperftools
 set(HDFS_CLIENT_DIR "hadoop-hdfs-project/hadoop-hdfs-native-client")
 ExternalProject_Add(hadoop
         URL https://github.com/apache/hadoop/archive/refs/tags/rel/release-2.8.4.tar.gz
+        https://pegasus-resource-package.oss-cn-beijing.aliyuncs.com/hadoop-release-2.8.4.tar.gz
```
acelyc111, line 111: > List of paths and/or URL(s) of the external project's source. When more than one URL is given, they are tried in turn until one succeeds.

So, better to keep URLs in order, aliyun OSS URL at first, then original URL, most developers are in China mainland after all.

acelyc111: I have talked with @neverchanje , you can add another aliyun oss url and define a new OSS_URL_PREFIX.

thirdparty/CMakeLists.txt
```diff
@@ -49,6 +49,7 @@ set(TP_OUTPUT ${PROJECT_SOURCE_DIR}/output)
 set_property(DIRECTORY PROPERTY EP_BASE ${TP_DIR}/build)
 
 set(OSS_URL_PREFIX "http://pegasus-thirdparties.oss-cn-beijing.aliyuncs.com")
+set(FRESH_OSS_URL_PREFIX "http://pegasus-thirdparty-package.oss-cn-beijing.aliyuncs.com")
```
acelyc111, line 52: Could you please upload all thirdparties to the new bucket? And then you can replace OSS_URL_PREFIX directly.

acelyc111 (approved):

foreverneverer (approved):

PR #981, opened 2021-12-09
feat(manual_compaction): replica report status part1. add function query_compact_status

hycdong:
As https://github.com/apache/incubator-pegasus/issues/850 shows, meta server should know the status of manual compaction.
This pull request is the part1 of the replica servers report manual compaction status to meta server:
1. move structure manual_compaction_status from `replica.h` to `metadata.thrift`
2. add a new function `query_compact_status` in `replication_app_base.h`, which will be implemented by pegasus repo to return the actual compaction status.

### DIFF
--- a/src/replica/replica.cpp
+++ b/src/replica/replica.cpp
@@ -550,7 +550,7 @@ void replica::init_disk_tag()
 {
     dsn::error_code err = _stub->_fs_manager.get_disk_tag(dir(), _disk_tag);
     if (dsn::ERR_OK != err) {
-        derror_replica("get disk tag of %s failed: %s, init it to empty ", dir(), err);
+        derror_replica("get disk tag of {} failed: {}, init it to empty ", dir(), err);
     }
 }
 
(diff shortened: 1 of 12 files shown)


Shuo-Jia (approved):

src/replica/replica.cpp
```diff
@@ -503,44 +503,10 @@ std::string replica::query_manual_compact_state() const
...
-    if (compact_state.find("recent start at") != std::string::npos) {
-        return kRunning;
-    } else if (compact_state.find("recent enqueue at") != std::string::npos) {
-        return kQueuing;
-    } else if (compact_state.find("last used") != std::string::npos) {
-        return kFinished;
-    } else {
-        return kIdle;
-    }
+    dassert_replica(_app != nullptr, "");
```
levy5307, line 508: dcheck_eq(_app, nullptr)

hycdong: `dcheck_eq` won't print `replica_name()`, I suppose using `dcheck_eq_replica` is better. 
If you agree with me, I will raise another pull request to update other three problems including it.

levy5307: OK.

levy5307 (approved):

PR #987, opened 2021-12-13
feat(manual_compaction): meta server support querying compaction status

hycdong:
As apache/incubator-pegasus#850 shows, meta server should know the status of manual compaction.
This pull request adds a new meta server interface to query manual compaction status, including following:
1. Add ddl and meta server interface `query_app_manual_compact`
2. Add helper function to reset compaction status and calculate compaction progress, in `meta_data.cpp`
3. Add unit test for function query_compaction_status


src/meta/meta_data.cpp
```diff
@@ -479,6 +479,37 @@ void app_state_helper::on_init_partitions()
     restore_states.resize(owner->partition_count);
 }
 
+void app_state_helper::reset_manual_compact_status()
+{
+    for (auto &cc : contexts) {
```
levy5307, line 484: for (const auto &cc : contexts) {

hycdong: This for sentence will update value of `contexts`, it is not a read-only iterator, so it can't be const.

src/meta/meta_data.cpp
```diff
@@ -479,6 +479,37 @@ void app_state_helper::on_init_partitions()
...
+                finish_count++;
+            }
+        }
+    }
+    // all replica of all partitions are idle
+    if (idle_count == total_replica_count) {
+        progress = 0;
+        return false;
+    }
+    progress = finish_count * 100 / total_replica_count;
```
levy5307, line 509: Does total_replica_count always greater than 0?

src/meta/meta_data.h
```diff
@@ -324,6 +324,11 @@ class app_state_helper
             cc.lb_actions.clear();
         }
     }
+
+    void reset_manual_compact_status();
+    // get replica group manual compact progress
+    // return false if partition is not executing manual compaction
+    bool get_manual_compact_progress(/*out*/ int32_t &progress) const;
```
levy5307, line 331: What about implement like this: 
```int32_t get_manual_compact_progress() const;```
The return value is progress. 0 means is not executing manual compaction

File: src/meta/meta_data.cpp [hunk 4]
```diff
@@ -479,6 +479,37 @@ void app_state_helper::on_init_partitions()
...
+    for (const auto &cc : contexts) {
+        for (const auto &r : cc.serving) {
+            if (r.compact_status == manual_compaction_status::IDLE) {
+                idle_count++;
+            } else if (r.compact_status == manual_compaction_status::FINISHED) {
+                finish_count++;
+            }
+        }
+    }
+    // all replica of all partitions are idle
```
levy5307, line 504: all replicas

src/meta/meta_data.cpp
hycdong: total_replica_count = partition_count * replica_count, both of them should always be greater than 0.

src/meta/meta_data.h
hycdong: The return value is to distinguish not executing manual compaction and compaction is running or queuing but progress is 0.

src/meta/meta_data.cpp
levy5307: So in some error case, total_replica_count may be 0. So I think it's better to add a judgement here.

src/meta/meta_data.h
levy5307: But I can't get these two cases in your implementation of `get_manual_compact_progress`

hycdong: I got your point. There are other manual compaction status: queuing and running, both of them is not finished status.

src/meta/meta_data.cpp
hycdong: I think it is almost a impossible error case, should coredump because of `total_replica_count=0`. I think it's okay to maintain the code or add an assert sentence for it. How about your suggestion?

levy5307: assert is better, in my opinion :)

hycdong: Done

src/meta/server_state.h
```diff
@@ -171,6 +171,9 @@ class server_state
 
     void on_query_restore_status(configuration_query_restore_rpc rpc);
 
+    // manual compaction
```
levy5307, line 174: I think you can add some meaningful notes here

hycdong: The comment here is just like division, showing the following functions are about manual compaction, because `server_state.h` defined lots of functions. You can reference `server_state.h` to see similar usage, such as `app options`, `dump & restore`.

levy5307 (dismissed):

Shuo-Jia (approved):

levy5307 (approved):

PR #991, opened 2021-12-17
feat(online_migration)：part1 - add new app_envs to control rocksdb allow_ingest_behind

hycdong:
As https://github.com/apache/incubator-pegasus/issues/851 show, we decide to support online migration. 
This pull request adds a new app_envs to control rocksdb allow_ingest_behind option, inluding:
1. Add a new app_envs `rocksdb.allow_ingest_behind`
2. This app_envs will also parse to replica server through on_config_sync, and update in `update_allow_ingest_behind`
3. The value of app_envs `rocksdb.allow_ingest_behind` will be store in `.app_info`, see in function `replica_app_info::store`


src/replica/replication_app_base.cpp
```diff
@@ -266,12 +267,18 @@ error_code replica_app_info::store(const char *file)
...
     marshall(writer, magic, DSF_THRIFT_BINARY);
-    // do not persistent envs to app info file
     if (_app->envs.empty()) {
         marshall(writer, *_app, DSF_THRIFT_JSON);
     } else {
+        // for most envs, do not persistent them to app info file
+        // ROCKSDB_ALLOW_INGEST_BEHIND should be persistent
         dsn::app_info tmp = *_app;
         tmp.envs.clear();
+        for (auto &kv : _app->envs) {
```
levy5307, line 277: const

Smityz (dismissed):

Smityz (dismissed):

src/replica/replication_app_base.cpp
```diff
@@ -266,12 +267,18 @@ error_code replica_app_info::store(const char *file)
...
     marshall(writer, magic, DSF_THRIFT_BINARY);
-    // do not persistent envs to app info file
     if (_app->envs.empty()) {
         marshall(writer, *_app, DSF_THRIFT_JSON);
     } else {
+        // for most envs, do not persistent them to app info file
+        // ROCKSDB_ALLOW_INGEST_BEHIND should be persistent
         dsn::app_info tmp = *_app;
         tmp.envs.clear();
+        for (const auto &kv : _app->envs) {
```
Shuo-Jia, line 277:
```suggestion
        if (_app->envs[ROCKSDB_ALLOW_INGEST_BEHIND].count(!= 0))  // or iterater.end  
        {
           tmp.env[ROCKSDB_ALLOW_INGEST_BEHIND] = _app->envs[ROCKSDB_ALLOW_INGEST_BEHIND];
        }
```


Need to loop

src/replica/replica.h
```diff
@@ -440,6 +440,12 @@ class replica : public serverlet<replica>, public ref_counter, public replica_ba
     void update_bool_envs(const std::map<std::string, std::string> &envs,
                           const std::string &name,
                           /*out*/ bool &value);
+    // get bool envs[name], return false if value is not bool
+    bool get_bool_envs(const std::map<std::string, std::string> &envs,
```
levy5307, line 444: these two functions have nothing to do with class replica. I think you needn't make them to be member functions for class replica

src/replica/replica_config.cpp
```diff
@@ -594,6 +607,31 @@ void replica::update_ac_allowed_users(const std::map<std::string, std::string> &
...
 
+void replica::update_allow_ingest_behind(const std::map<std::string, std::string> &envs)
+{
+    bool new_value = false;
+    if (!get_bool_envs(envs, replica_envs::ROCKSDB_ALLOW_INGEST_BEHIND, new_value)) {
+        return;
+    }
+    if (new_value != _allow_ingest_behind) {
+        // update it into file
+        auto info = _app_info;
```
levy5307, line 618: implement a function named like update_envs_in_info_file, and put L618->L622 into this func

hycdong: There are other code use similar code, I will add it in another pr, I add comment here.

src/replica/replica.h
hycdong: I seperate `get_bool_envs` out of replica class, but remain `update_bool_envs`. It is actually update the filed of replica local variety, I think it's better to remain it in replica class.

levy5307: Which one does `update_bool_envs` update? Sorry I didn't find it.

hycdong: https://github.com/XiaoMi/rdsn/blob/982c2bd180a081fe2c772191cad0ea279e0e2685/src/replica/replica_config.cpp#L557-L566

`_deny_client_write` and `_validate_partition_hash`, although they are passed into `update_bool_envs`, but actually update the value of replica class.

levy5307: I don't agree with you. These two local values are passed by parameters. No need to make it a member function

hycdong: I still insist on my opinion now. In my view, function `update_bool_envs` is only a help function to simply the code to update the bool envs of replica, it just like function `update_throttle_env_internal`, it is not a utility function doesn't have any relationship with the replica class.

levy5307: I know what you mean. It looks strange to me that when looks into the implementation of this function, it updates nothing related to replica.

levy5307: And envs belongs to app_info, not replica

hycdong: > function update_bool_envs is only a help function to simply the code to update the bool envs of replica

The meaning of the sentence above is that it is the help function to simply the code to update `_deny_client_write` and `_validate_partition_hash` of replica class, `the bool envs of replica` is the simple statement.

> I know what you mean. It looks strange to me that when looks into the implementation of this function, it updates nothing related to replica.

If it move out of replica class, it will be like:
```
void update_bool_envs(const std::string &replica_name,
                                      const std::map<std::string, std::string> &envs,
                                      const std::string &name,
                                       bool &value)
{
    bool new_value = false;
    if (!get_bool_envs(envs, name, new_value)) {
        dwarn_f("{}: invalid value of env {}", replica_name, name);
        return;
    }
    if (new_value != value) {
        ddebug_f("{}, switch env[{}] from {} to {}", replica_name, name, value, new_value);
        value = new_value;
    }
}
```
I got your point, but it is also strange to move this function out of class replica. As the comment I said, it just like function `update_throttle_env_internal`, you can reference this function.

levy5307 (approved):

Smityz (approved):

PR #992, opened 2021-12-17
refactor: move func get_current_cluster_name out of duplication_common.h

levy5307:
`get_current_cluster_name` has nothing to do with duplication. So I move this function into common.h which is new created


include/dsn/dist/replication/common.h
```diff
@@ -0,0 +1,30 @@
+/*
```
hycdong, line 1: Why would you add a new file called `common.h`, not using `replication_common.h`, what's the difference between them?

Shuo-Jia: +1, add into https://github.com/XiaoMi/rdsn/blob/v2.3.0/src/common/replication_common.cpp is ok

levy5307: replication_common.h has a lot of classes and functions that used for replication. 
I don't want to put  funcget_current_cluster_name into it.

hycdong: `replication_common.h` is not only used for replica, but also used for meta, although its name is `replication_common`.
It has already been included in lots of meta header files such as `meta_service.h` and `server_state.h`.  I think `replication_common.h` and `common.h` are same.

levy5307: I said replication, not replica

levy5307: In my opinion, replication_common.h is used for replication, using by meta and replica.
common.h may be used in everywhere

hycdong: It's my fault that my pervious comment didn't show my opinion clearly, actually `replication_common.h` can also be used everywhere, such as `ddl_client`. And `replication_common.h` is actually the common, but its name is not suitable. I didn't say you shouldn't add a new file, I just don't figure out the difference.
If you insist on adding a new `common.h`, I think you should explain the difference between it and the original common file, or you should add your future refacion plan about how to solve the two files who have smiliar names.

levy5307: I think replication_common.h can't be used in everywhere. Just as it's name, it's the common file for replication. I think we used it in a wrong way.

hycdong: Okay, its name may lead to misunderstanding. How about adding some TODO comments in `replication_common.h` to explain that we plan to refactor it?

Shuo-Jia: > In my opinion, replication_common.h is used for replication, using by meta and replica.
> common.h may be used in everywhere

You has put the header into `dist/replication` that it do mean `replication_common`, so I think this file is same as `replication_common`.

if you want add `general file` to support `all common` function, that is `tools` , or you shouldn't put current module.

However, define `get_cluster_name` under `replication` module seem to be suitable.

GiantKing: If we add a new file, shell we refactor the code and correct the mistake we made bebore? Or we can change "replication_common.h" to "common.h" directly?

levy5307: I objected to your suggestion. You get misunderstanding about  `tools`. It's not a tool in rdsn.
And I'd rather put it out of directory `dist/replication`.

levy5307: > Okay, its name may lead to misunderstanding, and we should update its wrong usage. How about adding some TODO comments in `replication_common.h` to explain that we plan to refactor it?

I agree with you

levy5307: > If we add a new file, shell we refactor the code and correct the mistake we made bebore? Or we can change "replication_common.h" to "common.h" directly?

I have thought about your idea. But there are a lot of inner types in rdsn, which won't expose outside. If we do it like this, we should move `replication_common.h` under directory `include`, which would expose to other code.

hycdong (approved):

Shuo-Jia (approved):

PR #1001, opened 2021-12-28
fix(rocksdb): upgrade rocksdb to fix NotifyOnFlushCompleted() for atomic flush

empiredan:
Previously in https://github.com/apache/incubator-pegasus/issues/745 we'd reported a bug that the values of all metrics of rocksdb flush are always 0.

Since this bug has been fixed in https://github.com/XiaoMi/pegasus-rocksdb/pull/34 by cherry-pick, we could upgrade the rocksdb to the latest fixed version.


neverchanje (approved):
LGTM

levy5307 (approved):

PR #1002, opened 2021-12-30
feat(online_migration): part2 - add ingest_behind option for bulk load

hycdong:
As apache/incubator-pegasus#851 show, we decide to support online migration.

This pull request add `ingest_behind` option for bulk load, including:
- Add `ingest_behind` in `start_bulk_load_request` and `app_bulk_load_info`
- Add parameter check for `ingest_behind`, reference function `validate_ingest_behind`
- Add related unit test

### DIFF
--- a/src/meta/meta_bulk_load_service.cpp
+++ b/src/meta/meta_bulk_load_service.cpp
@@ -16,6 +16,7 @@
 // under the License.
 
 #include <dsn/dist/fmt_logging.h>
+#include <dsn/dist/replication/replica_envs.h>
 #include <dsn/utility/fail_point.h>
 
 #include "meta_bulk_load_service.h"
@@ -91,13 +92,8 @@ void bulk_load_service::on_start_bulk_load(start_bulk_load_rpc rpc)
     }
 
     std::string hint_msg;
-    error_code e = check_bulk_load_request_params(request.app_name,
-                                                  request.cluster_name,
-                                                  request.file_provider_type,
-                                                  request.remote_root_path,
-                                                  app->app_id,
-                                                  app->partition_count,
-                                                  hint_msg);
+    error_code e = check_bulk_load_request_params(
+        request, app->app_id, app->partition_count, app->envs, hint_msg);
     if (e != ERR_OK) {
         response.err = e;
         response.hint_msg = hint_msg;
@@ -105,11 +101,13 @@ void bulk_load_service::on_start_bulk_load(start_bulk_load_rpc rpc)
         return;
     }
 
-    ddebug_f("app({}) start bulk load, cluster_name = {}, provider = {}, remote root path = {}",
+    ddebug_f("app({}) start bulk load, cluster_name = {}, provider = {}, remote root path = {}, "
+             "ingest_behind = {}",
              request.app_name,
              request.cluster_name,
              request.file_provider_type,
-             request.remote_root_path);
+             request.remote_root_path,
+             request.ingest_behind);
 
     // clear old bulk load result
     reset_local_bulk_load_states(app->app_id, app->app_name, true);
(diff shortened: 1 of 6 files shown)


src/meta/meta_bulk_load_service.cpp
```diff
@@ -1651,9 +1657,28 @@ void bulk_load_service::try_to_continue_app_bulk_load(
                                different_status_pidx_set));
 }
 
+// ThreadPool: THREAD_POOL_META_SERVER
+/*static*/ bool
+bulk_load_service::validate_ingest_behind(const std::map<std::string, std::string> &envs, bool val)
+{
+    bool app_allow_ingest_behind = false;
+    auto iter = envs.find(replica_envs::ROCKSDB_ALLOW_INGEST_BEHIND);
```
levy5307, line 1665: const auto &iter

src/meta/meta_bulk_load_service.cpp
```diff
@@ -1651,9 +1657,28 @@ void bulk_load_service::try_to_continue_app_bulk_load(
...
 }
 
+// ThreadPool: THREAD_POOL_META_SERVER
+/*static*/ bool
+bulk_load_service::validate_ingest_behind(const std::map<std::string, std::string> &envs, bool val)
+{
+    bool app_allow_ingest_behind = false;
+    const auto &iter = envs.find(replica_envs::ROCKSDB_ALLOW_INGEST_BEHIND);
+    if (iter != envs.end()) {
+        if (!buf2bool(iter->second, app_allow_ingest_behind)) {
```
levy5307, line 1667: Could you reuse func `check_bool_value` here?

src/meta/meta_bulk_load_service.h
```diff
@@ -134,12 +136,10 @@ class bulk_load_service
...
     // - ERR_CORRUPTION: bulk_load_info is damaged on file_provider
     // - ERR_INCONSISTENT_STATE: app_id or partition_count inconsistent
-    error_code check_bulk_load_request_params(const std::string &app_name,
-                                              const std::string &cluster_name,
-                                              const std::string &file_provider,
-                                              const std::string &remote_root_path,
+    error_code check_bulk_load_request_params(const start_bulk_load_request &request,
                                               const int32_t app_id,
                                               const int32_t partition_count,
+                                              const std::map<std::string, std::string> envs,
```
levy5307, line 142: value copy here

src/meta/meta_bulk_load_service.cpp
hycdong: It's good idea, but I tried and found out that couldn't reuse `check_bool_value` function.
https://github.com/XiaoMi/rdsn/blob/08af400e3b5a0b43c61baa1cb9e4c4dc8f3bbb64/src/meta/app_env_validator.cpp#L118-L126
`check_bool_value` function only check if the result is bool or not, can not get the result value.

levy5307 (approved):

Smityz (dismissed):

src/meta/meta_bulk_load_service.cpp
```diff
@@ -1651,9 +1657,28 @@ void bulk_load_service::try_to_continue_app_bulk_load(
...
+            dwarn_f("can not convert {} to bool", iter->second);
+            app_allow_ingest_behind = false;
+        }
+    }
+    if (val && !app_allow_ingest_behind) {
```
levy5307, line 1672: `return app_allow_ingest_behind && val;`
And I think it's better to rename value to ingest_behind

hycdong: The return value could not simplify as `app_allow_ingest_behind && val`, the return value is like:
val && app_allow_ingest_behind -> true
!val && !app_allow_ingest_behind -> true
!val && app_allow_ingest_behind -> true
val && !app_allow_ingest_behind -> false
By the way, rename val is a good idea, I have updated it.

levy5307 (approved):

Smityz (approved):

PR #1003, opened 2021-12-30
refactor: add a new function to store app_info into file

hycdong:


src/replica/replica.cpp
```diff
@@ -556,5 +557,16 @@ void replica::init_disk_tag()
     }
 }
 
+error_code replica::store_app_info(app_info info, const std::string &path)
```
levy5307, line 560: app_info &info

src/replica/replica.cpp
```diff
@@ -556,5 +557,16 @@ void replica::init_disk_tag()
     }
 }
 
+error_code replica::store_app_info(app_info &info, const std::string &path)
+{
+    replica_app_info new_info((app_info *)&info);
+    auto info_path = path.empty() ? utils::filesystem::path_combine(_dir, ".app-info") : path;
```
levy5307, line 563: const auto &info_path

src/replica/replica.h
```diff
@@ -451,6 +451,9 @@ class replica : public serverlet<replica>, public ref_counter, public replica_ba
 
     void init_disk_tag();
 
+    // store `ainfo` into `path` directory
```
levy5307, line 454: Could you add some comments here to show what would happen if path = ""?

src/replica/replica.cpp
```diff
@@ -556,5 +557,17 @@ void replica::init_disk_tag()
     }
 }
 
+error_code replica::store_app_info(app_info &info, const std::string &path)
+{
+    replica_app_info new_info((app_info *)&info);
+    const auto &info_path =
+        path.empty() ? utils::filesystem::path_combine(_dir, ".app-info") : path;
+    auto err = new_info.store(info_path.c_str());
+    if (err != ERR_OK) {
```
levy5307, line 566: I recommend to add dsn_unlikely if err != ERR_OK is rarely happen

levy5307 (approved):

src/replica/replica.cpp
```diff
@@ -556,5 +557,17 @@ void replica::init_disk_tag()
     }
 }
 
+error_code replica::store_app_info(app_info &info, const std::string &path)
+{
+    replica_app_info new_info((app_info *)&info);
+    const auto &info_path =
+        path.empty() ? utils::filesystem::path_combine(_dir, ".app-info") : path;
```
Shuo-Jia, line 564: suggest assign `str` to `const` variable: kAppInfo = ".app-info";

I had added variable: https://github.com/XiaoMi/rdsn/blob/982c2bd180a081fe2c772191cad0ea279e0e2685/src/replica/replica_disk_migrator.cpp#L35

you can refator it into `replica.h`

hycdong: Good idea, code updated.

Shuo-Jia (approved):

levy5307 (approved):

PR #1006, opened 2022-01-05
fix(network): fix asio core dump

Smityz:
ref: https://github.com/apache/incubator-pegasus/issues/863

The stress test will last for a week, after that this PR will ready for review


Smityz:
too many bugs during the test...

PR #1008, opened 2022-01-06
fix: message body size unset after parsed which leads to large io throughputs

padmejin:
## Issue
see https://github.com/apache/incubator-pegasus/issues/866

## Description

We have noticed pegasus 2.0 are using much more network/disk bandwidths than pegasus 1.12.3. After weeks of debugging, we have finally found that this is a bug introduced in #255, when a new thrift message parser that compatible with both old format and new format is refactored to replace the old one. Now we are trying to fix this.

The main cause is that when parsing the message body, instead of return exactly size of the body, it actually return the whole buffer that holds all the received messages. Thus, when there are many write requests piled up, from the moment the write request arrives,  to the moment it writes to mutation logs and to other nodes, the throughput is significantly amplified.

To fix that we merely add one line to set the message body size before calling create_message_from_request_blob().

We also noticed that when a message header is found invalid, the bad message will not be consumed and discarded, instead it stays in the buffer forever. We think it is a bug. To fix this bug, we consume the buffer before calling create_message_from_request_blob().

### DIFF
--- a/src/runtime/test/thrift_message_parser_test.cpp
+++ b/src/runtime/test/thrift_message_parser_test.cpp
@@ -30,7 +30,8 @@ DEFINE_TASK_CODE_RPC(RPC_TEST_THRIFT_MESSAGE_PARSER, TASK_PRIORITY_COMMON, THREA
 class thrift_message_parser_test : public testing::Test
 {
 public:
-    void mock_reader_read_data(message_reader &reader, const std::string &data, int message_count = 1)
+    void
+    mock_reader_read_data(message_reader &reader, const std::string &data, int message_count = 1)
     {
         char *buf = reader.read_buffer_ptr(data.length() * message_count);
         for (int i = 0; i < message_count; i++) {
--- a/src/runtime/test/thrift_message_parser_test.cpp
+++ b/src/runtime/test/thrift_message_parser_test.cpp
@@ -30,7 +30,7 @@ DEFINE_TASK_CODE_RPC(RPC_TEST_THRIFT_MESSAGE_PARSER, TASK_PRIORITY_COMMON, THREA
 class thrift_message_parser_test : public testing::Test
 {
 public:
-    void mock_reader_read_data(message_reader &reader, const std::string &data, int message_num)
+    void mock_reader_read_data(message_reader &reader, const std::string &data, int message_num = 1)
     {
         char *buf = reader.read_buffer_ptr(data.length() * message_num);
         for (int i = 0; i < message_num; i++) {
@@ -42,7 +42,7 @@ public:
     void test_get_message_on_receive_v0_data(message_reader &reader,
                                              apache::thrift::protocol::TMessageType messageType,
                                              bool is_request,
-                                             int message_num)
+                                             int message_num = 1)
     {
         /// write rpc message
         size_t body_length = 0;
@@ -122,7 +122,7 @@ public:
                                              apache::thrift::protocol::TMessageType messageType,
                                              bool is_request,
                                              bool is_backup_request,
-                                             int message_num)
+                                             int message_num = 1)
     {
         /// write rpc message
         size_t body_length = 0;
@@ -222,7 +222,7 @@ TEST_F(thrift_message_parser_test, get_message_on_receive_incomplete_second_fiel
         int read_next = 0;
         message_reader reader(64);
         data = std::string("THFT") + std::string(i, ' ');
-        mock_reader_read_data(reader, data, 1);
+        mock_reader_read_data(reader, data);
         ASSERT_EQ(reader._buffer_occupied, 4 + i);
         ASSERT_EQ(reader.buffer().size(), 4 + i);
 
@@ -253,7 +253,7 @@ TEST_F(thrift_message_parser_test, get_message_on_receive_incomplete_v0_hdr_len)
         out.write_u32(0);
         out.write_u32(48);
 
-        mock_reader_read_data(reader, data, 1);
+        mock_reader_read_data(reader, data);
         ASSERT_EQ(reader.buffer().size(), data.length());
 
         message_ex *msg = parser.get_message_on_receive(&reader, read_next);
@@ -279,7 +279,7 @@ TEST_F(thrift_message_parser_test, get_message_on_receive_invalid_v0_hdr_length)
         // hdr_length = i
         out.write_u32(i);
 
-        mock_reader_read_data(reader, data, 1);
+        mock_reader_read_data(reader, data);
         message_ex *msg = parser.get_message_on_receive(&reader, read_next);
         ASSERT_EQ(msg, nullptr);
         ASSERT_EQ(read_next, -1);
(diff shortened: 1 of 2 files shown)


src/runtime/test/thrift_message_parser_test.cpp
```diff
@@ -215,7 +222,7 @@ TEST_F(thrift_message_parser_test, get_message_on_receive_incomplete_second_fiel
         int read_next = 0;
         message_reader reader(64);
         data = std::string("THFT") + std::string(i, ' ');
-        mock_reader_read_data(reader, data);
+        mock_reader_read_data(reader, data, 1);
```
neverchanje, line 225: Here we can set `mock_reader_read_data` with default message_num to 1.

src/runtime/test/thrift_message_parser_test.cpp
```diff
@@ -30,16 +30,19 @@ DEFINE_TASK_CODE_RPC(RPC_TEST_THRIFT_MESSAGE_PARSER, TASK_PRIORITY_COMMON, THREA
 class thrift_message_parser_test : public testing::Test
 {
 public:
-    void mock_reader_read_data(message_reader &reader, const std::string &data)
+    void mock_reader_read_data(message_reader &reader, const std::string &data, int message_num)
```
neverchanje, line 33:
```suggestion
    void mock_reader_read_data(message_reader &reader, const std::string &data, int message_num = 1)
```

src/runtime/test/thrift_message_parser_test.cpp
```diff
@@ -30,16 +30,19 @@ DEFINE_TASK_CODE_RPC(RPC_TEST_THRIFT_MESSAGE_PARSER, TASK_PRIORITY_COMMON, THREA
...
+            memcpy(buf + i * data.length(), data.c_str(), data.length());
+            reader.mark_read(data.length());
+        }
     }
 
     void test_get_message_on_receive_v0_data(message_reader &reader,
                                              apache::thrift::protocol::TMessageType messageType,
-                                             bool is_request)
+                                             bool is_request,
+                                             int message_num)
```
neverchanje, line 45:
```suggestion
                                             bool is_request,
                                             int message_num = 1)
```

acelyc111:
Better to check that there is no extra buffer space after message_ex been constructed, i.e. at the end of parse_request_body_v0/v1, check msg->header->body_length is equal to msg->buffers[1].length()

hycdong:
I raise a new issue to describe this bug：https://github.com/apache/incubator-pegasus/issues/866. Could you please fulfill it according to your pull request?

padmejin:
> I raise a new issue to describe this bug：[apache/incubator-pegasus#866](https://github.com/apache/incubator-pegasus/issues/866). Could you please fulfill it according to your pull request?

done.

acelyc111 (dismissed):
LGTM

src/runtime/test/thrift_message_parser_test.cpp
padmejin: done.

Shuo-Jia (changes requested):
@levy5307 need re-check it

src/runtime/test/thrift_message_parser_test.cpp
```diff
@@ -30,16 +30,19 @@ DEFINE_TASK_CODE_RPC(RPC_TEST_THRIFT_MESSAGE_PARSER, TASK_PRIORITY_COMMON, THREA
 class thrift_message_parser_test : public testing::Test
 {
 public:
-    void mock_reader_read_data(message_reader &reader, const std::string &data)
+    void mock_reader_read_data(message_reader &reader, const std::string &data, int message_num = 1)
```
levy5307, line 33: `message_count` is a more accurate name

levy5307:
> We also noticed that when a message header is found invalid, the bad message will not be consumed and discarded, instead it stays in the buffer forever. We think it is a bug.
Could you add a unit test for this bug?

src/runtime/test/thrift_message_parser_test.cpp
padmejin: done

padmejin:
>

acelyc111 (approved):

padmejin:
> > We also noticed that when a message header is found invalid, the bad message will not be consumed and discarded, instead it stays in the buffer forever. We think it is a bug.
> 
> Could you add a unit test for this bug?

I meant to add a new test case, but then I found an old case that already cover this scenario. For some unknown reason, the author truncated the buffer before next message was parsed. I add a new line to check if the buffer is consumed. 

Before this patch this case failed on the assert, but it passed after this patch. So I believed it works.

Shuo-Jia:
> instead of return exactly size of the body, it actually return the whole buffer that holds all the received messages. Thus, when there are many write requests piled up, from the moment the write request arrives

Do you mean:

The request size task piled up queue as follow:
1KB 1KB 1KB 1KB 1KB 1KB

Then mutation log flush size will: 1KB、1K+1KB、1KB+1KB+1KB、1KB+1KB+1KB+1KB.......

padmejin:
> > instead of return exactly size of the body, it actually return the whole buffer that holds all the received messages. Thus, when there are many write requests piled up, from the moment the write request arrives
> 
> Do you mean:
> 
> The request size task piled up queue as follow: 1KB 1KB 1KB 1KB 1KB 1KB
> 
> Then mutation log flush size will: 1KB、1K+1KB、1KB+1KB+1KB、1KB+1KB+1KB+1KB.......

Yes.

Actually I printed the buffer size out and used awk & sort & uniq to analyze the distribution. The buffer size was up to 10000+, and all the number in a way conformed to arithmetic sequence. Not exactly, but followed some kinds of rules.

I also wrote a script to analyze all the slog file, and It turned out that a certain key pair appears multiple times in adjacent blocks, but when i dumped by `mlog_dump` I found there were no duplicated key pairs. I guessed those were just dirty data.

padmejin:
The following is a part of `mlog_dump` result. These three mutation log length were: [19971, 19649, 19327]. 

```
mutation [4.0.3.3873136]: gpid=4.0, ballot=3, decree=3873136, timestamp=1640322274780591, last_committed_decree=3873133, log_offset=22216353421, log_length=19971, update_count=1
  [MULTI_PUT] 1
    [PUT] "hashKeyPrefix_00000000000000000000000000000000000000000261387976" : "hashKeyPrefix_00000000000000000000000000000000000000000261387976" => 0 : "pElLpiaC0juYshL86KXI30N5mYvNlwQNkO5k44hKInyhyomgdO4Eswz2RWDrODyP"
mutation [4.0.3.3873137]: gpid=4.0, ballot=3, decree=3873137, timestamp=1640322274781203, last_committed_decree=3873135, log_offset=22216373392, log_length=19649, update_count=1
  [MULTI_PUT] 1
    [PUT] "hashKeyPrefix_00000000000000000000000000000000000000000054160672" : "hashKeyPrefix_00000000000000000000000000000000000000000054160672" => 0 : "drsrinPuO8Zty7nP0YU2RRyQKoSQCBBBcGUaFKOz4Ufk5C2pgPHf0pekCVDCuE7I"
mutation [4.0.3.3873138]: gpid=4.0, ballot=3, decree=3873138, timestamp=1640322274781608, last_committed_decree=3873135, log_offset=22216393041, log_length=19327, update_count=1
  [MULTI_PUT] 1
    [PUT] "hashKeyPrefix_00000000000000000000000000000000000000000336035963" : "hashKeyPrefix_00000000000000000000000000000000000000000336035963" => 0 : "c8kxhTIEw6c6TSAf0jDlV8R01QulIeHSIK3O364HDD7PP8CQGTNs4PldV9mz1PLr"
```

acelyc111 (approved):

Shuo-Jia (approved):

PR #1011, opened 2022-01-10
fix(bulk_load): update ingestion error handling with move_files mode

hycdong:
As https://github.com/apache/incubator-pegasus/issues/877 shows, move files during ingestion will speech up ingestion and decrease disk write throughput. (https://github.com/apache/incubator-pegasus/pull/864)

However, it exposes that in some error handling cases, it is possible to execute an out-dated ingestion request repeated. 

In previous `copy_files` mode, it won't have problems, just ingest files repeatly, now is `move_files` mode, once ingestion is succeed, the original files doesn't exist, which will lead to unexpected ingestion failed.

This pull request is to update error handling during ingestion, including:
- Delay to send ingestion_request to make replica server receive request with ingestion status
- Add ballot in `ingestion_request` (It will be used in pegasus repo to do ballot check)

### DIFF
--- a/include/dsn/dist/replication/replication_app_base.h
+++ b/include/dsn/dist/replication/replication_app_base.h
@@ -106,7 +106,7 @@ public:
     // Whether this replica is duplicating.
     bool is_duplicating() const;
 
-    ballot get_ballot() const;
+    const ballot &get_ballot() const;
 
     //
     // Open the app.
--- a/include/dsn/dist/replication/replication_app_base.h
+++ b/include/dsn/dist/replication/replication_app_base.h
@@ -106,6 +106,8 @@ public:
     // Whether this replica is duplicating.
     bool is_duplicating() const;
 
+    ballot get_ballot() const;
+
     //
     // Open the app.
     //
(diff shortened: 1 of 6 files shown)


include/dsn/dist/replication/replication_app_base.h
```diff
@@ -106,6 +106,8 @@ class replication_app_base : public replica_base
     // Whether this replica is duplicating.
     bool is_duplicating() const;
 
+    ballot get_ballot() const;
```
levy5307, line 109: const ballot& get_ballot() const;

acelyc111: In my opinon, that's OK, ballot is just a int64.

levy5307: I don't think so. It's better to use reference to all of the user-defined type. Because it may be changed someday.

levy5307: Or change it to `int64 get_ballot() const;`

hycdong: Done

levy5307 (approved):

Smityz (approved):

PR #1015, opened 2022-01-10
refactor: remove deprecated replica .info file

acelyc111:
related issue:
[apache/incubator-pegasus#876](https://github.com/apache/incubator-pegasus/issues/876)

### DIFF
--- a/src/replica/replication_app_base.cpp
+++ b/src/replica/replication_app_base.cpp
@@ -45,107 +45,26 @@ namespace replication {
 
 error_code replica_init_info::load(const std::string &dir)
 {
-    std::string old_info_path = utils::filesystem::path_combine(dir, ".info");
-    std::string new_info_path = utils::filesystem::path_combine(dir, ".init-info");
-    std::string load_path;
-    error_code err;
-    if (utils::filesystem::path_exists(new_info_path)) {
-        load_path = new_info_path;
-        err = load_json(new_info_path.c_str());
-    } else {
-        load_path = old_info_path;
-        err = load_binary(old_info_path.c_str());
-    }
-
-    ERR_LOG_AND_RETURN_NOT_OK(err, "load replica_init_info from {} failed", load_path);
-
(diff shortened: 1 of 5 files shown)


levy5307 (dismissed):

src/replica/replication_app_base.cpp
```diff
@@ -45,107 +45,26 @@ namespace replication {
...
-                                  old_info_path,
-                                  new_info_path);
-
-        ddebug_f(
-            "change replica_init_info file from {} to {} succeed", old_info_path, new_info_path);
-        dassert_f(
-            utils::filesystem::remove_path(old_info_path), "remove file {} failed", old_info_path);
-    }
-
+    std::string info_path = utils::filesystem::path_combine(dir, ".init-info");
```
Shuo-Jia, line 48: I suggest `.init-info` defined as a const variable, which will more safe when someone to update it, like https://github.com/XiaoMi/rdsn/blob/6af2c3204c889a83dd0767383ec60ea65f089a9b/src/replica/replica.cpp#L53

src/replica/replication_app_base.cpp
```diff
@@ -45,107 +45,26 @@ namespace replication {
...
+    dassert_f(utils::filesystem::path_exists(info_path), "file({}) not exist", info_path);
+    ERR_LOG_AND_RETURN_NOT_OK(
+        load_json(info_path.c_str()), "load replica_init_info from {} failed", info_path);
+    ddebug_f("load replica_init_info from {} succeed: {}", info_path, to_string());
```
Shuo-Jia, line 52: you seem change the origin logic, the origin code allow to load `.info`?

src/replica/replication_app_base.cpp
hycdong: +1

Shuo-Jia (approved):

levy5307 (approved):

PR #1017, opened 2022-01-10
fix: use fsync() to prevent .app-info/.init-info lost risk on XFS after power outage

acelyc111:
ref issue:
https://github.com/apache/incubator-pegasus/issues/879

### DIFF
--- a/src/replica/replication_app_base.cpp
+++ b/src/replica/replication_app_base.cpp
@@ -30,6 +30,7 @@
 #include <dsn/utils/latency_tracer.h>
 #include <dsn/dist/fmt_logging.h>
 #include <dsn/dist/replication/replication_app_base.h>
+#include <dsn/utility/defer.h>
 #include <dsn/utility/factory_store.h>
 #include <dsn/utility/filesystem.h>
 #include <dsn/utility/crc.h>
@@ -53,6 +54,7 @@ error_code write_blob_to_file(const std::string &file, const blob &data)
     std::string tmp_file = file + ".tmp";
     disk_file *hfile = file::open(tmp_file.c_str(), O_WRONLY | O_CREAT | O_BINARY | O_TRUNC, 0666);
     ERR_LOG_AND_RETURN_NOT_TRUE(hfile, ERR_FILE_OPERATION_FAILED, "open file {} failed", tmp_file);
+    auto cleanup = defer([tmp_file]() { utils::filesystem::remove_path(tmp_file); });
 
     error_code err;
     size_t sz = 0;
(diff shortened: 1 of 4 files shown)


src/replica/replication_app_base.cpp
```diff
@@ -177,18 +194,32 @@ error_code replica_app_info::store(const char *file)
...
-    std::ofstream os(tmp_file.c_str(),
-                     (std::ofstream::out | std::ios::binary | std::ofstream::trunc));
-    ERR_LOG_AND_RETURN_NOT_TRUE(
-        os.is_open(), ERR_FILE_OPERATION_FAILED, "open file {} failed", tmp_file);
-
-    auto data = writer.get_buffer();
-    os.write((const char *)data.data(), (std::streamsize)data.length());
-    os.close();
-
+    disk_file *hfile = file::open(tmp_file.c_str(), O_WRONLY | O_CREAT | O_BINARY | O_TRUNC, 0666);
```
levy5307, line 200: I find a lot of same code between L106~L130 and L200~L222, so could you refactor it by encapsulating them into a standalone function?

src/replica/replication_app_base.cpp
```diff
@@ -45,6 +45,46 @@ namespace replication {
...
+                                   },
+                                   0);
+    dassert_f(tsk, "create file::write task failed");
+    tracker.wait_outstanding_tasks();
+    file::flush(hfile);
+    file::close(hfile);
+    ERR_LOG_AND_RETURN_NOT_OK(err, "write file {} failed", tmp_file);
+    dcheck_eq(data.length(), sz);
+    // TODO(yingchun): need fsync too？
+    ERR_LOG_AND_RETURN_NOT_TRUE(utils::filesystem::rename_path(tmp_file, file),
```
levy5307, line 78: Should we remove tmp file if `rename_path` returns error?

acelyc111: In fact, I prefer to use assert here, defensive programming will make our code too complex, and the upper layer doesn't deal with every failed cases well actually.
So keep the logic the same as before for this PR, we can do more refactor in the future.

src/replica/replication_app_base.cpp
```diff
@@ -45,12 +45,52 @@ namespace replication {
 
 const std::string replica_init_info::kInitInfo = ".init-info";
 
+DEFINE_TASK_CODE_AIO(LPC_AIO_INFO_WRITE, TASK_PRIORITY_COMMON, THREAD_POOL_DEFAULT)
+
+namespace {
+error_code write_blob_to_file(const std::string &file, const blob &data)
+{
+    std::string tmp_file = file + ".tmp";
+    disk_file *hfile = file::open(tmp_file.c_str(), O_WRONLY | O_CREAT | O_BINARY | O_TRUNC, 0666);
```
levy5307, line 54: I mean add `auto cleanup = defer([hfile]() {  // remove hfile; });` here. 
so we can delete the temp file before coredump.

levy5307 (approved):

Smityz (approved):

PR #1018, opened 2022-01-17
feat(bulk_load): avoid unecessary repeated ingestion

hycdong:
As https://github.com/apache/incubator-pegasus/issues/861 shows, this pull request implements avoiding repeated ingestion, including:
- Add `ever_ingest_succeed` and `addresses` in `partition_bulk_load_info` to store address of replica server who has executed ingestion succeed, this structure will store in zk while partition ingestion succeed, reference function`update_partition_info_on_remote_storage`
- Add `check_ever_ingestion_succeed` function before sending ingestion request, if this partition has already executed ingestion succeed and replica nodes not changed, it shoudn't send ingestion, and convert partition bulk load status into `succeed`.
- Mark download into succeed transition valid in replica, reference function `replica_bulk_loader::validate_status`
- Add and update related unit tests

### DIFF
--- a/src/meta/meta_bulk_load_service.cpp
+++ b/src/meta/meta_bulk_load_service.cpp
@@ -951,7 +951,7 @@ void bulk_load_service::update_partition_info_on_remote_storage(const std::strin
     }
 
     _partitions_pending_sync_flag[pid] = true;
-    update_partition_info(pid, new_status, pinfo);
+    update_partition_info_unlock(pid, new_status, pinfo);
 
     blob value = json::json_forwarder<partition_bulk_load_info>::encode(pinfo);
     _meta_svc->get_meta_storage()->set_data(
@@ -966,9 +966,9 @@ void bulk_load_service::update_partition_info_on_remote_storage(const std::strin
 }
 
 // ThreadPool: THREAD_POOL_META_STATE
-void bulk_load_service::update_partition_info(const gpid &pid,
-                                              bulk_load_status::type new_status,
-                                              /*out*/ partition_bulk_load_info &pinfo)
+void bulk_load_service::update_partition_info_unlock(const gpid &pid,
+                                                     bulk_load_status::type new_status,
+                                                     /*out*/ partition_bulk_load_info &pinfo)
 {
     auto old_status = pinfo.status;
     pinfo.status = new_status;
@@ -979,7 +979,7 @@ void bulk_load_service::update_partition_info(const gpid &pid,
         return;
     }
     pinfo.addresses.clear();
-    const auto state = _partitions_bulk_load_state[pid];
+    const auto &state = _partitions_bulk_load_state[pid];
     for (const auto &kv : state) {
         pinfo.addresses.emplace_back(kv.first);
     }
--- a/src/meta/meta_bulk_load_service.cpp
+++ b/src/meta/meta_bulk_load_service.cpp
@@ -287,6 +287,7 @@ void bulk_load_service::create_partition_bulk_load_dir(const std::string &app_na
 {
     partition_bulk_load_info pinfo;
     pinfo.status = bulk_load_status::BLS_DOWNLOADING;
+    pinfo.ever_ingest_succeed = false;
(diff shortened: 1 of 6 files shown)


src/meta/test/meta_bulk_load_service_test.cpp
```diff
@@ -185,6 +228,16 @@ class bulk_load_service_test : public meta_test_base
         return bulk_svc().get_app_bulk_load_status_unlocked(app_id);
     }
 
+    partition_bulk_load_info get_partition_bulk_load_info(const gpid &pid)
```
levy5307, line 231: const partition_bulk_load_info&

src/meta/meta_bulk_load_service.cpp
```diff
@@ -293,6 +293,7 @@ void bulk_load_service::create_partition_bulk_load_dir(const std::string &app_na
 {
     partition_bulk_load_info pinfo;
     pinfo.status = bulk_load_status::BLS_DOWNLOADING;
+    pinfo.ever_ingest_succeed = false;
```
Shuo-Jia, line 296: has the last status stored in zk? if stored, can you use the status, but no define new variable?

hycdong: `last status` won't store in zk, only current status will be stored.

src/meta/meta_bulk_load_service.cpp
```diff
@@ -950,31 +951,53 @@ void bulk_load_service::update_partition_info_on_remote_storage(const std::strin
...
+    auto old_status = pinfo.status;
+    pinfo.status = new_status;
+    if (old_status != bulk_load_status::BLS_INGESTING ||
+        new_status != bulk_load_status::BLS_SUCCEED ||
+        _partitions_bulk_load_state.find(pid) == _partitions_bulk_load_state.end()) {
+        // no need to update other field of partition_bulk_load_info
+        return;
+    }
+    pinfo.addresses.clear();
+    const auto state = _partitions_bulk_load_state[pid];
```
levy5307, line 982: const auto &state

src/meta/meta_bulk_load_service.cpp
```diff
@@ -1120,6 +1143,41 @@ void bulk_load_service::update_app_status_on_remote_storage_reply(const app_bulk
     }
 }
 
+// ThreadPool: THREAD_POOL_META_STATE
+bool bulk_load_service::check_ever_ingestion_succeed(const partition_configuration &config,
+                                                     const std::string &app_name,
+                                                     const gpid &pid)
+{
+    partition_bulk_load_info pinfo;
```
levy5307, line 1151: partition_bulk_load_info *pinfo;
    {
        zauto_read_lock l(_lock);
        pinfo = &_partition_bulk_load_info[pid];
    }

src/meta/meta_bulk_load_service.cpp
Shuo-Jia: if current status stored, when failed and re-start,  will it change to `last_status`?

src/meta/meta_bulk_load_service.cpp
Shuo-Jia: Why use * ？

src/meta/test/meta_bulk_load_service_test.cpp
```diff
@@ -163,6 +163,49 @@ class bulk_load_service_test : public meta_test_base
         return flag;
     }
 
+    void set_partition_bulk_load_info(const gpid &pid,
+                                      bool ever_ingest_succeed,
+                                      bool use_secondary3 = false)
+    {
+        partition_bulk_load_info pinfo;
```
levy5307, line 170: partition_bulk_load_info &pinfo =  bulk_svc()._partition_bulk_load_info[pid]

src/meta/meta_bulk_load_service.cpp
hycdong: In this function, `_partition_bulk_load_info[pid]` won't be updated. I think it's okay to use the value copy here.

levy5307 (approved):

Shuo-Jia (approved):

PR #1025, opened 2022-01-19
fix: break when disable latency tracer in macro

Shuo-Jia:
https://github.com/apache/incubator-pegasus/issues/883

macro with latency tracer should break when disable.


levy5307:
fixed in https://github.com/XiaoMi/rdsn/pull/1023

PR #1032, opened 2022-01-20
refactor: change the type of prepare_list from raw pointer to std::unique_ptr

levy5307:
Use `std::unique_ptr` to avoid memory leak, for example: L132 in src/replica/split/test/replica_split_test.cpp

https://github.com/apache/incubator-pegasus/issues/887

### DIFF
--- a/src/replica/split/test/replica_split_test.cpp
+++ b/src/replica/split/test/replica_split_test.cpp
@@ -129,7 +129,7 @@ public:
                 mu->set_logged();
             }
         }
-        rep->_prepare_list = _mock_plist;
+        rep->_prepare_list.reset(_mock_plist);
     }
 
     void mock_parent_states()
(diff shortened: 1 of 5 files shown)


src/replica/split/test/replica_split_test.cpp
```diff
@@ -129,7 +129,7 @@ class replica_split_test : public replica_test_base
                 mu->set_logged();
             }
         }
-        rep->_prepare_list = _mock_plist;
```
levy5307, line 132: memory leak in previous implementation

hycdong (approved):

Shuo-Jia (approved):

src/replica/test/mock_utils.h
```diff
@@ -149,7 +149,7 @@ class mock_replica : public replica
     void set_replica_config(replica_configuration &config) { _config = config; }
     void set_partition_status(partition_status::type status) { _config.status = status; }
     void set_last_committed_decree(decree d) { _prepare_list->reset(d); }
-    prepare_list *get_plist() { return _prepare_list; }
+    prepare_list *get_plist() const { return _prepare_list.get(); }
```
Smityz, line 152: why not return unique_ptr directly

PR #1033, opened 2022-01-20
feat: implement long adder to optimize the counter of new metrics system

empiredan:
https://github.com/apache/incubator-pegasus/issues/889


Smityz:
LGTM

Smityz:
Can you construct a benchmark for its performance compared with the origin perf_counter?

empiredan:
> Can you construct a benchmark for its performance compared with the origin perf_counter?

Actually the benchmark given in https://github.com/apache/incubator-pegasus/issues/889 has involved the comparison with the original perf-counter: `divided_long_adder` is nearly the same with the original `dsn::perf_counter_number_atomic` except that its virtual functions are removed since other long adders also have no virtual function.

The implementation of `divided_long_adder` can be found in `src/utils/long_adder_bench/long_adder_bench.cpp`.

acelyc111:
What's th difference between this PR's implemention and the original https://github.com/apache/kudu/blob/master/src/kudu/util/striped64.h?

empiredan:
> What's th difference between this PR's implemention and the original https://github.com/apache/kudu/blob/master/src/kudu/util/striped64.h?

The data structures and algorithms between both striped long adder are based on striped64. The differences are:

- This PR defines a default constructor for the atomic cache-line-aligned int64, while kudu defines a non-trivial one.
- This PR encapsulates the "create function" of the atomic cache-line-aligned int64 as a common utility which is also exposed and can be used by external modules.
- This PR introduces `unique_ptr<...>` to manage the memory of cells, while kudu releases memory by calling `free()` in the destructor.
- This PR supports `fetch_and_reset`, since we have `volatile` scenarios where only recent count will be showed.
- This PR fixes some styles, for example, replace `if (c)` with `if (c != nullptr)`.
- This PR uses `dsn::utils::get_current_tid()` to get the initial hash code, while kudu uses `Random` to get.
- In "internal reset", this PR only sets "base" with target value, with cells set to zero, while kudu set all of "base" and cells to target value.
- Besides striped_long_adder, this PR also implements `long_adder_wrapper` to optimize the scenario that the adder is updated very frequently.

acelyc111 (approved):
LGTM

Smityz (approved):

PR #1034, opened 2022-01-23
feat(build): support to build on MacOS

acelyc111:
ref issue: https://github.com/apache/incubator-pegasus/issues/813

This patch includes the following modifies to support building both on Linux and MacOS:
- Skip finding some libs, add some include/link paths, disable gperf, in cmake.
    - This is a initial work to build on MacOS, we can optmize them in the future.
- Improve syntax to compile by clang
    - E.g. include necessary headers, 
- Improve portability
    - E.g. include/dsn/utility/endians.h, include/dsn/utility/ports.h, src/meta/dump_file.h, src/runtime/task/task_worker.cpp.
- Remove Windows platform related code.
- Bump poco from 1.7.8 to 1.11.1, and add patch for FDS to support build on MacOS
- Build hadoop lib from source to obtain native libs


bin/dsn.cmake
```diff
@@ -267,19 +267,27 @@ function(dsn_setup_system_libs)
...
     set(DSN_SYSTEM_LIBS ${DSN_SYSTEM_LIBS} ${OPENSSL_CRYPTO_LIBRARY})
+    if (APPLE)
+        include_directories(SYSTEM ${OPENSSL_ROOT_DIR}/include)
+        link_directories("${OPENSSL_ROOT_DIR}/lib")
+    endif()
 
-    if(ENABLE_GPERF)
-        set(DSN_SYSTEM_LIBS ${DSN_SYSTEM_LIBS} tcmalloc_and_profiler)
-        add_definitions(-DDSN_ENABLE_GPERF)
+    if (NOT APPLE)
```
Shuo-Jia, line 286: Is better defining `MACOS`

levy5307: +1

src/block_service/fds/fds_service.cpp
```diff
@@ -674,12 +673,10 @@ dsn::task_ptr fds_file_object::download(const download_request &req,
     std::shared_ptr<std::ofstream> handle(new std::ofstream(
         req.output_local_name, std::ios::binary | std::ios::out | std::ios::trunc));
     if (!handle->is_open()) {
-        char buffer[512];
-        char *ptr = strerror_r(errno, buffer, 512);
-        derror("fds download failed: fail to open localfile(%s) when download(%s), error(%s)",
-               req.output_local_name.c_str(),
-               _fds_path.c_str(),
-               ptr);
+        derror_f("fds download failed: fail to open localfile({}) when download({}), error({})",
```
levy5307, line 676: I found that there are a log code refactor in this pull request. So in my opinion,  it's better to split it into server small pull requests.

bin/dsn.cmake
acelyc111: `APPLE ` is a variable in cmake, https://cmake.org/cmake/help/v3.1/variable/APPLE.html
I didn't find where is `MACOS` defined.

Shuo-Jia: Well, it's ok

src/block_service/fds/fds_service.cpp
acelyc111: In fact this is also a fix to adapt to macOS, using the portable function `safe_strerror` instead of `strerror_r`.

Smityz (approved):

neverchanje (approved):
Perfectly LGTM, but not yet tested myself.

Shuo-Jia (approved):

PR #1035, opened 2022-01-24
feat(bulk_load): support disk_level ingesting restriction part1 - add ingestion_context class

hycdong:
As https://github.com/apache/incubator-pegasus/issues/886 shows, we plan to provide a disk-level concurrent ingesting count restriction. This pull request adds a new class called `ingestion_context`, recording the ingesting partitions and every node and disk ingesting count.

This pull request also adds two configurations to implememt the restriction:
- `bulk_load_node_max_ingesting_count` - restrict node max ingesting partition count
- `bulk_load_node_min_disk_count` - node min disk count

For example:
- `node_max_count=1`, `disk_count=3`: only one partition can do ingestion
- `node_max_count=3`, `disk_count=3`: each disk can only have one partition do ingestion
- `node_max_count=4`, `disk_count=3`: each disk can have 2 partitions do ingestion, but node concurrent count is 4

Configuration changes
```diff
[meta_server]
+ bulk_load_node_max_ingesting_count = 4
+ bulk_load_node_min_disk_count = 1
```

### DIFF
--- a/src/meta/meta_bulk_load_ingestion_context.cpp
+++ b/src/meta/meta_bulk_load_ingestion_context.cpp
@@ -63,8 +63,8 @@ void ingestion_context::node_context::init_disk(const std::string &disk_tag)
     disk_count[disk_tag] = 0;
 }
 
-int32_t
-ingestion_context::node_context::get_max_disk_count(const int32_t max_node_ingestion_count) const
+int32_t ingestion_context::node_context::get_max_disk_ingestion_count(
+    const int32_t max_node_ingestion_count) const
 {
     FAIL_POINT_INJECT_F("ingestion_node_context_disk_count", [](string_view count_str) -> int32_t {
         auto count = 0;
@@ -81,13 +81,13 @@ ingestion_context::node_context::get_max_disk_count(const int32_t max_node_inges
     return max_node_ingestion_count / node_disk_count + 1;
 }
 
-bool ingestion_context::node_context::try_add(const std::string &disk_tag)
+bool ingestion_context::node_context::check_if_add(const std::string &disk_tag)
 {
     auto max_node_ingestion_count = FLAGS_bulk_load_node_max_ingesting_count;
-    if (count >= max_node_ingestion_count) {
+    if (node_count >= max_node_ingestion_count) {
         dwarn_f("node[{}] has {} partition executing ingestion, max_count = {}",
                 address.to_string(),
-                count,
+                node_count,
                 max_node_ingestion_count);
         return false;
     }
@@ -95,7 +95,7 @@ bool ingestion_context::node_context::try_add(const std::string &disk_tag)
     if (disk_count.find(disk_tag) == disk_count.end()) {
         disk_count[disk_tag] = 0;
     }
-    auto max_disk_ingestion_count = get_max_disk_count(max_node_ingestion_count);
+    auto max_disk_ingestion_count = get_max_disk_ingestion_count(max_node_ingestion_count);
     if (disk_count[disk_tag] >= max_disk_ingestion_count) {
         dwarn_f("node[{}] disk[{}] has {} partition executing ingestion, max_count = {}",
                 address.to_string(),
@@ -110,12 +110,12 @@ bool ingestion_context::node_context::try_add(const std::string &disk_tag)
 void ingestion_context::node_context::add(const std::string &disk_tag)
 {
     disk_count[disk_tag]++;
-    count++;
+    node_count++;
 }
 
 void ingestion_context::node_context::decrease(const std::string &disk_tag)
 {
-    count--;
+    node_count--;
     disk_count[disk_tag]--;
 }
 
@@ -144,7 +144,7 @@ bool ingestion_context::check_node_ingestion(const rpc_address &node, const std:
     if (_nodes_context.find(node) == _nodes_context.end()) {
         _nodes_context[node] = node_context(node, disk_tag);
     }
-    return _nodes_context[node].try_add(disk_tag);
+    return _nodes_context[node].check_if_add(disk_tag);
 }
 
 void ingestion_context::add_partition(const partition_node_info &info)
--- a/src/meta/meta_bulk_load_ingestion_context.cpp
+++ b/src/meta/meta_bulk_load_ingestion_context.cpp
@@ -66,7 +66,7 @@ uint32_t ingestion_context::node_context::get_max_disk_ingestion_count(
 {
     FAIL_POINT_INJECT_F("ingestion_node_context_disk_count", [](string_view count_str) -> uint32_t {
         auto count = 0;
-        buf2int32(count_str, count);
+        buf2uint32(count_str, count);
         return count;
     });
 
--- a/src/meta/meta_bulk_load_ingestion_context.cpp
+++ b/src/meta/meta_bulk_load_ingestion_context.cpp
@@ -1,0 +1,204 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#include "meta_bulk_load_ingestion_context.h"
+
+#include <dsn/dist/fmt_logging.h>
+#include <dsn/utility/fail_point.h>
+
+namespace dsn {
+namespace replication {
+
+DSN_DEFINE_uint32("meta_server",
+                  bulk_load_max_ingestion_concurrent_count,
+                  4,
+                  "max partition_count executing ingestion for one node at the same time");
+DSN_TAG_VARIABLE(bulk_load_max_ingestion_concurrent_count, FT_MUTABLE);
+
+DSN_DEFINE_uint32("meta_server", bulk_load_node_min_disk_count, 1, "min disk count of one node");
+DSN_TAG_VARIABLE(bulk_load_node_min_disk_count, FT_MUTABLE);
+
+ingestion_context::ingestion_context() { reset_all(); }
+
+ingestion_context::~ingestion_context() { reset_all(); }
+
+// ingestion_context::partition_node_info
+void ingestion_context::partition_node_info::create(const partition_configuration &config,
+                                                    const config_context &cc)
+{
+    pid = config.pid;
+    std::unordered_set<rpc_address> current_nodes;
+    current_nodes.insert(config.primary);
+    for (const auto &secondary : config.secondaries) {
+        current_nodes.insert(secondary);
+    }
+    for (const auto &node : current_nodes) {
+        std::string disk_tag;
+        if (cc.get_disk_tag(node, disk_tag)) {
+            node_disk[node] = disk_tag;
+        }
+    }
+}
+
+// ingestion_context::node_context
+void ingestion_context::node_context::init_disk(const std::string &disk_tag)
+{
+    if (disk_count.find(disk_tag) != disk_count.end()) {
+        return;
+    }
+    disk_count[disk_tag] = 0;
+}
+
+int32_t
+ingestion_context::node_context::get_max_disk_count(const int32_t max_node_ingestion_count) const
+{
+    FAIL_POINT_INJECT_F("ingestion_node_context_disk_count", [](string_view count_str) -> int32_t {
+        auto count = 0;
+        buf2int32(count_str, count);
+        return count;
+    });
+
+    auto node_disk_count = disk_count.size() > FLAGS_bulk_load_node_min_disk_count
+                               ? disk_count.size()
+                               : FLAGS_bulk_load_node_min_disk_count;
+    if (max_node_ingestion_count % node_disk_count == 0) {
+        return max_node_ingestion_count / node_disk_count;
+    }
+    return max_node_ingestion_count / node_disk_count + 1;
+}
+
+bool ingestion_context::node_context::try_add(const std::string &disk_tag)
+{
+    auto max_node_ingestion_count = FLAGS_bulk_load_max_ingestion_concurrent_count;
+    if (count >= max_node_ingestion_count) {
+        dwarn_f("node[{}] has {} partition executing ingestion, max_count = {}",
+                address.to_string(),
+                count,
+                max_node_ingestion_count);
+        return false;
+    }
+
+    if (disk_count.find(disk_tag) == disk_count.end()) {
+        disk_count[disk_tag] = 0;
+    }
+    auto max_disk_ingestion_count = get_max_disk_count(max_node_ingestion_count);
+    if (disk_count[disk_tag] >= max_disk_ingestion_count) {
+        dwarn_f("node[{}] disk[{}] has {} partition executing ingestion, max_count = {}",
+                address.to_string(),
+                disk_tag,
+                disk_count[disk_tag],
+                max_disk_ingestion_count);
+        return false;
+    }
+    return true;
+}
+
+void ingestion_context::node_context::add(const std::string &disk_tag)
+{
+    disk_count[disk_tag]++;
+    count++;
+}
+
+void ingestion_context::node_context::decrease(const std::string &disk_tag)
+{
+    count--;
+    disk_count[disk_tag]--;
+}
+
+// ingestion_context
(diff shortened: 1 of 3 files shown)


src/meta/meta_bulk_load_ingestion_context.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+ingestion_context::ingestion_context() { reset_all(); }
+
+ingestion_context::~ingestion_context() { reset_all(); }
+
+// ingestion_context::partition_node_info
+void ingestion_context::partition_node_info::create(const partition_configuration &config,
+                                                    const config_context &cc)
+{
+    pid = config.pid;
```
levy5307, line 43: I found that pid is useless

Shuo-Jia:
> * `bulk_load_node_max_ingesting_count` - restrict node max ingesting partition count

Actually, if you  restrict `disk count`,  `bulk_load_node_max_ingesting_count` seem to be useless, it's ok that you just control disk load. 

I  think it's more complex to consider  two `count`,  you can consider that just use one `count` whether more simple?

Smityz:
> 

+1, I think `max_ingesting_count_per_disk`  can meet demands, 2 options to restrict bulkload will increase the burden of understanding.

Smityz:
> > * `bulk_load_node_max_ingesting_count` - restrict node max ingesting partition count
> 
> Actually, if you restrict `disk count`, `bulk_load_node_max_ingesting_count` seem to be useless, it's ok that you just control disk load.
> 
> I think it's more complex to consider two `count`, you can consider that just use one `count` whether more simple?

> 

+1, I think `max_ingesting_count_per_disk`  can meet demands, 2 options to restrict bulkload will increase the burden of understanding.

hycdong:
@Shuo-Jia @Smityz Thanks for your suggestions~ I meant to define only one configuration called `max_disk_ingesting_count`, but used the `node_count+disk_count`, there are two reasons:
1. In different production environment, node may have different disk count. 
- For example, 3 disks, 7 disks, 12 disks, or only one disks. **It is not clear enough to show how many partitions are ingesting in one node at the same time.** 
- If we have node_count+disk_count, for example, node_count is 4, disk count is 3, it means for each node, there are at most 4 partitions ingesting in this node, and for each disk, there may have 1 partition or 2 partitions ingesting. It is clear for admin use to adjust ingesting restriction.
2. Using two configurations, the logic can be implemented more graceful. 
- In current meta server design, the node address of partition will store in zk, but its disk information are reported by node, and only store in meta server memory structure called `config_context`, meta server even doesn't store the disk count of a replica server. Meta server can only traversal each partition's `config_context` structure, then calculate how many disks a replica server have. 
- The disk count is only an example, there also have similar calculations which will make code complex, I think adding a configuration(`disk_count`) is a easier and more graceful way.

src/meta/meta_bulk_load_ingestion_context.cpp
hycdong: It just like a name of `partition_node_info` to show which partition it is.

levy5307: What I mean is in this function, the variable named pid is useless. I think you got a mistake

hycdong: Ohh, this function name is `create`, just used to initialize structure `partition_node_info`, so it will init all members of it.

src/meta/meta_bulk_load_ingestion_context.cpp
```diff
@@ -0,0 +1,204 @@
...
+    auto node_disk_count = disk_count.size() > FLAGS_bulk_load_node_min_disk_count
+                               ? disk_count.size()
+                               : FLAGS_bulk_load_node_min_disk_count;
+    if (max_node_ingestion_count % node_disk_count == 0) {
+        return max_node_ingestion_count / node_disk_count;
+    }
+    return max_node_ingestion_count / node_disk_count + 1;
+}
+
+bool ingestion_context::node_context::try_add(const std::string &disk_tag)
```
Shuo-Jia, line 84: The method is just judge `whether_to_add`like `check_node_ingestion`, but `try_add`?  I don't see `add logic code`, name it `check_if_add`?

src/meta/meta_bulk_load_ingestion_context.cpp
```diff
@@ -0,0 +1,204 @@
...
+                disk_count[disk_tag],
+                max_disk_ingestion_count);
+        return false;
+    }
+    return true;
+}
+
+void ingestion_context::node_context::add(const std::string &disk_tag)
+{
+    disk_count[disk_tag]++;
```
Shuo-Jia, line 112: same, I suggest you name all "count" type variable more detail name

src/meta/meta_bulk_load_ingestion_context.cpp
```diff
@@ -0,0 +1,204 @@
...
+    node_count++;
+}
+
+void ingestion_context::node_context::decrease(const std::string &disk_tag)
+{
+    node_count--;
```
Shuo-Jia, line 118: People may understand `node_count` means `count of node`, is better named `node_ingesting_count` ?

The follow is same,  `disk_ingesting_count`?

hycdong: Those belongs to `ingestion_context` class, and they won't be accessed out of this class, I think adding `ingesting` for their name to a little bit duplicated.

Shuo-Jia (dismissed):

src/meta/meta_bulk_load_ingestion_context.cpp
```diff
@@ -0,0 +1,204 @@
...
+int32_t ingestion_context::node_context::get_max_disk_ingestion_count(
+    const int32_t max_node_ingestion_count) const
+{
+    FAIL_POINT_INJECT_F("ingestion_node_context_disk_count", [](string_view count_str) -> int32_t {
+        auto count = 0;
+        buf2int32(count_str, count);
+        return count;
+    });
+
+    auto node_disk_count = disk_count.size() > FLAGS_bulk_load_node_min_disk_count
```
levy5307, line 75: auto node_disk_count = std::min(disk_count.size(), FLAGS_bulk_load_node_min_disk_count)

src/meta/meta_bulk_load_ingestion_context.cpp
```diff
@@ -0,0 +1,204 @@
...
+                               ? disk_count.size()
+                               : FLAGS_bulk_load_node_min_disk_count;
+    if (max_node_ingestion_count % node_disk_count == 0) {
```
levy5307, line 78: `return (max_node_ingestion_count + node_disk_count - 1) / node_disk_count;`

src/meta/meta_bulk_load_ingestion_context.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+void ingestion_context::remove_partition(const gpid &pid)
+{
+    FAIL_POINT_INJECT_F("ingestion_context_remove_partition",
+                        [=](string_view) { _running_partitions.erase(pid); });
+
+    if (_running_partitions.find(pid) == _running_partitions.end()) {
+        return;
+    }
+    auto info = _running_partitions[pid];
```
levy5307, line 166: const auto &info

src/meta/meta_bulk_load_ingestion_context.h
```diff
@@ -0,0 +1,92 @@
...
+
+    struct node_context
+    {
+        rpc_address address;
+        // node ingesting partition count
+        int32_t node_count;
+        // disk tag -> ingesting partition count
+        std::map<std::string, int32_t> disk_count;
+
+        node_context() : node_count(0) {}
```
levy5307, line 58: In my opinion, it's better to init all of the members in cstor function

src/meta/meta_bulk_load_ingestion_context.h
```diff
@@ -0,0 +1,92 @@
...
+
+        partition_node_info() {}
+        partition_node_info(const partition_configuration &config, const config_context &cc)
+        {
+            create(config, cc);
+        }
+        void create(const partition_configuration &config, const config_context &cc);
+    };
+
+    struct node_context
```
levy5307, line 50: Why did you use struct here?

src/meta/meta_bulk_load_ingestion_context.h
```diff
@@ -0,0 +1,92 @@
...
+
+        node_context() : node_count(0) {}
+        node_context(const rpc_address &address, const std::string &disk_tag)
+            : address(address), node_count(0)
+        {
+            init_disk(disk_tag);
+        }
+
+        void init_disk(const std::string &disk_tag);
+        int32_t get_max_disk_ingestion_count(const int32_t max_node_ingestion_count) const;
```
levy5307, line 66: uint32_t

src/meta/meta_bulk_load_ingestion_context.cpp
```diff
@@ -0,0 +1,204 @@
...
+    for (const auto &kv : info.node_disk) {
+        _nodes_context[kv.first].decrease(kv.second);
+    }
+    _running_partitions.erase(pid);
+}
+
+int32_t ingestion_context::get_app_running_count(const int32_t app_id) const
```
levy5307, line 173: uint32_t

src/meta/meta_bulk_load_ingestion_context.cpp
```diff
@@ -0,0 +1,204 @@
...
+DSN_TAG_VARIABLE(bulk_load_node_max_ingesting_count, FT_MUTABLE);
+
+DSN_DEFINE_uint32("meta_server", bulk_load_node_min_disk_count, 1, "min disk count of one node");
+DSN_TAG_VARIABLE(bulk_load_node_min_disk_count, FT_MUTABLE);
+
+ingestion_context::ingestion_context() { reset_all(); }
+
+ingestion_context::~ingestion_context() { reset_all(); }
+
+// ingestion_context::partition_node_info
```
levy5307, line 39: I think this annotation is useless

src/meta/meta_bulk_load_ingestion_context.h
```diff
@@ -0,0 +1,92 @@
...
+            create(config, cc);
+        }
+        void create(const partition_configuration &config, const config_context &cc);
+    };
+
+    struct node_context
+    {
+        rpc_address address;
+        // node ingesting partition count
+        int32_t node_count;
```
levy5307, line 54: uint32_t

src/meta/meta_bulk_load_ingestion_context.cpp
```diff
@@ -0,0 +1,204 @@
...
+    auto max_node_ingestion_count = FLAGS_bulk_load_node_max_ingesting_count;
+    if (node_count >= max_node_ingestion_count) {
+        dwarn_f("node[{}] has {} partition executing ingestion, max_count = {}",
+                address.to_string(),
+                node_count,
+                max_node_ingestion_count);
+        return false;
+    }
+
+    if (disk_count.find(disk_tag) == disk_count.end()) {
```
levy5307, line 95: I think these lines are useless

src/meta/meta_bulk_load_ingestion_context.h
levy5307, line 54: rename it to ingesting_partition_count

src/meta/meta_bulk_load_ingestion_context.cpp
hycdong: This sentence should not simplify as get the min value, when disk_count.size() < FLAGS_bulk_load_node_min_disk_count,  node_disk_count should be FLAGS_bulk_load_node_min_disk_count.

src/meta/meta_bulk_load_ingestion_context.h
```diff
@@ -0,0 +1,91 @@
...
+namespace dsn {
+namespace replication {
+
+DSN_DECLARE_uint32(bulk_load_node_max_ingesting_count);
+DSN_DECLARE_uint32(bulk_load_node_min_disk_count);
+
+class ingestion_context
+{
+public:
+    explicit ingestion_context();
```
Smityz, line 32:
```suggestion
    ingestion_context();
```
No need to use explicit here

src/meta/meta_bulk_load_ingestion_context.h
```diff
@@ -0,0 +1,91 @@
...
+        }
+        void create(const partition_configuration &config, const config_context &cc);
+    };
+
+    struct node_context
+    {
+        rpc_address address;
+        uint32_t node_ingesting_count;
+        // disk tag -> ingesting partition count
+        std::map<std::string, int32_t> disk_ingesting_counts;
```
Smityz, line 55: Why use `std::map` here? Other codes in this PR are using `std::unordered_map`

levy5307 (dismissed):

src/meta/meta_bulk_load_ingestion_context.cpp
```diff
@@ -0,0 +1,195 @@
...
+        return count;
+    });
+
+    const auto node_disk_count = disk_ingesting_counts.size() > FLAGS_bulk_load_node_min_disk_count
+                                     ? disk_ingesting_counts.size()
+                                     : FLAGS_bulk_load_node_min_disk_count;
+    return (max_node_ingestion_count + node_disk_count - 1) / node_disk_count;
+}
+
+bool ingestion_context::node_context::check_if_add(const std::string &disk_tag)
```
Smityz, line 79:
```suggestion
bool ingestion_context::node_context::check_if_add(const std::string &disk_tag) const
```

src/meta/meta_bulk_load_ingestion_context.cpp
```diff
@@ -0,0 +1,195 @@
...
+{
+    auto max_node_ingestion_count = FLAGS_bulk_load_node_max_ingesting_count;
```
Smityz, line 81: Why not use `FLAGS_bulk_load_node_max_ingesting_count` directly?

src/meta/meta_bulk_load_ingestion_context.h
```diff
@@ -0,0 +1,91 @@
...
+        bool check_if_add(const std::string &disk_tag);
+        void add(const std::string &disk_tag);
+        void decrease(const std::string &disk_tag);
+    };
+
+    bool try_partition_ingestion(const partition_configuration &config, const config_context &cc);
+    bool check_node_ingestion(const rpc_address &node, const std::string &disk_tag);
+    void add_partition(const partition_node_info &info);
+    void remove_partition(const gpid &pid);
+    uint32_t get_app_running_count(const uint32_t app_id) const;
```
Smityz, line 75:
```suggestion
    uint32_t get_ingesting_partition_count(const uint32_t app_id) const;
```
Maybe it's easier to understand

src/meta/meta_bulk_load_ingestion_context.h
```diff
@@ -0,0 +1,91 @@
...
+#include "meta_service.h"
+#include "server_state.h"
+
+namespace dsn {
+namespace replication {
+
+DSN_DECLARE_uint32(bulk_load_node_max_ingesting_count);
+DSN_DECLARE_uint32(bulk_load_node_min_disk_count);
+
+class ingestion_context
```
Smityz, line 29: Can you add some description about this class?

src/meta/meta_bulk_load_ingestion_context.cpp
```diff
@@ -0,0 +1,195 @@
...
+    uint32_t running_count = 0;
+    for (const auto &kv : _running_partitions) {
+        if (kv.first.get_app_id() == app_id) {
+            running_count++;
+        }
+    }
+    return running_count;
+}
+
+void ingestion_context::reset_app(const uint32_t app_id)
```
Smityz, line 175: `remove_partition` won't do everything if one `gpid` has been removed from `_running_partitions`
Maybe no need to duplicate `gpid` here

Smityz (dismissed):

src/meta/meta_bulk_load_ingestion_context.cpp
hycdong: This function use disk_ingesting_counts[disk_tag] simplifier, it can't be const.

src/meta/meta_bulk_load_ingestion_context.cpp
hycdong: max_node_ingestion_count will pass to another function, FLAGS_bulk_load_node_max_ingesting_count is a dynamic value. It is safety to use same value in check logic.

Smityz: ok~

src/meta/meta_bulk_load_ingestion_context.h
hycdong: I think all class should be define as `explicit`, why no need to use here?

src/meta/meta_bulk_load_ingestion_context.h
hycdong: I update it as `get_app_ingesting_count`.

src/meta/meta_bulk_load_ingestion_context.h
Smityz: There is no chance of implicit conversions

Smityz (approved):

PR #1037, opened 2022-01-27
fix: remove min_live_node_count_for_unfreeze from meta_options to ensure it can be correctly updated dynamically

empiredan:
https://github.com/apache/incubator-pegasus/issues/896

### DIFF
--- a/src/meta/meta_service.cpp
+++ b/src/meta/meta_service.cpp
@@ -58,7 +58,7 @@ DSN_DEFINE_uint64("meta_server",
                   "minimum live node count without which the state is freezed");
 DSN_TAG_VARIABLE(min_live_node_count_for_unfreeze, FT_MUTABLE);
 DSN_DEFINE_validator(min_live_node_count_for_unfreeze,
-                     [](uint64_t min_live_node_count) -> bool { return min_live_node_count >= 0; });
+                     [](uint64_t min_live_node_count) -> bool { return min_live_node_count > 0; });
 
 meta_service::meta_service()
     : serverlet("meta_service"), _failure_detector(nullptr), _started(false), _recovering(false)
--- a/src/meta/meta_service.cpp
+++ b/src/meta/meta_service.cpp
@@ -52,6 +52,14 @@
 namespace dsn {
 namespace replication {
 
+DSN_DEFINE_uint64("meta_server",
+                  min_live_node_count_for_unfreeze,
+                  3,
+                  "minimum live node count without which the state is freezed");
+DSN_TAG_VARIABLE(min_live_node_count_for_unfreeze, FT_MUTABLE);
+DSN_DEFINE_validator(min_live_node_count_for_unfreeze,
+                     [](uint64_t min_live_node_count) -> bool { return min_live_node_count >= 0; });
+
 meta_service::meta_service()
     : serverlet("meta_service"), _failure_detector(nullptr), _started(false), _recovering(false)
 {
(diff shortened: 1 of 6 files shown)


src/meta/meta_service.cpp
```diff
@@ -52,6 +52,14 @@
 namespace dsn {
 namespace replication {
 
+DSN_DEFINE_uint64("meta_server",
+                  min_live_node_count_for_unfreeze,
+                  3,
+                  "minimum live node count without which the state is freezed");
+DSN_TAG_VARIABLE(min_live_node_count_for_unfreeze, FT_MUTABLE);
+DSN_DEFINE_validator(min_live_node_count_for_unfreeze,
+                     [](uint64_t min_live_node_count) -> bool { return min_live_node_count >= 0; });
```
levy5307, line 61: why did you change the validator from `>` to `>=`?
And the type of min_live_node_count_for_unfreeze is uint64_t, so if there is no need to add a validator for `>=`

src/meta/test/meta_test_base.cpp
```diff
@@ -88,7 +88,8 @@ void meta_test_base::wait_all() { _ms->tracker()->wait_outstanding_tasks(); }
 
 void meta_test_base::set_min_live_node_count_for_unfreeze(uint64_t node_count)
 {
-    _ms->_meta_opts.min_live_node_count_for_unfreeze = node_count;
+    auto res = update_flag("min_live_node_count_for_unfreeze", std::to_string(node_count));
```
levy5307, line 91: ==> `FLAGS_min_live_node_count_for_unfreeze = true;`

src/meta/test/update_configuration_test.cpp
```diff
@@ -406,7 +406,8 @@ void meta_service_test_app::apply_balancer_test()
 void meta_service_test_app::cannot_run_balancer_test()
 {
     std::shared_ptr<null_meta_service> svc(new null_meta_service());
-    svc->_meta_opts.min_live_node_count_for_unfreeze = 0;
+    auto res = update_flag("min_live_node_count_for_unfreeze", "0");
```
levy5307, line 409: same

levy5307:
Thanks for you contribution!!! It looks great to me :)

src/meta/test/meta_test_base.cpp
empiredan: Means `FLAGS_min_live_node_count_for_unfreeze = node_count` ?

src/meta/meta_service.cpp
empiredan: Actually in the unit test it is set 0:
```
svc->_meta_opts.min_live_node_count_for_unfreeze = 0;
```

Since it was previously set as a property of `meta_options`, zero is allowed; however, once it's not allowed zero for `FLAGS_min_live_node_count_for_unfreeze`, its validation will fail. So I changed it to ">=". Can we set it directly by `FLAGS_min_live_node_count_for_unfreeze  = 0` around the validator with it still kept as ">"?

I reserved the validator of `min_live_node_count_for_unfreeze`  for the possible update future

src/meta/test/meta_test_base.cpp
levy5307: Yes. Sorry about that mistake

src/meta/meta_service.cpp
levy5307: I think it's ok to do like you said

hycdong (approved):

levy5307 (approved):

PR #1038, opened 2022-01-27
feat(dup_enhancement#1): add new dup status to support new dup  

Shuo-Jia:
# Ref-Issue

# Change

This PR refactor `DS_START` to three stage: `DS_PREPARE`, `DS_APP`, `DS_LOG`, and update the status transfer logic

### DIFF
--- a/src/meta/duplication/meta_duplication_service.cpp
+++ b/src/meta/duplication/meta_duplication_service.cpp
@@ -79,18 +79,6 @@ void meta_duplication_service::modify_duplication(duplication_modify_rpc rpc)
 
     duplication_info_s_ptr dup = it->second;
     auto to_status = request.__isset.status ? request.status : dup->status();
-    if (to_status == duplication_status::DS_PAUSE && dup->status() != duplication_status::DS_LOG) {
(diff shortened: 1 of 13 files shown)


src/meta/duplication/meta_duplication_service.cpp
```diff
@@ -183,10 +183,6 @@ void meta_duplication_service::do_add_duplication(std::shared_ptr<app_state> &ap
                                                   duplication_add_rpc &rpc)
 {
     dup->start();
-    if (rpc.request().freezed) {
-        dup->persist_status();
-        dup->alter_status(duplication_status::DS_PAUSE);
-    }
```
hycdong, line 189: What's those code usage? Why would you delete them?

Shuo-Jia: In the past, we usually freeze(DS_PAUSE) dup before uploading checkpoint completed. Now we use new status to `auto-control` it, which will keep DS_PREPARE before checkpoint

hycdong (approved):

levy5307 (approved):

PR #1045, opened 2022-02-07
feat(dup_enhancement#3): update load_meta_servers to avoid crush when duplication config is wrong

Shuo-Jia:
# Ref-Issue

# Change
replica_helper::load_meta_servers will return false if config is wrong

- [ ] some usage of https://github.com/apache/incubator-pegasus will need be update

### DIFF
--- a/include/dsn/dist/replication/replication_other_types.h
+++ b/include/dsn/dist/replication/replication_other_types.h
@@ -86,6 +86,7 @@ public:
     static bool get_replica_config(const partition_configuration &partition_config,
                                    ::dsn::rpc_address node,
                                    /*out*/ replica_configuration &replica_config);
+    // true if meta_list's value of config is valid, otherwise return false
     static bool load_meta_servers(/*out*/ std::vector<dsn::rpc_address> &servers,
                                   const char *section = "meta_server",
                                   const char *key = "server_list");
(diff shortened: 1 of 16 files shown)


src/common/replication_common.cpp
```diff
@@ -463,12 +463,16 @@ void replica_helper::load_meta_servers(/*out*/ std::vector<dsn::rpc_address> &se
     for (auto &s : lv) {
         ::dsn::rpc_address addr;
         if (!addr.from_string_ipv4(s.c_str())) {
-            dassert(
-                false, "invalid address '%s' specified in config [%s].%s", s.c_str(), section, key);
+            derror_f("invalid address '{}' specified in config [{}].{}", s.c_str(), section, key);
```
levy5307, line 466: s.c_str() ==> s

include/dsn/dist/replication/replication_other_types.h
```diff
@@ -86,7 +86,7 @@ class replica_helper
     static bool get_replica_config(const partition_configuration &partition_config,
                                    ::dsn::rpc_address node,
                                    /*out*/ replica_configuration &replica_config);
-    static void load_meta_servers(/*out*/ std::vector<dsn::rpc_address> &servers,
+    static bool load_meta_servers(/*out*/ std::vector<dsn::rpc_address> &servers,
```
levy5307, line 89: add some explanation about the return value

levy5307 (approved):

hycdong (approved):

PR #1046, opened 2022-02-07
feat(dup_enhancement#4): implement `create_follower_app_for_duplication` function

Shuo-Jia:
# Ref-Issue

# Change 
This pr support `create_follower_app_for_duplication` which will send `create table` reqeust with `duplication_option`


src/meta/duplication/meta_duplication_service.cpp
```diff
@@ -308,11 +310,66 @@ void meta_duplication_service::duplication_sync(duplication_sync_rpc rpc)
...
+                                 fmt::format("{}[{}]({})",
+                                             get_current_cluster_name(),
+                                             _meta_svc->get_meta_list_string(),
+                                             app->app_name));
+
+    rpc_address meta_servers;
+    meta_servers.assign_group(dup->follower_cluster_name.c_str());
+    meta_servers.group_address()->add_list(dup->follower_cluster_metas);
+
+    dsn::message_ex *msg = dsn::message_ex::create_request(RPC_CM_CREATE_APP);
```
levy5307, line 341: Is memory leak here for msg?

src/dsn.layer2.thrift
```diff
@@ -45,6 +45,12 @@ enum app_status
     AS_RECALLING
 }
 
+struct duplication_options {
+    1:string app_name;
```
hycdong, line 49: Is it neccessary to have `app_name` in `duplication_options`, it seems to be duplicated.

src/meta/duplication/meta_duplication_service.cpp
```diff
@@ -308,11 +310,66 @@ void meta_duplication_service::duplication_sync(duplication_sync_rpc rpc)
     }
 }
 
-// todo(jiashuo1) wait detail implementation
 void meta_duplication_service::trigger_follower_duplicate_checkpoint(
```
hycdong, line 313: It seems that function `trigger_follower_duplicate_checkpoint` will send a create table request to follower cluster meta server. I have following question about it:
1. Is this function half-implementation? I don't figure out any trigger follower checkpoint logic in this pull request, will it be added in your further pull request? If not, I suggesting that rename this function.
2. Why create follower table in duplication process, not before the duplication process?

src/meta/duplication/meta_duplication_service.cpp
```diff
@@ -308,11 +310,66 @@ void meta_duplication_service::duplication_sync(duplication_sync_rpc rpc)
...
+    request.options.envs = app->envs;
+    request.options.is_stateful = app->is_stateful;
+
+    duplication_options opts;
+    opts.app_name = app->app_name;
+    opts.cluster_name = get_current_cluster_name();
+    opts.metas = _meta_svc->_opts.meta_servers;
+    request.options.__set_duplication(opts);
+
+    request.options.envs.emplace(duplication_constants::kDuplicationMasterFlag,
```
hycdong, line 331: It seems that follower table will store master meta address list in app_envs, please adding comments for it.

src/dsn.layer2.thrift
```diff
@@ -75,4 +81,5 @@ struct app_info
     // New fields for bulk load
     // Whether this app is executing bulk load
     14:optional bool    is_bulk_loading = false;
+    15:optional duplication_options dup_options;
```
hycdong, line 84: According to the current `trigger_follower_duplicate_checkpoint` implementation. Follower table will store master meta address in app_envs. `duplication_options` store master cluster name and master meta address. Is `duplication_options` necessary? I think you can adding those informations once.

src/meta/test/meta_duplication_service_test.cpp
```diff
@@ -741,5 +753,51 @@ TEST_F(meta_duplication_service_test, fail_mode)
     ASSERT_EQ(dup->fail_mode(), duplication_fail_mode::FAIL_SKIP);
 }
 
+TEST_F(meta_duplication_service_test, trigger_follower_duplicate_checkpoint)
+{
+    struct TestCase
```
levy5307, line 758: TestCase ==> test_case

src/meta/duplication/meta_duplication_service.cpp
Shuo-Jia: No, rpc task will release it

src/meta/duplication/meta_duplication_service.cpp
Shuo-Jia: 1. current side(master cluster) just send trigger rpc, follower will execute some code based the rpc and duplication_option, which will in next follower side pr
2. duplication must start use a `empty` table, so force require create new table

src/dsn.layer2.thrift
Shuo-Jia: `app_envs` is string format which has a little complex to convert vector<rpc_dress>, and the `env `actually just used for show flag info

src/meta/duplication/meta_duplication_service.cpp
hycdong: OK, I suggest that add a new function called `create_follower_table` or other similar name to implement sending create app request.

src/dsn.layer2.thrift
hycdong: I have difference opinion with you. Structure `app_info` will store in remote storage(zk) and replica server disk, I think we shouldn't add too many fileds in this structure, especially for duplicated structure. 
I still suggest that you can remain only one structure. If the env is only used for showing flag info, it's better to showing the duplication_options but not adding it into envs.

src/meta/duplication/meta_duplication_service.cpp
Shuo-Jia: This function just has `one ability`: create_follower_table, so I can rename it `create_follower_table` but add another

src/dsn.layer2.thrift
Shuo-Jia: `duplication_options` is optional, it will not storage into zk and it also no need.

Add new structure will make code more simple, from `127.0.0.1:34801` string to `rpc_address`(I ever try, we first split , then convert string to int, and then create `rpc_address`) is complex, we don't have to add these redundant code.

hycdong: OK. If `duplication_options will not storage into zk and it also no need`, why would you add it into `app_info`? How about adding `duplication_options` in `create_app_request`?

Shuo-Jia: `create_app_request` has existed variable `create_options`, `duplication_options` is a subset, so I put directly in it

Shuo-Jia: And, I see， follower will need the info via `app_info`, if put `create_app_request`, I have to create new variable in `meta` and sync to `replica` server, it's very complex

hycdong: Your answer doesn't persuade me. 
`app_info` is used to store app metadata in remote storage and transfer it to replica server. As `duplication_options` is only used in rpc, why not adding it into rpc request?

Shuo-Jia: Ok，I change it into `create_app_request`

include/dsn/cpp/serialization_helper/dsn.layer2_types.h
```diff
@@ -276,6 +278,58 @@ inline std::ostream &operator<<(std::ostream &out, const configuration_query_by_
     return out;
 }
 
+typedef struct _duplication_options__isset
```
hycdong, line 281: As your new commit doesn't update `dsn.layer2.thrift` files, please discard the update of `dsn.layer2_types.h` and `dsn.layer2_types.cpp`

Shuo-Jia: done

hycdong (approved):

src/meta/test/meta_duplication_service_test.cpp
```diff
@@ -123,6 +124,17 @@ class meta_duplication_service_test : public meta_test_base
         wait_all();
     }
 
+    void trigger_follower_duplicate_checkpoint(const std::shared_ptr<duplication_info> &dup,
+                                               const std::shared_ptr<app_state> &app)
+    {
+        dup_svc().create_follower_app_for_duplication(dup, app);
+    }
+
+    duplication_status::type next_status(const std::shared_ptr<duplication_info> &dup)
```
levy5307, line 133: duplication_status::type next_status(const std::shared_ptr<duplication_info> &dup) const

src/meta/meta_service.h
```diff
@@ -165,6 +165,15 @@ class meta_service : public serverlet<meta_service>
     void unlock_meta_op_status();
     meta_op_status get_op_status() const { return _meta_op_status.load(); }
 
+    std::string get_meta_list_string()
```
levy5307, line 168: std::string get_meta_list_string() const

levy5307 (approved):

hycdong (approved):

PR #1047, opened 2022-02-08
fix: fail_point is failed when use `FAIL_POINT_INJECT_VOID_F`

Shuo-Jia:
# Issue-Ref
https://github.com/apache/incubator-pegasus/issues/903

# Change
https://github.com/XiaoMi/rdsn/pull/818 add `FAIL_POINT_INJECT_VOID_F` to support some no-need return `point`, but use `off()` is error, which will casue any`point` will be executed, though it not be register

### DIFF
--- a/src/utils/fail_point_impl.h
+++ b/src/utils/fail_point_impl.h
@@ -45,15 +45,13 @@ struct fail_point
     enum task_type
     {
         // `action` contain `off()`, which would `close` the fail_point whose `function` passed will
-        // not
-        // be executed;
+        // not be executed;
         Off,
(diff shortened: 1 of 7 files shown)


src/utils/fail_point_impl.h
```diff
@@ -47,6 +47,7 @@ struct fail_point
         Off,
         Return,
         Print,
+        Void,
```
hycdong, line 50: How about adding some comments for the fail_point::task_type to describe the difference between them and how to use it?

Shuo-Jia: done

hycdong (approved):

PR #1048, opened 2022-02-09
feat: make option empty_write_disable mutable

levy5307:
https://github.com/apache/incubator-pegasus/issues/859

Empty write is used for flushing WAL log entry which is submit asynchronously. Make sure it can work well if you diable it.

### DIFF
--- a/src/common/replication_common.cpp
+++ b/src/common/replication_common.cpp
@@ -41,6 +41,10 @@ DSN_DEFINE_int32("replication",
                  5,
                  "concurrent bulk load downloading replica count");
 
+/**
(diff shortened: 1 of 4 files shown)


Shuo-Jia (approved):

hycdong:
I suggest that adding more in your description to describe the affect of removing empty_write.

levy5307:
> I suggest that you can more in your description to describe the affect of removing empty_write.

Good idea. And I add a annotation above the definition of empty_write_disable.

hycdong (approved):

Shuo-Jia (approved):

PR #1050, opened 2022-02-10
refactor: drop redundant config_status::pending_proposal

empiredan:
https://github.com/apache/incubator-pegasus/issues/908

### DIFF
--- a/src/meta/meta_data.h
+++ b/src/meta/meta_data.h
@@ -57,14 +57,12 @@ namespace replication {
 enum class config_status
 {
     not_pending,
-    pending_proposal,
     pending_remote_sync,
     invalid_status
 };
 
 ENUM_BEGIN(config_status, config_status::invalid_status)
 ENUM_REG(config_status::not_pending)
-ENUM_REG(config_status::pending_proposal)
 ENUM_REG(config_status::pending_remote_sync)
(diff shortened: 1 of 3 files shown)


acelyc111:
Plz check whether `config_status` is appeared in RPC or disk storage, remove this state may cause incompatible problem.

acelyc111:
I think mark an enum value as deprecated and remove all usage would be better.

levy5307:
> Plz check whether `config_status` is appeared in RPC or disk storage, remove this state may cause incompatible problem.

+1

hycdong:
> I think mark an enum value as deprecated and remove all usage would be better.

+1

empiredan:
> > I think mark an enum value as deprecated and remove all usage would be better.
> 
> +1

Actually since pegasus v1.8 (or even older version) `pending_proposal` has not yet been used to assign any variable. It's only used to compare with some read operations rather than write to some variable. On the other hand, the only usage of it is to assign to `config_context::stage`, a variable in memory to record current state of the configuration of a partition which doesn't need to synced to disk.

However, in case of unexpected I think we'd better to mark it deprecated and remove all usages, without dropping the definitions.

levy5307 (approved):

acelyc111 (approved):

PR #1052, opened 2022-02-11
fix: shutdown error while closing replica

hycdong:
https://github.com/apache/incubator-pegasus/issues/906
This pull request should be reviewed with https://github.com/apache/incubator-pegasus/pull/909 together.

### DIFF
--- a/src/replica/replica.cpp
+++ b/src/replica/replica.cpp
@@ -434,10 +434,6 @@ void replica::close()
         _checkpoint_timer = nullptr;
     }
 
-    if (_app != nullptr) {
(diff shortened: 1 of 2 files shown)


src/replica/replica.cpp
```diff
@@ -434,10 +434,6 @@ void replica::close()
         _checkpoint_timer = nullptr;
     }
 
-    if (_app != nullptr) {
```
levy5307, line 437: It makes me confused that why did you remove these lines?

levy5307 (approved):

Shuo-Jia (approved):

PR #1056, opened 2022-02-17
feat(dup_enhancement#9): master replica support `on_query_last_checkpoint_info` rpc for follower replica

Shuo-Jia:
# Ref-Issue

# Change
When follower start copy checkpoint, it need get checkpoint info firstly, so here add the rpc api support it


src/replica/replica_chkpt.cpp
```diff
@@ -191,46 +198,29 @@ void replica::init_checkpoint(bool is_emergency)
         _stub->_counter_recent_trigger_emergency_checkpoint_count->increment();
 }
 
-// @ secondary
-void replica::on_copy_checkpoint(const replica_configuration &request,
```
hycdong, line 195: I notice that you update the error code of this function, will it affect previous learn process?

Shuo-Jia: it's not used in the past

src/replica/replica_stub.h
```diff
@@ -135,7 +135,7 @@ class replica_stub : public serverlet<replica_stub>, public ref_counter
     void on_add_learner(const group_check_request &request);
     void on_remove(const replica_configuration &request);
     void on_group_check(group_check_rpc rpc);
-    void on_copy_checkpoint(copy_checkpoint_rpc rpc);
+    void on_query_last_checkpoint(query_last_checkpoint_info_rpc rpc);
```
hycdong, line 138: Please add some comments to should it is used for duplication, same as the function definition in `replica.h`.

src/replica/replica_chkpt.cpp
```diff
@@ -191,46 +198,29 @@ void replica::init_checkpoint(bool is_emergency)
...
-    }
-
     if (_app->last_durable_decree() == 0) {
-        response.err = ERR_OBJECT_NOT_FOUND;
+        response.err = ERR_PATH_NOT_FOUND;
         return;
     }
 
     blob placeholder;
     int err = _app->get_checkpoint(0, placeholder, response.state);
```
hycdong, line 212: I suggest you can refactor function `get_checkpoint` in your further pull request, it seems only the last parameter useful, you can delete the first two parameters of this function.

src/replica/replica_chkpt.cpp
```diff
@@ -191,46 +198,29 @@ void replica::init_checkpoint(bool is_emergency)
...
-        for (std::string &file_name : response.state.files) {
-            std::size_t last_splitter = file_name.find_last_of("/\\");
-            if (last_splitter != std::string::npos)
-                file_name = file_name.substr(last_splitter + 1);
-        }
+        response.base_local_dir = utils::filesystem::path_combine(
+            _app->data_dir(), checkpoint_folder(response.state.to_decree_included));
         response.address = _stub->_primary_address;
+        for (auto &file : response.state.files) {
+            file = file.substr(response.base_local_dir.length() + 1);
```
hycdong, line 222: I am a little bit confused about the usage of `response.base_local_dir` and this `substr` sentence for `response.state.files`. Could you please explain it to me and add comments in code?

src/replica/replica_chkpt.cpp
Shuo-Jia: get_checkpoint is `pegasus` implementation

hycdong: Yes, I have read the code of `get_checkpoint`, the first two parameters of this function is useless, you can refactor it in future.

src/replica/replica_chkpt.cpp
Shuo-Jia: it's same as `learn` code: 
https://github.com/XiaoMi/rdsn/blob/master/src/replica/replica_learn.cpp#L543

The peer need the `file path` list to copy

Shuo-Jia: comments done.

src/replica/replica_chkpt.cpp
```diff
@@ -191,46 +198,32 @@ void replica::init_checkpoint(bool is_emergency)
...
-        for (std::string &file_name : response.state.files) {
-            std::size_t last_splitter = file_name.find_last_of("/\\");
-            if (last_splitter != std::string::npos)
-                file_name = file_name.substr(last_splitter + 1);
-        }
+        // for example: base_local_dir = "./data" + "checkpoint.1024" = "./data/checkpoint.1024"
+        response.base_local_dir = utils::filesystem::path_combine(
+            _app->data_dir(), checkpoint_folder(response.state.to_decree_included));
         response.address = _stub->_primary_address;
+        for (auto &file : response.state.files) {
```
levy5307, line 222: `for (const auto &file : response.state.files)`

levy5307 (approved):

PR #1059, opened 2022-02-22
fix(dup_enhancement#10): follower meta need return `ERR_APP_EXIST` when receive create same app

Shuo-Jia:
# Ref-Issue


src/meta/server_state.cpp
```diff
@@ -1089,11 +1089,20 @@ void server_state::create_app(dsn::message_ex *msg)
     bool will_create_app = false;
     dsn::unmarshall(msg, request);
 
-    ddebug("create app request, name(%s), type(%s), partition_count(%d), replica_count(%d)",
-           request.app_name.c_str(),
-           request.options.app_type.c_str(),
-           request.options.partition_count,
-           request.options.replica_count);
+    auto env_iterator =
```
hycdong, line 1092: env_iterator -> duplication_env_iterator

levy5307, line 1092: const auto &duplication_env_iterator

levy5307 (approved):

hycdong (approved):

PR #1060, opened 2022-02-23
feat(dup_enhancement#11): follower replica add `replica_follower` to support duplicate checkpoint when open replica

Shuo-Jia:
# Ref-Issue

# Change
This pr add `replica_follower.cpp` to support duplicate checkpoint when open new replica


src/replica/duplication/replica_follower.cpp
```diff
@@ -0,0 +1,66 @@
...
+        need_duplicate = true;
+
+        _master_cluster_name = envs.at(duplication_constants::kDuplicationEnvMasterClusterKey);
+        _master_app_name = _replica->get_app_info()->app_name;
+
+        const auto &meta_list_str = envs.at(duplication_constants::kDuplicationEnvMasterMetasKey);
+        std::vector<std::string> metas;
+        boost::split(metas, meta_list_str, boost::is_any_of(","));
+        dassert_f(!metas.empty(), "master cluster meta list is invalid!");
+        for (const auto &m : metas) {
```
levy5307, line 54: for (const auto &meta : metas) {

src/replica/replica_init.cpp
```diff
@@ -91,6 +94,21 @@ error_code replica::initialize_on_new()
...
+        (err = rep->get_replica_follower()->duplicate_checkpoint()) != dsn::ERR_OK) {
+        derror_f("{} try to duplicate replica checkpoint failed, error({})",
+                 rep->name(),
+                 err.to_string());
+        rep->close();
+        delete rep;
+        rep = nullptr;
+
+        // clear work on failure
+        utils::filesystem::remove_path(dir);
```
levy5307, line 107:
```
void clear_on_failure() {
        utils::filesystem::remove_path(dir);
        stub->_fs_manager.remove_replica(gpid);
}
```
You can do like this.

src/replica/replica_stub.cpp
```diff
@@ -1923,10 +1923,11 @@ void replica_stub::on_disk_stat()
     ddebug("finish to update disk stat, time_used_ns = %" PRIu64, dsn_now_ns() - start);
 }
 
-::dsn::task_ptr replica_stub::begin_open_replica(const app_info &app,
-                                                 gpid id,
-                                                 std::shared_ptr<group_check_request> req,
-                                                 std::shared_ptr<configuration_update_request> req2)
+::dsn::task_ptr replica_stub::begin_open_replica(
```
levy5307, line 1926: remove ::

src/replica/replica.h
```diff
@@ -113,6 +114,7 @@ class replica : public serverlet<replica>, public ref_counter, public replica_ba
                          gpid gpid,
                          const app_info &app,
                          bool restore_if_necessary,
+                         bool duplicate_if_necessary,
```
hycdong, line 117: `duplicate_if_necessary = true` means it is duplication follower, and `duplicate_if_necessary = false` means it is duplication master or not in duplication. How about rename it as `is_duplication_follower`?

src/replica/replica.cpp
```diff
@@ -52,8 +53,12 @@ namespace replication {
...
 const std::string replica::kAppInfo = ".app-info";
 
-replica::replica(
-    replica_stub *stub, gpid gpid, const app_info &app, const char *dir, bool need_restore)
+replica::replica(replica_stub *stub,
+                 gpid gpid,
+                 const app_info &app,
+                 const char *dir,
+                 bool need_restore,
+                 bool need_duplicate)
```
hycdong, line 61: Same as `duplicate_if_necessary`, it seems duplication follower and master should also set `need_duplicate` as true.

src/replica/duplication/replica_follower.cpp
```diff
@@ -0,0 +1,66 @@
...
+    init_master_info();
+}
+
+replica_follower::~replica_follower() { _tracker.wait_outstanding_tasks(); }
+
+void replica_follower::init_master_info()
+{
+    const auto &envs = _replica->get_app_info()->envs;
+
+    if (envs.find(duplication_constants::kDuplicationEnvMasterClusterKey) != envs.end() &&
```
hycdong, line 43: How about using:
```
if(envs.find(key1) == envs.end() || envs.find(key2) == envs.end()){
    return;
}
// original code
```

src/replica/duplication/replica_follower.cpp
```diff
@@ -0,0 +1,66 @@
...
+        envs.find(duplication_constants::kDuplicationEnvMasterMetasKey) != envs.end()) {
+        need_duplicate = true;
+
+        _master_cluster_name = envs.at(duplication_constants::kDuplicationEnvMasterClusterKey);
+        _master_app_name = _replica->get_app_info()->app_name;
+
+        const auto &meta_list_str = envs.at(duplication_constants::kDuplicationEnvMasterMetasKey);
+        std::vector<std::string> metas;
+        boost::split(metas, meta_list_str, boost::is_any_of(","));
+        dassert_f(!metas.empty(), "master cluster meta list is invalid!");
```
hycdong, line 53: I think it's better to ignore invalid address, print error logs rather than assert.

src/replica/replica.cpp
```diff
@@ -421,6 +429,7 @@ void replica::close()
 {
     dassert_replica(status() == partition_status::PS_ERROR ||
                         status() == partition_status::PS_INACTIVE ||
+                        _disk_migrator->status() == disk_migration_status::IDLE ||
```
hycdong, line 432: Will duplication follower affect disk migrator status?

src/replica/duplication/replica_follower.cpp
Shuo-Jia: This code will be executed when inited but not running, if config is error, we shouldn't let open succeed. just error won't let us find the error

src/replica/replica.cpp
Shuo-Jia: Here is a bug and can't find in previous, added unit test won't pass when test `close`

hycdong (approved):

levy5307 (approved):

PR #1063, opened 2022-02-28
feat(dup_enhancement#12): follower replica add framework and implement the main function

Shuo-Jia:
# Ref-Issue

# Change
- This pr add `replica_follower.cpp` framework
- support `tracker` mark `if success`


include/dsn/tool-api/task_tracker.h
```diff
@@ -166,11 +167,17 @@ class task_tracker
     // return not finished task count
     int cancel_but_not_wait_outstanding_tasks();
 
+    // todo(jiashuo1) it's not safe
```
levy5307, line 170: What does "it's not safe" mean?

Shuo-Jia: thread-safe, will update

src/replica/duplication/replica_follower.cpp
```diff
@@ -61,8 +62,86 @@ void replica_follower::init_master_info()
...
+    dsn::message_ex *msg = dsn::message_ex::create_request(
+        RPC_CM_QUERY_PARTITION_CONFIG_BY_INDEX, 0, get_gpid().thread_hash());
+    dsn::marshall(msg, meta_config_request);
+    rpc::call(meta_servers,
+              msg,
+              &_tracker,
+              [&](error_code err, configuration_query_by_index_response &&resp) mutable {
+                  tasking::enqueue(LPC_DUPLICATE_CHECKPOINT, &_tracker, [=]() mutable {
+                      FAIL_POINT_INJECT_F("duplicate_checkpoint_ok", [&](string_view s) -> void {
+                          _tracker.set_result(ERR_OK);
```
hycdong, line 101: I consider it is strange to pass the copy-duplication-checkpoint result through tracker, especially in this pr, it is only used in unit test.
How about using a local variety, like _checkpoint_result?

Shuo-Jia: - it's not just used for `unit test`,  duplicate_checkpoint of [caller](https://github.com/XiaoMi/rdsn/blob/duplication_dev/src/replica/replica_init.cpp#L91) will use the result
- tracker of `rdsn`  has the shortcoming, when we want get some `async work` result, we had to define `local variable` to update the result in every time we use it, I don't think it's one good usage. define `result` in `tracker` can help us track and get async work result,  avoid define one local variable after another

hycdong: - My meaning is that the result seems will always be ERR_OK, will it be a error only in test case?
- If the tracker only has one task, it sounds a good solution. Actually, task_tacker has a task list, it is difficult for you to track which result belonging to which task. Besides, as my first comment, will the `result` be other error code rather than `ERR_OK`?

hycdong: As we discuss offline, I recommend using `bool` instead of `error_code`.

include/dsn/tool-api/task_tracker.h
```diff
@@ -166,11 +167,16 @@ class task_tracker
...
+    void set_failure() { _all_task_success = false; }
+
+    bool all_tasks_success() const { return _all_task_success; }
+
 private:
     friend class trackable_task;
     const int _task_bucket_count;
     ::dsn::utils::ex_lock_nr_spin *_outstanding_tasks_lock;
     dlink *_outstanding_tasks;
+    bool _all_task_success{true};
```
hycdong, line 179: It seems the default value of `_all_task_success` should be false, it will be true through function `set_succeed`.

Shuo-Jia: Oh, it's mistake..

Shuo-Jia: done

src/replica/duplication/replica_follower.h
```diff
@@ -41,15 +41,41 @@ class replica_follower : replica_base
...
+        if (_master_replica_config.primary != rpc_address::s_invalid_address) {
+            return fmt::format("{}({}|{})",
+                               app_info,
+                               _master_replica_config.primary.to_string(),
+                               _master_replica_config.pid.to_string());
+        }
+        return app_info;
+    }
+
+    void async_duplicate_checkpoint_from_master_replica();
```
hycdong, line 68: In your future pull request, I recommend you can add comments in your head files or add some duplication flow graphs.

hycdong (approved):

src/replica/duplication/replica_follower.h
Shuo-Jia: Good! however it has a little complex, add the link https://github.com/apache/incubator-pegasus/issues/892 is ok?

levy5307 (approved):

src/replica/duplication/replica_follower.h
hycdong: Okay, maybe you can finally add a rfc document.

PR #1064, opened 2022-03-01
feat(dup_enhancement#13): follower replica support query and update master partition config

Shuo-Jia:
# Ref-Issue

# Change

follower need query master replica config for sending nfs target


src/replica/duplication/replica_follower.cpp
```diff
@@ -104,27 +105,77 @@ void replica_follower::async_duplicate_checkpoint_from_master_replica()
...
-// todo(jiashuo1)
-void replica_follower::update_master_replica_config_callback(
-    error_code err, configuration_query_by_index_response &&resp)
+error_code
+replica_follower::update_master_replica_config(error_code err,
+                                               configuration_query_by_index_response &&resp)
 {
-    copy_master_replica_checkpoint(_master_replica_config.primary, _master_replica_config.pid);
+    error_code err_code = err != ERR_OK ? err : resp.err;
+    if (err_code != ERR_OK) {
```
levy5307, line 120: dsn_unlikely

src/replica/duplication/replica_follower.cpp
```diff
@@ -104,27 +105,77 @@ void replica_follower::async_duplicate_checkpoint_from_master_replica()
...
+                          copy_master_replica_checkpoint();
+                      }
                   });
               });
 }
 
-// todo(jiashuo1)
-void replica_follower::update_master_replica_config_callback(
-    error_code err, configuration_query_by_index_response &&resp)
+error_code
```
hycdong, line 115: I have a suggestion that you can add the thread pool those functions running on.

src/replica/duplication/replica_follower.cpp
```diff
@@ -104,27 +105,77 @@ void replica_follower::async_duplicate_checkpoint_from_master_replica()
...
+    _master_replica_config = resp.partitions[0];
+    return ERR_OK;
 }
 
-// todo(jiashuo1)
-void replica_follower::copy_master_replica_checkpoint(const rpc_address &node, const gpid &pid)
+void replica_follower::copy_master_replica_checkpoint()
 {
-    query_last_checkpoint_info_callback(ERR_OK, learn_response() /* todo(jiashuo1): placeholder */);
+    if (_master_replica_config.primary == rpc_address::s_invalid_address ||
```
hycdong, line 156: I think check if primary address of master replica is invalid can move into function `update_master_replica_config`.
When master replica is under reconfiguration, it primary address is invalid. Same as the partition index check.

src/replica/duplication/replica_follower.cpp
Shuo-Jia: done

hycdong (approved):

levy5307 (approved):

PR #1065, opened 2022-03-01
feat(dup_enhancement#14): follower replica support nfs copy checkpoint

Shuo-Jia:
# Ref-Issue

# Change

follower need query master replica config for sending nfs target


include/dsn/dist/replication/duplication_common.h
```diff
@@ -73,6 +73,7 @@ inline bool is_cluster_id_configured(uint8_t cid)
 // todo(jiashuo1) it will be filled more constants
```
hycdong, line 73: You can smiliar remove comments if this is the last pr of rdsn.

hycdong (approved):

PR #1068, opened 2022-03-03
refactor: make ctor/dtor function of singleton classes to be private

levy5307:
make ctor functions of singleton to be private, to avoid creating more than one object for singleton classes

https://github.com/apache/incubator-pegasus/issues/887

### DIFF
--- a/include/dsn/utility/singleton.h
+++ b/include/dsn/utility/singleton.h
@@ -17,18 +17,16 @@
 
 #pragma once
 
+#include <boost/noncopyable.hpp>
+
 namespace dsn {
 namespace utils {
 
 template <typename T>
-class singleton
+class singleton : private boost::noncopyable
 {
 public:
     singleton() = default;
-
-    // disallow copy and assign
-    singleton(const singleton &) = delete;
-    singleton &operator=(const singleton &) = delete;
 
     static T &instance()
     {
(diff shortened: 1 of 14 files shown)


hycdong (approved):

levy5307:
> Should we also make the deconstructor as private?

I think it doesn't make sense. No one would call the destructor explicitly, and what I want to do is to prevent create a sington outside of instance() function.

levy5307:
> Should we also make the deconstructor as private?

It looks great to me : )

include/dsn/utility/singleton.h
```diff
@@ -17,19 +17,17 @@
...
 
 template <typename T>
-class singleton
+class singleton : private boost::noncopyable
 {
 public:
     singleton() = default;
 
-    // disallow copy and assign
-    singleton(const singleton &) = delete;
```
levy5307, line 30: It is more clear to use `boost::noncopyable`

acelyc111 (approved):

Smityz:
Maybe we should add some comments in `include/dsn/utility/singleton.h` to tell others how to use it, like setting its deconstructor as private, and tell them why. It's more friendly for c++ beginners

levy5307:
> Maybe we should add some comments in `include/dsn/utility/singleton.h` to tell others how to use it, like setting its deconstructor as private, and tell them why. It's more friendly for c++ beginners

Done

Smityz (approved):

hycdong (approved):

PR #1069, opened 2022-03-09
feat(direct_io): download files from hdfs with direct I/O

GiantKing:
There are two stages when pegasus download files from HDFS:
1. Download file to memory. There is a speed limiter, it will get a file within several batchs.
2. Save file to disk. Writing the whole file into disk at once. 

To reduce disk io critical status in stage two, I add a new class that writing files with direct I/O. 

Issues: https://github.com/apache/incubator-pegasus/issues/926

# Config-Update
There are two GFLAGS:
```diff
+ hdfs_enable_direct_io = false //Whether to enable direct I/O when download files from hdfs, default false
+ direct_io_buffer_pages = 1 // Number of pages we need to set to direct io buffer, default 1, 64 is recommend in my test.

### CODE REVIEW

**Comment by Shuo-Jia** [comment]
No `issue link`, please add it like others

File: src/block_service/hdfs/hdfs_service.cpp [hunk 1]
```diff
@@ -55,6 +56,12 @@ DSN_DEFINE_uint64("replication",
                   "hdfs write batch size, the default value is 64MB");
 DSN_TAG_VARIABLE(hdfs_write_batch_size_bytes, FT_MUTABLE);
 
+DSN_DEFINE_bool("replication",
+                hdfs_enable_direct_io,
+                false,
+                "Whether to enable direct I/O when download files from hdfs");
+DSN_TAG_VARIABLE(hdfs_enable_direct_io, FT_MUTABLE);
```
Shuo-Jia, line 63: Actually, it's not `hdfs_direct_io`, you just only support saving hdfs file to local via direct_io.

You may consider that define general `enable_direct_io`  in `block_service.h` or `directio_writable_file.h`  and used for any remote filesystem to save file

src/block_service/directio_writable_file.h
```diff
@@ -0,0 +1,54 @@
...
+{
+public:
+    explicit direct_io_writable_file(const std::string &file_path);
+    ~direct_io_writable_file();
+
+    // can't copy or move
+    direct_io_writable_file(const direct_io_writable_file &) = delete;
+    direct_io_writable_file &operator=(const direct_io_writable_file &) = delete;
+    direct_io_writable_file(direct_io_writable_file &&) = delete;
+    direct_io_writable_file &operator=(direct_io_writable_file &&) = delete;
```
acelyc111, line 36: How about use `DISALLOW_COPY_AND_ASSIGN` instead to keep the same style?

src/block_service/directio_writable_file.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+direct_io_writable_file::direct_io_writable_file(const std::string &file_path)
+    : _fd(-1),
+      _file_size(0),
+      _buffer(),
+      _buffer_size(FLAGS_direct_io_buffer_pages * g_page_size),
+      _offset(0)
+{
+    if (posix_memalign(&_buffer, g_page_size, _buffer_size) != 0) {
+        derror_f("Allocate memaligned buffer failed, errno = {}", errno);
```
acelyc111, line 58: How will the server act when the ctor failed?

src/block_service/directio_writable_file.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+direct_io_writable_file::~direct_io_writable_file()
+{
+    if (_buffer == nullptr || _fd < 0) {
+        return;
+    }
+
+    if (_offset > 0) {
+        if (::write(_fd, _buffer, _buffer_size) != _buffer_size) {
+            derror_f("Failed to write last chunk, errno = {}", errno);
```
acelyc111, line 79: Will the data lost if just logging an error?

GiantKing: Yeah, move this logic to direct_io_writable_file::write is better.

src/block_service/directio_writable_file.cpp
GiantKing: OK, I add a init function.

src/block_service/directio_writable_file.h
GiantKing: Sure, tks

src/block_service/directio_writable_file.cpp
```diff
@@ -0,0 +1,116 @@
...
+#include <unistd.h> // getpagesize
+
+#include <dsn/dist/fmt_logging.h>
+#include <dsn/utility/flags.h>
+
+#include "block_service/directio_writable_file.h"
+
+DSN_DEFINE_uint32("replication",
+                  direct_io_buffer_pages,
+                  1,
```
Smityz, line 34: why not set 64 as a default value?

GiantKing: OK

src/block_service/directio_writable_file.cpp
GiantKing: I have to write last chunk in destructor, otherwise direct_io_writable_file::write can't be called repeatedly. Is there any ideas better?

acelyc111: We add some functions like `init()` to do the work as what you do in ctor now, and add `finalize()` to do the work as what you do in dector. Both `init()` and `finalize()` have a return value, you can check it in caller.

acelyc111: Maybe a 'state' is needed, we must call `init` before `write`, and must call `finalize` before destory it.

src/block_service/directio_writable_file.cpp
```diff
@@ -0,0 +1,137 @@
...
+        ftruncate(_fd, _file_size);
+    }
+    return true;
+}
+
+uint32_t direct_io_writable_file::write(const char *s, uint32_t n)
+{
+    if (!_buffer || _fd < 0) {
+        derror_f("initialize the instance first");
+        return -1;
```
acelyc111, line 112: `-1` is not an `uint32_t`.
Any more, I think use asserts would be better than defensive programing, all the code is controlled by programers, this case will not happend if we use `direct_io_writable_file` correctly, i.e. calling sequence is ordered in ctor() -> initialize() -> write() -> finalize() -> dector().

src/block_service/directio_writable_file.h
```diff
@@ -0,0 +1,52 @@
...
+    std::string _file_path;
+    int _fd;
+    uint32_t _file_size;
+
+    // page size aligned buffer
+    void *_buffer;
+    // buffer size
+    uint32_t _buffer_size;
+    // buffer offset
+    uint32_t _offset;
```
acelyc111, line 47:
```suggestion
    std::string _file_path;
    int _fd;
    uint32_t _file_size;

    // page size aligned buffer
    void *_buffer;
    // buffer size
    uint32_t _buffer_size;
    // buffer offset
    uint32_t _offset;

    DISALLOW_COPY_AND_ASSIGN(direct_io_writable_file);    
```

src/block_service/directio_writable_file.cpp
```diff
@@ -0,0 +1,137 @@
...
+{
+}
+
+direct_io_writable_file::~direct_io_writable_file()
+{
+    if (!_buffer || _fd < 0) {
+        return;
+    }
+    // Here is an ensurance, users shuold call finalize manually
+    if (_offset > 0 && !finalize()) {
```
acelyc111, line 66: Same, IMO, asserts would be better than defensive programing.
We can add a state machine, like: uninitialize -> initialized -> writable -> finalized.
Then, we can use `assert(_state == finalized)` here instead.

GiantKing: Assert is OK. But this is not a complex process, I think no need to add a state machine. _buffer and _fd are vailid stand for initialized, while _offset is 0 means finalize() has been called.

src/block_service/directio_writable_file.cpp
GiantKing: Thinks.

src/block_service/directio_writable_file.cpp
acelyc111: OK

acelyc111 (dismissed):
LGTM

Shuo-Jia (dismissed):

src/block_service/directio_writable_file.cpp
```diff
@@ -0,0 +1,132 @@
...
+DSN_TAG_VARIABLE(enable_direct_io, FT_MUTABLE);
+
+const uint32_t g_page_size = getpagesize();
+
+direct_io_writable_file::direct_io_writable_file(const std::string &file_path)
+    : _file_path(file_path),
+      _fd(-1),
+      _file_size(0),
+      _buffer(nullptr),
+      _buffer_size(FLAGS_direct_io_buffer_pages * g_page_size),
```
levy5307, line 55: It can't be updated dynamically if you init `_buffer_size` like this,

Shuo-Jia (approved):

PR #1070, opened 2022-03-09
feat(new_metrics): implement the metric entity & its prototype

empiredan:
https://github.com/apache/incubator-pegasus/issues/925

### DIFF
--- a/include/dsn/utility/metrics.h
+++ b/include/dsn/utility/metrics.h
@@ -21,11 +21,12 @@
 #include <unordered_map>
 
 #include <dsn/utility/autoref_ptr.h>
+#include <dsn/utility/ports.h>
 
 // A metric library (for details pls see https://github.com/apache/incubator-pegasus/issues/922)
 // inspired by Kudu metrics (https://github.com/apache/kudu/blob/master/src/kudu/util/metrics.h).
 //
-// 
+//
 // Example of defining and instantiating a metric entity
 // -----------------------------------------------------
 // Define an entity type at the top of your .cpp file (not within any namespace):
@@ -41,8 +42,7 @@
 #define METRIC_DEFINE_entity(name) ::dsn::metric_entity_prototype METRIC_ENTITY_##name(#name)
 
 // The following macros act as forward declarations for entity types and metric prototypes.
-#define METRIC_DECLARE_entity(name) \
-  extern ::dsn::metric_entity_prototype METRIC_ENTITY_##name
+#define METRIC_DECLARE_entity(name) extern ::dsn::metric_entity_prototype METRIC_ENTITY_##name
 
 namespace dsn {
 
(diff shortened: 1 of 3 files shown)


src/utils/metrics.cpp
```diff
@@ -0,0 +1,56 @@
...
+    ref_ptr<metric_entity> entity(new metric_entity(id, std::move(attrs)));
+
+    return entity;
+}
+
+ref_ptr<metric_entity> metric_entity_prototype::instantiate(const std::string &id) const
+{
+    metric_entity::attr_map attrs;
+
+    return instantiate(id, attrs);
```
acelyc111, line 49:
```suggestion

    return instantiate(id, {});
```
Can be replaced by this?

src/utils/metrics.cpp
```diff
@@ -0,0 +1,56 @@
...
+}
+
+metric_entity::~metric_entity() {}
+
+ref_ptr<metric_entity> metric_entity_prototype::instantiate(const std::string &id,
+                                                            metric_entity::attr_map attrs) const
+{
+    dassert_f(attrs.find("entity") == attrs.end(), "{}'s attribute \"entity\" is reserved", id);
+
+    attrs["entity"] = _name;
```
acelyc111, line 39: Why add an 'entity' key to indicate its name? where will it be used?

include/dsn/utility/metrics.h
```diff
@@ -0,0 +1,92 @@
...
+// METRIC_DEFINE_entity(my_entity);
+//
+// To use the entity type, declare it at the top of any .h/.cpp file (not within any namespace):
+// METRIC_DECLARE_entity(my_entity);
+//
+// Instantiating the entity in whatever class represents the entity:
+// entity_instance = METRIC_ENTITY_my_entity.instantiate(my_entity_id, ...);
+
+// Define a new entity type.
+#define METRIC_DEFINE_entity(name) ::dsn::metric_entity_prototype METRIC_ENTITY_##name(#name)
```
Shuo-Jia, line 42:
```suggestion
#define METRIC_DEFINE_entity(name) dsn::metric_entity_prototype METRIC_ENTITY_##name(#name)
```

src/utils/metrics.cpp
empiredan: > Why add an 'entity' key to indicate its name? where will it be used?

Later while we are taking snapshots from metrics which will be emitted to the monitoring system such as Prometheus,  all attributes will be set as labels. `entity` is one of them and its value is **server**, **table** or **replica**. Thus it's added here.

include/dsn/utility/metrics.h
empiredan: The prepended scope resolution operator can ensure that `::dsn::metric_entity_prototype` be resolved from global namespace even if  `METRIC_DEFINE_entity` is used within a namespace. Thus I think we can keep it.

src/utils/metrics.cpp
acelyc111: Why not use `_name` directly in this case?

acelyc111 (approved):

Shuo-Jia (approved):

PR #1071, opened 2022-03-09
fix(dup_enhancement#19): change checkpoint status after manual checkpoint completed

Shuo-Jia:
# Ref-Issue

# Change
 `manual checkpoint` status need recovery to `false` when completed.  In the past, the status is updated by [old_decree < last_durable](https://github.com/XiaoMi/rdsn/blob/duplication_dev/src/replica/replica_chkpt.cpp#L144), But once checkpoint may not catch up with `old_decree`, and the following `catch up with task` request will be block. This pr fix it

### DIFF
--- a/src/meta/duplication/meta_duplication_service.cpp
+++ b/src/meta/duplication/meta_duplication_service.cpp
@@ -188,7 +188,11 @@ void meta_duplication_service::do_add_duplication(std::shared_ptr<app_state> &ap
                                                   duplication_info_s_ptr &dup,
                                                   duplication_add_rpc &rpc)
 {
-    dup->start();
+    const auto err = dup->start();
+    if (err != ERR_OK) {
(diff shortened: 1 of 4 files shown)


src/meta/duplication/meta_duplication_service.cpp
```diff
@@ -188,7 +188,11 @@ void meta_duplication_service::do_add_duplication(std::shared_ptr<app_state> &ap
                                                   duplication_info_s_ptr &dup,
                                                   duplication_add_rpc &rpc)
 {
-    dup->start();
+    const auto err = dup->start();
+    if (err != ERR_OK) {
```
levy5307, line 192: dsn_unlikely

hycdong (approved):

levy5307 (approved):

PR #1072, opened 2022-03-10
feat(update_replication_factor#2): add RPC for setting the replication factor of each table

empiredan:
https://github.com/apache/incubator-pegasus/issues/865


src/meta/test/meta_app_operation_test.cpp
```diff
@@ -128,6 +129,33 @@ class meta_app_operation_test : public meta_test_base
...
+
+        return rpc.response();
+    }
+
+    void set_app_and_all_partitions_max_replica_count(const std::string &app_name,
+                                                      int32_t max_replica_count)
+    {
+        auto app = find_app(app_name);
+        dassert_f(app != nullptr, "app({}) does not exist", app_name);
+
```
levy5307, line 151: reduce cast operation
```
uint32_t partition_size =  static_cast<int>(app->partitions.size());
for (int i = 0; i < partition_size; ++i) {
```

levy5307 (approved):

PR #1073, opened 2022-03-11
feat(new_metrics): implement the metric registry

empiredan:
https://github.com/apache/incubator-pegasus/issues/927

### DIFF
--- a/include/dsn/utility/metrics.h
+++ b/include/dsn/utility/metrics.h
@@ -22,8 +22,8 @@
 #include <unordered_map>
 
 #include <dsn/utility/autoref_ptr.h>
-#include <dsn/utility/singleton.h>
 #include <dsn/utility/ports.h>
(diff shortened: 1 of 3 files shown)


include/dsn/utility/metrics.h
```diff
@@ -17,10 +17,12 @@
 
 #pragma once
 
+#include <mutex>
 #include <string>
 #include <unordered_map>
 
 #include <dsn/utility/autoref_ptr.h>
+#include <dsn/utility/singleton.h>
 #include <dsn/utility/ports.h>
```
acelyc111, line 26:
```suggestion
#include <dsn/utility/ports.h>
#include <dsn/utility/singleton.h>
```
Better to keep the headers in lexicographical order

acelyc111 (approved):
LGTM

PR #1074, opened 2022-03-14
feat(verify_file): update the file validation logic for bulk load downloading to avoid io-read workload

GiantKing:
I download files from HDFS and write them to disk with direct I/O. In my test, it got a lot of benefits to IO(avgqu-sz). But compare to the old way(without direct I/O), it causes a lot of reads workload to disk. It confused me a lot. I check the IO cost for every thread, and trace their stacks. Finally I found the read workload is caused by verifying files that downloaded from HDFS. It will read the whole file from disk, and check the md5. When writing files without direct I/O, system will cache the file. So It will not lead to read workload.

So I move the md5-check logic before writing a file, calc the md5 in memory. After the file is written to disk, I just check the file size.

Issues: https://github.com/apache/incubator-pegasus/issues/926

### DIFF
--- a/src/block_service/block_service_manager.cpp
+++ b/src/block_service/block_service_manager.cpp
@@ -112,11 +112,13 @@ static create_file_response create_block_file_sync(const std::string &remote_fil
     return ret;
 }
 
-static download_response
-download_block_file_sync(const std::string &local_file_path, block_file *bf, task_tracker *tracker)
+static download_response download_block_file_sync(const std::string &local_file_path,
+                                                  const std::string &md5,
+                                                  block_file *bf,
+                                                  task_tracker *tracker)
 {
     download_response ret;
-    bf->download(download_request{local_file_path, 0, -1},
+    bf->download(download_request{local_file_path, 0, -1, md5},
                  TASK_CODE_EXEC_INLINED,
                  [&ret](const download_response &resp) { ret = resp; },
                  tracker);
@@ -124,10 +126,20 @@ download_block_file_sync(const std::string &local_file_path, block_file *bf, tas
     return ret;
 }
 
+error_code block_service_manager::download_file(const std::string &remote_dir,
+                                                const std::string &local_dir,
+                                                const std::string &file_name,
+                                                block_filesystem *fs,
+                                                /*out*/ uint64_t &download_file_size)
+{
+    return download_file(remote_dir, local_dir, file_name, "", fs, download_file_size);
+}
+
 // ThreadPool: THREAD_POOL_REPLICATION, THREAD_POOL_DEFAULT
 error_code block_service_manager::download_file(const std::string &remote_dir,
                                                 const std::string &local_dir,
                                                 const std::string &file_name,
+                                                const std::string &md5,
                                                 block_filesystem *fs,
                                                 /*out*/ uint64_t &download_file_size)
 {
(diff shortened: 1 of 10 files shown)


src/block_service/block_service_manager.cpp
```diff
@@ -112,22 +112,34 @@ static create_file_response create_block_file_sync(const std::string &remote_fil
     return ret;
 }
 
-static download_response
-download_block_file_sync(const std::string &local_file_path, block_file *bf, task_tracker *tracker)
+static download_response download_block_file_sync(const std::string &local_file_path,
+                                                  const std::string &md5,
```
Shuo-Jia, line 116: Not all the download need check md5, better pass md5 with default value

src/replica/bulk_load/replica_bulk_loader.cpp
```diff
@@ -453,31 +453,27 @@ void replica_bulk_loader::download_sst_file(const std::string &remote_dir,
...
     if (ec == ERR_PATH_ALREADY_EXIST) {
-        if (utils::filesystem::verify_file(file_name, f_meta.md5, f_meta.size)) {
-            // local file exist and is verified
-            ec = ERR_OK;
-            f_size = f_meta.size;
-            verified = true;
+        // We are not sure if the file was cached by system. And we couldn't
+        // afford the io overhead which is cased by reading file in verify_file(),
+        // so if file exist remove the files directly
+        derror_replica("file({}) exists, try to remove local file "
```
Shuo-Jia, line 464: dwarning

src/utils/filesystem.cpp
```diff
@@ -843,6 +844,43 @@ bool verify_file(const std::string &fname,
...
+    }
+    return true;
+}
+
+bool verify_data(const std::string &fname,
+                 const char *data,
+                 const size_t data_size,
+                 const std::string &expected_md5)
+{
+    std::string md5 = string_md5(data, data_size);
```
Shuo-Jia, line 873: Had you check the `string_md5` result is equal with the origin `md5sum`?

src/utils/filesystem.cpp
```diff
@@ -843,6 +844,43 @@ bool verify_file(const std::string &fname,
...
+        derror_f("verify file({}) size failed, because file damaged, size: {} VS {}",
+                 fname,
+                 f_size,
+                 expected_fsize);
+        return false;
+    }
+    return true;
+}
+
+bool verify_data(const std::string &fname,
```
Shuo-Jia, line 868: verify_data_md5

src/block_service/test/fds_service_test.cpp
```diff
@@ -774,7 +774,7 @@ TEST_F(FDSClientTest, test_concurrent_upload_download)
         for (unsigned int i = 0; i < total_files; ++i) {
             block_file_ptr p = block_files[i];
             dsn::task_ptr t =
-                p->download(download_request{filenames[i] + ".b", 0, -1},
+                p->download(download_request{filenames[i] + ".b", 0, -1, ""},
```
Shuo-Jia, line 777: passing `empty` will not cover your added code, you should add new case to test this pr

hycdong: +1

src/utils/filesystem.cpp
hycdong: +1

Shuo-Jia (approved):

src/utils/filesystem.cpp
```diff
@@ -843,6 +844,43 @@ bool verify_file(const std::string &fname,
     return true;
 }
 
+bool verify_file_size(const std::string &fname, const int64_t &expected_fsize)
```
hycdong, line 847: I suggest that you can refactor function `verify_file` to decrease similar code in your next pull request.

hycdong (approved):

PR #1077, opened 2022-03-16
feat(update_replication_factor#3): support to set the replication factor for all partitions of each table

empiredan:
https://github.com/apache/incubator-pegasus/issues/865

### DIFF
--- a/src/meta/server_state.cpp
+++ b/src/meta/server_state.cpp
@@ -3199,11 +3199,13 @@ void server_state::set_max_replica_count(configuration_set_max_replica_count_rpc
     auto &response = rpc.response();
 
     int32_t app_id = 0;
+    int32_t partition_count = 0;
+    std::shared_ptr<app_state> app;
 
     {
         zauto_read_lock l(_lock);
 
-        auto app = get_app_and_check_exist(app_name, response);
+        app = get_app_and_check_exist(app_name, response);
         if (app == nullptr) {
             response.old_max_replica_count = 0;
             dwarn_f("failed to set max_replica_count: app_name={}, error_code={}, hint_message={}",
@@ -3217,6 +3219,7 @@ void server_state::set_max_replica_count(configuration_set_max_replica_count_rpc
 
         if (!check_max_replica_count_consistent(app, response)) {
             response.old_max_replica_count = 0;
+            response.hint_message = fmt::format("app({}) does not exist", app_name);
             derror_f("failed to set max_replica_count: app_name={}, app_id={}, error_code={}, "
                      "hint_message={}",
                      app_name,
@@ -3227,6 +3230,20 @@ void server_state::set_max_replica_count(configuration_set_max_replica_count_rpc
         }
 
         response.old_max_replica_count = app->max_replica_count;
+
+        if (app->status != app_status::AS_AVAILABLE) {
+            response.err = ERR_INVALID_PARAMETERS;
+            response.hint_message = fmt::format("app({}) is not in available status", app_name);
+            derror_f("failed to set max_replica_count: app_name={}, app_id={}, error_code={}, "
+                     "hint_message={}",
+                     app_name,
+                     app_id,
+                     response.err.to_string(),
+                     response.hint_message);
+            return;
+        }
+
+        partition_count = app->partition_count;
     }
 
     auto level = _meta_svc->get_function_level();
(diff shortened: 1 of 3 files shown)


src/meta/server_state.cpp
```diff
@@ -3269,8 +3269,373 @@ void server_state::set_max_replica_count(configuration_set_max_replica_count_rpc
...
+                                                      configuration_set_max_replica_count_rpc rpc)
+{
+    const auto &app_name = rpc.request().app_name;
+    auto &response = rpc.response();
+
+    zauto_write_lock l(_lock);
+
+    auto app = get_app_and_check_exist(app_name, response);
+    if (app == nullptr) {
+        response.hint_message +=
```
Shuo-Jia, line 3290: no need `+=`

src/meta/server_state.cpp
```diff
@@ -3269,8 +3269,373 @@ void server_state::set_max_replica_count(configuration_set_max_replica_count_rpc
...
+        derror_f("failed to set max_replica_count: app_name={}, error_code={}, hint_message={}",
+                 app_name,
+                 response.err.to_string(),
+                 response.hint_message);
+        return;
+    }
+
+    const auto app_id = app->app_id;
+
+    dassert_f(app->is_stateful,
```
Shuo-Jia, line 3301: if not `main business logic`, such as `read/write`, not suggest execute dassert, which will cause unnecessary downtime of the cluster. You just need to return error to disable this operation

src/meta/server_state.cpp
```diff
@@ -3269,8 +3269,373 @@ void server_state::set_max_replica_count(configuration_set_max_replica_count_rpc
...
+                const auto partition_index = gpid.get_partition_index();
+
+                const auto &app_name = rpc.request().app_name;
+                auto &response = rpc.response();
+
+                zauto_write_lock l(_lock);
+
+                auto app = get_app_and_check_exist(app_name, response);
+                if (app == nullptr) {
+                    response.hint_message += fmt::format(
```
Shuo-Jia, line 3418: same, use `+=` reason?

src/meta/server_state.cpp
```diff
@@ -3269,8 +3269,373 @@ void server_state::set_max_replica_count(configuration_set_max_replica_count_rpc
...
+             app_id,
+             partition_index,
+             new_max_replica_count,
+             new_ballot);
+
+    zauto_write_lock l(_lock);
+
+    auto app = get_app_and_check_exist(app_name, response);
+    if (app == nullptr) {
+        response.hint_message +=
```
Shuo-Jia, line 3490: +1

src/meta/server_state.cpp
```diff
@@ -3269,8 +3269,373 @@ void server_state::set_max_replica_count(configuration_set_max_replica_count_rpc
...
+                 "hint_message={}",
+                 app_name,
+                 app_id,
+                 response.err.to_string(),
+                 response.hint_message);
+        return;
+    }
+
+    // if multiple threads exist in the thread pool, the check may failed
+    dassert_f(
```
Shuo-Jia, line 3503: same, however if the dassert for `transaction` , that is to say, if all previous process need drop via core-dump when persist remote failed, here may be necessary

src/meta/server_state.cpp
```diff
@@ -3269,8 +3269,373 @@ void server_state::set_max_replica_count(configuration_set_max_replica_count_rpc
...
+                context.pending_sync_task =
+                    update_partition_max_replica_count_on_remote(new_partition_config, rpc);
+            },
+            server_state::sStateHash,
+            std::chrono::seconds(1));
+        return;
+    }
+
+    if (ec != ERR_OK) {
+        dassert_f(false, "we can't handle this right now: error_code={}", ec.to_string());
```
Shuo-Jia, line 3547: same

Shuo-Jia:
I find you use many `dassert` to validate the `update` action, I have comment for some. My suggestion is that updating replica count is not a critical read-write path. Its failure will not have a real-time impact on the business. You just need to return an error and tell us that it has not succeeded, rather than downtime the machine

src/meta/server_state.cpp
empiredan: > if not `main business logic`, such as `read/write`, not suggest execute dassert, which will cause unnecessary downtime of the cluster. You just need to return error to disable this operation

Actually `dassert(app->is_stateful)` is put here according to existing code,  such as 
 L1860, `server_state::on_update_configuration`. I think this assertion is just for the stateful table, is it ? On the other hand, it seems that for now there's no stateless table ?

hycdong:
I think it is strange to define two functions called `update_partition_max_replica_count`, and `update_next_partition_max_replica_count`.
You can update each partition's max replica count through while, such as
```
for(auto i = 0; i < partition_count; i++){
    update_partition_max_replica_count(i);
}
```
Besides, I notice that #1072 support update app's max_replica_count, and in this pr, if partition update max_replica_count failed, app's value should be revert. I suggest that you can update the code like:
```
void update_app_max_replica_count(int new_replica_count) {
    // 1. mark updating replica count to avoid repeated request(this flag should store on zk, you can add a new filed in envs)
    // 2. update each partition replica count 
   //  3. after all partition update succeed, then update app replica count and mark not updating replica count
   //  function server_state::do_app_drop and server_state::process_one_partition may be helpful.
}
```

src/meta/server_state.cpp
empiredan: > same, use `+=` reason?

Actually there are 2 cases that a table does not exist. One is the check at the beginning of setting max_replica_count that once the table does not exist, just returning the error is ok since user might have inputed a wrong table name. 

However, another case is that the table becomes disappeared during the process of updating the max_replica_count. Therefore I append some hints here to tell where this exception has happened.

src/meta/server_state.cpp
Shuo-Jia: Yeah,  there's no stateless table. But`server_state::on_update_configuration` is `main logic`, it must make sure current state is correct, so it `dassert` is acceptable. This added feature is not necessary for  `either successful or coredump`,  just return error log message is ok.

src/meta/server_state.cpp
Shuo-Jia: Don' t understand, if one rpc completed,  it fill `hint_message` and return, why exist `append`,  will it happend `two error case` meanwhile

src/meta/server_state.cpp
```diff
@@ -3269,8 +3285,321 @@ void server_state::set_max_replica_count(configuration_set_max_replica_count_rpc
...
+}
+
+// ThreadPool: THREAD_POOL_META_STATE
+void server_state::do_update_max_replica_count(std::shared_ptr<app_state> &app,
+                                               int32_t app_id,
+                                               int32_t partition_count,
+                                               configuration_set_max_replica_count_rpc rpc)
+{
+    std::shared_ptr<std::vector<error_code>> results(new std::vector<error_code>(partition_count));
+    std::shared_ptr<std::atomic_int> updates_in_progress(new std::atomic_int(partition_count));
```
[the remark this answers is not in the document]
empiredan: > Why would you define `results` and `updates_in_progress` as shared_ptr?

Because both of them will be dispatched as callback into `update_partition_max_replica_count` of each partition, to be gathered at last.

src/meta/server_state.cpp
```diff
@@ -3269,8 +3285,321 @@ void server_state::set_max_replica_count(configuration_set_max_replica_count_rpc
...
+        auto &response = rpc.response();
+        response.err = ERR_OK;
+    };
+
+    const auto new_max_replica_count = rpc.request().max_replica_count;
+    update_all_partitions_max_replica_count(app, new_max_replica_count, on_partition_updated);
+}
+
+// ThreadPool: THREAD_POOL_META_STATE
+void server_state::update_all_partitions_max_replica_count(std::shared_ptr<app_state> &app,
```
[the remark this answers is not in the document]
empiredan: > I think this function can be added into function `do_update_max_replica_count`.

Ok.

src/meta/server_state.cpp
```diff
@@ -3269,8 +3285,321 @@ void server_state::set_max_replica_count(configuration_set_max_replica_count_rpc
...
-    // TODO: update partition-level and app-level max_replica_count
-    response.err = ERR_OK;
+    do_update_max_replica_count(app, app_id, partition_count, rpc);
+}
+
+// ThreadPool: THREAD_POOL_META_STATE
+void server_state::do_update_max_replica_count(std::shared_ptr<app_state> &app,
+                                               int32_t app_id,
+                                               int32_t partition_count,
+                                               configuration_set_max_replica_count_rpc rpc)
```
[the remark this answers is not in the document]
empiredan: > `app_id` and `partition_count` are redundant, because it can also get in `app`.

Actually in this function both of `app_id` and `partition_count` are not protected at least by a read lock, so I think we can pass them by parameters ?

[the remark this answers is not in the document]
hycdong: I think you can add read lock to read them in function `do_update_max_replica_count`, your current implementation also need to add read lock in function `set_max_replica_count` and pass them as parameters.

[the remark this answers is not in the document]
empiredan: Ok.

src/meta/server_state.cpp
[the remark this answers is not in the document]
hycdong: You can use the following variety called `partitions_in_progress`, no need to define a shared_ptr called `updates_in_progress`.
https://github.com/XiaoMi/rdsn/blob/b8942ecdd775dd3b22cb67941453ba812c847bc7/src/meta/meta_data.h#L308

As the `results` store the each partition update error code, would you like to do more error handler if any of it is not error_ok?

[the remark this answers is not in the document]
empiredan: > You can use the following variety called `partitions_in_progress`, no need to define a shared_ptr called `updates_in_progress`.
> 
> https://github.com/XiaoMi/rdsn/blob/b8942ecdd775dd3b22cb67941453ba812c847bc7/src/meta/meta_data.h#L308
> 
> As the `results` store the each partition update error code, would you like to do more error handler if any of it is not error_ok?

Yes, this is a synchronization to gather all results, then decide how to process these errors if any, or do response to client though currently dassert(false) is executed if there is any error.

hycdong (approved):

Shuo-Jia (approved):

acelyc111 (approved):

PR #1082, opened 2022-03-25
feat: support aarch64

GehaFearless:
https://github.com/apache/incubator-pegasus/issues/932https://github.com/apache/incubator-pegasus/issues/932


thirdparty/fix_zk_for_aarch64.patch
```diff
@@ -0,0 +1,20 @@
+--- src/c/src/mt_adaptor.c	2022-03-24 11:30:19.018883299 +0000
++++ mt_adaptor.c	2022-03-24 11:30:23.558847831 +0000
+@@ -483,6 +483,9 @@
+ int32_t fetch_and_add(volatile int32_t* operand, int incr)
+ {
+ #ifndef WIN32
++#if defined(__GNUC__)
```
acelyc111, line 7: Should be `#if defined(__aarch64__)`?

Geha-fighting: from https://issues.apache.org/jira/browse/ZOOKEEPER-1374

acelyc111 (dismissed):

PR #1086, opened 2022-04-11
feat: add deny table env to reject client read/write request

foreverneverer:
# Ref-Issue
https://github.com/apache/incubator-pegasus/issues/941

# Change
- I refactor `replica.deny_client_write` to `replica.deny_client_request` to supprt both `read` and `write`
- I add `timeout/reconfig` argument to decide if response `ERR_INVLIDE_STATE`or just `return`. 
- `ERR_INVLIDE_STATE` will trigger client update config  which will used for `trigger to switch master cluster and follower cluster  by server side under duplication state`
- `return` will avoid client retry frequently(the case is same as [origin design](https://github.com/XiaoMi/rdsn/blob/289eb4609ae781e3bcc6bbc6bfeae64dfd8fa785/src/replica/replica_2pc.cpp#L57))

# Usage
```shell
set_app_envs replica.deny_client reconfig*write # deny write and response `ERR_INVLIDE_STATE`
set_app_envs replica.deny_client reconfig*read # deny read and response `ERR_INVLIDE_STATE`

set_app_envs replica.deny_client timeout*write # deny write but not response any err code
set_app_envs replica.deny_client timeout*read # deny read but not response any err code

set_app_envs replica.deny_client delay*all # deny all but not response any err code
set_app_envs replica.deny_client reconfig*all # deny all  and response `ERR_INVLIDE_STATE`
```
# Test
- manual test via java-client
- unit test


src/meta/app_env_validator.cpp
```diff
@@ -46,6 +46,27 @@ bool check_slow_query(const std::string &env_value, std::string &hint_message)
...
+    utils::split_args(env_value.c_str(), sub_sargs, '*', true);
+
+    std::string invalid_hint_message = "Invalid deny client args, valid include: reject*all, "
+                                       "reject*write, reject*read; delay*all, delay*write, "
+                                       "delay*read";
+    if (sub_sargs.size() != 2) {
+        hint_message = invalid_hint_message;
+        return false;
+    }
+    if (sub_sargs[0].empty() || sub_sargs[1].empty() ||
```
acelyc111, line 61: `sub_sargs[0].empty() || sub_sargs[1].empty() ||` is not necessary while you have checked their values.

src/meta/app_env_validator.cpp
```diff
@@ -158,9 +179,23 @@ void app_env_validator::register_all_validators()
...
+         std::bind(&utils::token_bucket_throttling_controller::validate,
+                   std::placeholders::_1,
+                   std::placeholders::_2)},
+        {replica_envs::SPLIT_VALIDATE_PARTITION_HASH,
+         std::bind(&check_bool_value, std::placeholders::_1, std::placeholders::_2)},
+        {replica_envs::USER_SPECIFIED_COMPACTION, nullptr},
+        {replica_envs::BACKUP_REQUEST_QPS_THROTTLING,
+         std::bind(&check_throttling, std::placeholders::_1, std::placeholders::_2)},
+        {replica_envs::ROCKSDB_ALLOW_INGEST_BEHIND,
+         std::bind(&check_bool_value, std::placeholders::_1, std::placeholders::_2)},
```
acelyc111, line 194: Is this related to this patch? If not, better to seperate it to another PR, it would be much easier to track the relevant changes and review the code.

acelyc111:
> ERR_INVLIDE_STATE will trigger client update config which will used for trigger to switch master cluster and follower cluster by server side under duplication state

ERR_INVLIDE_STATE has been used in some other scenarios, better to distinguish them and add another error code.

src/meta/test/meta_app_envs_test.cpp
```diff
@@ -140,7 +140,18 @@ TEST_F(meta_app_envs_test, update_app_envs_test)
          "",
          "200"},
         {replica_envs::BUSINESS_INFO, "300", ERR_OK, "", "300"},
-        {replica_envs::DENY_CLIENT_WRITE, "400", ERR_OK, "", "400"},
+        {replica_envs::DENY_CLIENT,
+         "400",
+         ERR_INVALID_PARAMETERS,
+         "Invalid deny client args, valid include: reject*all, reject*write, reject*read; "
+         "delay*all, delay*write, delay*read",
+         "400"},
```
acelyc111, line 148: There are some paths in `check_deny_client` not covered, better to add more test cases.

src/replica/replica_config.cpp
```diff
@@ -626,6 +626,25 @@ void replica::update_allow_ingest_behind(const std::map<std::string, std::string
...
+{
+    auto env_iter = envs.find(replica_envs::DENY_CLIENT);
+    if (env_iter == envs.end()) {
+        _deny_client.reset();
+        return;
+    }
+
+    std::vector<std::string> sub_sargs;
+    utils::split_args(env_iter->second.c_str(), sub_sargs, '*', true);
+    if (sub_sargs.size() != 2) {
```
acelyc111, line 639: It has been checked by `check_deny_client`, so just check_eq is OK?

src/replica/test/replica_test.cpp
```diff
@@ -94,6 +94,15 @@ class replica_test : public replica_test_base
             envs, replica_envs::ROCKSDB_ALLOW_INGEST_BEHIND, _mock_replica->_allow_ingest_behind);
     }
 
+    const deny_client &update_deny_client(const std::string &env_name, std::string env_value)
+    {
+
```
acelyc111, line 99: remove blank lines

src/replica/replica.h
```diff
@@ -100,6 +100,25 @@ bool get_bool_envs(const std::map<std::string, std::string> &envs,
                    const std::string &name,
                    /*out*/ bool &value);
 
+struct deny_client
+{
+    bool read;
+    bool write;
+    bool response;
```
acelyc111, line 107: Better to add some comments for `response`, it's bit of confused.

acelyc111, line 107: Also need initial value.

src/common/replication_common.cpp
```diff
@@ -591,7 +591,7 @@ replication_options::check_if_in_black_list(const std::vector<std::string> &blac
     return false;
 }
 
-const std::string replica_envs::DENY_CLIENT_WRITE("replica.deny_client_write");
+const std::string replica_envs::DENY_CLIENT("replica.deny_client");
```
acelyc111, line 594: Keep in mind this will cause imcompatiable issue, there may be some community users use `replica.deny_client_write`.

acelyc111:
> > ERR_INVLIDE_STATE has been used in some other scenarios, better to distinguish them and add another error code.
> 
> Yeah，but it need also modify client and require user to update client version, for simplicity, I re-use the err-code though it has some ambiguous

How the client switch to another cluster when got ERR_INVLIDE_STATE?

src/meta/app_env_validator.cpp
```diff
@@ -46,6 +46,27 @@ bool check_slow_query(const std::string &env_value, std::string &hint_message)
...
 }
 
+bool check_deny_client(const std::string &env_value, std::string &hint_message)
+{
+    std::vector<std::string> sub_sargs;
+    utils::split_args(env_value.c_str(), sub_sargs, '*', true);
+
+    std::string invalid_hint_message = "Invalid deny client args, valid include: reject*all, "
+                                       "reject*write, reject*read; delay*all, delay*write, "
+                                       "delay*read";
```
hycdong, line 56: I notice that the `delay` policy is not really delay, it is not return any response to client, both of them are reject, the difference is whether return error code to client to trigger query meta. I recommend not using delay and rename it.

foreverneverer: `return` no resp will cause `ERR_TIMEOUT`, it is like `delay some seconds until timeout` for client side, so I use it and no find other name

src/common/replication_common.cpp
foreverneverer: Yeah, I ever consider whether keep it. Maybe I need add some `hint_message` when user use `replica.deny_client_write`

src/meta/app_env_validator.cpp
foreverneverer: I just move the method position, these are commented `no implement check` in the past: https://github.com/XiaoMi/rdsn/blob/v2.3.0/src/meta/app_env_validator.cpp#L170

foreverneverer:
> It's too implict when user set 'reject' config to a table then the client will re-fetch config map, if there is no metaproxy, then the meta server will get flood request?

The reason for this feature is that the server side cannot switch master/slave when user uses metaproxy. In the past, I needed drop table to trigger the re-fetch of the client, which is inelegant.

But consider implementation, I hope it is more general than just used in this scenario. For users who do not use master/slave, they do not need to pay attention to these implict details

acelyc111:
> > It's too implict when user set 'reject' config to a table then the client will re-fetch config map, if there is no metaproxy, then the meta server will get flood request?
> 
> The reason for this feature is that the server side cannot switch master/slave when user uses metaproxy. In the past, I needed drop table to trigger the re-fetch of the client, which is inelegant.
> 
> But consider implementation, I hope it is more general than just used in this scenario. For users who do not use master/slave, they do not need to pay attention to these implict details

There are some scenarios we want to deny read/write to a table, for example, migrate a table to another cluster, but not wanna clients re-fetch config.
Re-fetch config from meta server in high frequent (near read/write ops/sec) may cause some problems.
If you want clients fetch config, use some other method, I suggest not to mix with "reject" or "delay", for example, add another env.

foreverneverer:
> for example, add another env.

OK, add another env seem be better

acelyc111 (approved):

hycdong (approved):

PR #1095, opened 2022-04-19
feat(new_metrics): make counter monotonic

empiredan:
https://github.com/apache/incubator-pegasus/issues/951


include/dsn/utility/metrics.h
```diff
@@ -364,9 +401,9 @@ class counter : public metric
         return _adder.fetch_and_reset();
     }
 
+    // NOTICE: x MUST be positive number.
```
acelyc111, line 404: You can use a dassert for this

acelyc111 (approved):

PR #1099, opened 2022-04-26
fix: crash when huge write

ZhongChaoqiang:
# Related-Issue
https://github.com/apache/incubator-pegasus/issues/903

# Change
when buffer of slog is more 2GB, then flush to slog file, it maybe crash.

### DIFF
--- a/src/aio/disk_engine.cpp
+++ b/src/aio/disk_engine.cpp
@@ -53,7 +53,7 @@ static disk_engine_initializer disk_engine_init;
 aio_task *disk_write_queue::unlink_next_workload(void *plength)
 {
     uint64_t next_offset = 0;
-    uint32_t &sz = *(uint32_t *)plength;
+    uint64_t &sz = *(uint64_t *)plength;
     sz = 0;
 
     aio_task *first = _hdr._first, *current = first, *last = first;
@@ -126,9 +126,9 @@ aio_task *disk_file::on_write_completed(aio_task *wk, void *ctx, error_code err,
         if (err == ERR_OK) {
             size_t this_size = (size_t)wk->get_aio_context()->buffer_size;
             dassert(size >= this_size,
-                    "written buffer size does not equal to input buffer's size: %d vs %d",
-                    (int)size,
-                    (int)this_size);
+                    "written buffer size does not equal to input buffer's size: %lld vs %lld",
+                    size,
+                    this_size);
 
             wk->enqueue(err, this_size);
             size -= this_size;
@@ -167,7 +167,7 @@ public:
     virtual void exec() override
     {
         auto df = (disk_file *)_tasks->get_aio_context()->file_object;
-        uint32_t sz;
+        uint64_t sz;
 
         auto wk = df->on_write_completed(_tasks, (void *)&sz, error(), get_transferred_size());
         if (wk) {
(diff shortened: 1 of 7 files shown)


foreverneverer (approved):

src/aio/disk_engine.cpp
```diff
@@ -126,9 +126,9 @@ aio_task *disk_file::on_write_completed(aio_task *wk, void *ctx, error_code err,
         if (err == ERR_OK) {
             size_t this_size = (size_t)wk->get_aio_context()->buffer_size;
             dassert(size >= this_size,
```
acelyc111, line 128: Better to use `dcheck_ge` to simplify code.

acelyc111:
@ZhongChaoqiang The CI failed at https://github.com/XiaoMi/rdsn/runs/6189338284?check_suite_focus=true#step:8:1807, plz check it.

ZhongChaoqiang:
ok, I’m checking it

从 Windows 版邮件<https://go.microsoft.com/fwlink/?LinkId=550986>发送

发件人: Yingchun ***@***.***>
发送时间: 2022年4月27日 17:22
收件人: ***@***.***>
抄送: Zhong ***@***.***>; ***@***.***>
主题: Re: [XiaoMi/rdsn] fix: crash when huge write (PR #1099)


@ZhongChaoqiang<https://github.com/ZhongChaoqiang> The CI failed at https://github.com/XiaoMi/rdsn/runs/6189338284?check_suite_focus=true#step:8:1807, plz check it.

―

acelyc111:
@ZhongChaoqiang 
See details [here](https://github.com/XiaoMi/rdsn/runs/6189338415?check_suite_focus=true#step:8:1834), it tells you what's the problem is.
The variable in src/aio/disk_engine.cpp L269 is `uint32_t sz;`, but in src/aio/disk_engine.cpp L56, we want to dereference it as an uint64_t (`uint64_t &sz = *(uint64_t *)plength;`). So stack-buffer-overflow error occurs, and may cause following coredump.

src/aio/disk_engine.cpp
ZhongChaoqiang: done

ZhongChaoqiang:
> @ZhongChaoqiang See details [here](https://github.com/XiaoMi/rdsn/runs/6189338415?check_suite_focus=true#step:8:1834), it tells you what the problem is. The variable in src/aio/disk_engine.cpp L269 is `uint32_t sz;`, but in src/aio/disk_engine.cpp L56, we want to dereference it as an uint64_t (`uint64_t &sz = *(uint64_t *)plength;`). So stack-buffer-overflow error occurs, and may cause following coredump.

Thanks very much

acelyc111 (approved):

acelyc111:
Thanks for your contribution!
Could you describe how do you troubleshooting the bug you found when write with huge thoughput, and create a specify issue?

xihong08:
> when to be written data size more 4GB,  it maybe happened on the server side in busying write status，crash  happened in aio_task::collapse  ::memcpy(dest, b.buffer, b.size);  this case resulting by memory overwriting .  and root cause is integer overflow，coz aio_context's buffer_size is 4 bytes,  in dsn::file::write_vector file_io.cpp L128 , the aio_context's buffer_size may be overflow when to be written data size more 4GB,  so crash it later.

xihong08:
> Thanks for your contribution! Could you describe how do you troubleshooting the bug you found when write with huge thoughput, and create a specify issue?
    i can describe detailed case.   this maybe happened on the server side in busying write status，and crash happened in aio_task::collapse ::memcpy(dest, b.buffer, b.size);  it resulting by memory overwriting . and root cause is integer overflow，coz aio_context's buffer_size is 4 bytes, in dsn::file::write_vector file_io.cpp L128 , the aio_context's buffer_size may be overflow when to be written data size more 4GB, so crash it later.

foreverneverer (approved):

PR #1103, opened 2022-05-11
feat(bulk_load): support clear last bulk load state rpc

hycdong:
[Pegasus#872](https://github.com/apache/incubator-pegasus/issues/872).
Meta server will store last bulk load state for client to query last bulk load result, this pull request adds a clear rpc interface to clear  this state.


acelyc111 (approved):
LGTM
Is there a mechanism that the bulk load state can be auto sweep up?

hycdong:
> LGTM Is there a mechanism that the bulk load state can be auto sweep up?

Meta server will reserve last bulk load state, including success and failure, and will clear it when next bulk load process is triggered. I think it is a way to auto sweep up state. This rpc provides the manual interface, which is facility for user-tools.
In our production environment, user will always retry whole bulk load process if they got error, which will sometimes lead to performance downgrade. As a result, our user tool will restrict error retry time, and new bulk load can not be triggered through tools, can only be triggered by admin-cli. To make it more flexible, we finally decide to add this manual interface.

foreverneverer (approved):

PR #1105, opened 2022-05-17
feat(bulk_load): support different tables can execute bulk load concurrently

hycdong:
[Pegasus#872](https://github.com/apache/incubator-pegasus/issues/872)
In previous implementation, we restrict only one table can execute bulk load in cluster to guarantee performance. We have already added more restriction to control bulk load download and ingestion, it is safety to permit concurrent bulk load. This pull request adds a gflags to control it.
```diff
[meta_server]
+ enable_concurrent_bulk_load = false
```

### DIFF
--- a/src/meta/meta_bulk_load_service.cpp
+++ b/src/meta/meta_bulk_load_service.cpp
@@ -40,6 +40,12 @@ DSN_DEFINE_bool("meta_server",
                 "verify files according to metadata before ingest");
 DSN_TAG_VARIABLE(bulk_load_verify_before_ingest, FT_MUTABLE);
 
+DSN_DEFINE_bool("meta_server",
+                concurrent_bulk_load,
+                false,
+                "different app can execute bulk load at the same time if this value is true");
+DSN_TAG_VARIABLE(concurrent_bulk_load, FT_MUTABLE);
+
 bulk_load_service::bulk_load_service(meta_service *meta_svc, const std::string &bulk_load_dir)
     : _meta_svc(meta_svc), _state(meta_svc->get_server_state()), _bulk_load_root(bulk_load_dir)
 {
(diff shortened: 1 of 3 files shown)


foreverneverer (approved):

src/meta/meta_bulk_load_service.cpp
```diff
@@ -40,6 +40,12 @@ DSN_DEFINE_bool("meta_server",
                 "verify files according to metadata before ingest");
 DSN_TAG_VARIABLE(bulk_load_verify_before_ingest, FT_MUTABLE);
 
+DSN_DEFINE_bool("meta_server",
+                concurrent_bulk_load,
```
acelyc111, line 44: How about naming like 'enable_concurrent_bulk_load' ?

src/meta/meta_bulk_load_service.cpp
```diff
@@ -40,6 +40,12 @@ DSN_DEFINE_bool("meta_server",
...
+                false,
+                "different app can execute bulk load at the same time if this value is true");
```
acelyc111, line 46: How about:
'whether to enable different apps to execute bulk load at the same time' ?

src/meta/meta_bulk_load_service.cpp
```diff
@@ -68,7 +74,8 @@ void bulk_load_service::on_start_bulk_load(start_bulk_load_rpc rpc)
     auto &response = rpc.response();
     response.err = ERR_OK;
 
-    if (!_meta_svc->try_lock_meta_op_status(meta_op_status::BULKLOAD)) {
+    if (!FLAGS_enable_concurrent_bulk_load &&
+        !_meta_svc->try_lock_meta_op_status(meta_op_status::BULKLOAD)) {
```
GiantKing, line 78: Bulkload process will be continued when meta failover, and it will lock meta_op_status too in func: do_continue_app_bulk_load. Add this check?

hycdong: Done

acelyc111 (approved):

GiantKing:
LGTM

foreverneverer (approved):

PR #1110, opened 2022-05-25
feat(update_replication_factor#8): recover uncompleted update of replication factor during initialization of meta service

empiredan:
This is a subtask of https://github.com/apache/incubator-pegasus/issues/865.

Since in https://github.com/XiaoMi/rdsn/pull/1087 we've supported to set env for the replication factor in case there are duplicate requests trying to update the same table, there is still a problem that the request for updating replication factor may fail once meta server crashes. It's possible that there are partial steps finished while others are uncompleted.

Therefore, After meta server has got primary lock, it will initialize, during which recovering replication factor should be relaunched.  Any table whose env of updating replication factor is not removed (once updating replication factor is complete, the env will be eliminated) will be the candidate that needs to recover as long as it's available. All tables should have been restored before the initialization is finished.

### DIFF
--- a/src/meta/server_state.cpp
+++ b/src/meta/server_state.cpp
@@ -3736,5 +3736,108 @@ void server_state::update_partition_max_replica_count_locally(
              new_config_str);
 }
 
+void server_state::recover_from_max_replica_count_env()
+{
+    std::vector<std::pair<std::shared_ptr<app_state>, int32_t>> tasks;
+    {
+        zauto_read_lock l(_lock);
+        for (auto &e : _exist_apps) {
+            auto &app = e.second;
+            auto iter = app->envs.find(replica_envs::UPDATE_MAX_REPLICA_COUNT);
+            if (iter == app->envs.end()) {
+                continue;
+            }
+
+            std::vector<std::string> args;
+            utils::split_args(iter->second.c_str(), args, ';');
+            if (args[0] != "updating") {
+                continue;
+            }
+
+            int32_t max_replica_count = 0;
+            if (args.size() < 2 || !dsn::buf2int32(args[1], max_replica_count) || max_replica_count <= 0) {
+                dassert_f(false, "invalid max_replica_count_env: app_name={}, app_id={}, max_replica_count={}, {}={}", app->app_name, app->app_id, app->max_replica_count, iter->first, iter->second);
+            }
+
+            tasks.emplace_back(app, max_replica_count);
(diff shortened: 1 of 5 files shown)


src/meta/server_state.cpp
```diff
@@ -3736,5 +3734,197 @@ void server_state::update_partition_max_replica_count_locally(
...
+            }
+
+            auto iter = app->envs.find(replica_envs::UPDATE_MAX_REPLICA_COUNT);
+            if (iter == app->envs.end()) {
+                continue;
+            }
+
+            std::vector<std::string> args;
+            utils::split_args(iter->second.c_str(), args, ';');
+            if (args[0] != "updating") {
```
acelyc111, line 3756: Judge !args.empty() firstly.

src/meta/server_state.cpp
```diff
@@ -3736,5 +3734,197 @@ void server_state::update_partition_max_replica_count_locally(
...
+            int32_t max_replica_count = 0;
+            if (args.size() < 2 || !dsn::buf2int32(args[1], max_replica_count) ||
+                max_replica_count <= 0) {
+                dassert_f(false,
+                          "invalid max_replica_count_env: app_name={}, app_id={}, "
+                          "max_replica_count={}, {}={}",
+                          app->app_name,
+                          app->app_id,
+                          app->max_replica_count,
+                          iter->first,
```
acelyc111, line 3769: it must be `replica_envs::UPDATE_MAX_REPLICA_COUNT`

src/meta/server_state.cpp
```diff
@@ -3736,5 +3734,197 @@ void server_state::update_partition_max_replica_count_locally(
...
+            tasks.emplace_back(app, max_replica_count);
+        }
+    }
+
+    dsn::task_tracker tracker;
+
+    for (auto &task : tasks) {
+        recover_all_partitions_max_replica_count(task.first, task.second, tracker);
+    }
+    tracker.wait_outstanding_tasks();
```
acelyc111, line 3782: How long will it cost, if there is any bug cause the tasks never complete, the meta server will not startup normally, right?

src/meta/server_state.cpp
empiredan: > Judge !args.empty() firstly.

OK

acelyc111 (approved):

hycdong (approved):

PR #1112, opened 2022-06-01
feat(new_metrics): implement the percentile

empiredan:
https://github.com/apache/incubator-pegasus/issues/991


src/utils/metrics.cpp
```diff
@@ -100,4 +111,46 @@ metric_prototype::~metric_prototype() {}
...
+    auto initial_interval_ms = get_initial_interval_ms(_interval_ms);
+
+    _timer->expires_from_now(boost::posix_time::milliseconds(initial_interval_ms));
+    _timer->async_wait(std::bind(&percentile_timer::on_timer, this, std::placeholders::_1));
+}
+
+void percentile_timer::on_timer(const boost::system::error_code &ec)
+{
+    // as the callback is not in tls context, so the log system calls like ddebug, dassert will
+    // cause a lock
```
acelyc111, line 142: so what?

acelyc111, line 142: so what?

src/utils/metrics.cpp
```diff
@@ -100,4 +111,46 @@ metric_prototype::~metric_prototype() {}
...
+    if (!ec) {
```
acelyc111, line 143: dsn_likely

acelyc111, line 143: dsn_likely

src/utils/shared_io_service.cpp
```diff
@@ -0,0 +1,52 @@
...
+// under the License.
+
+#include "shared_io_service.h"
+
+#include <dsn/utility/flags.h>
+
+namespace dsn {
+namespace tools {
+
+// Original default value for timer_service_worker_count is set to 2, and new default value is
```
acelyc111, line 25: No need to mention 'original' balabala, after this patch been merged, 'original' seems confused.

src/utils/shared_io_service.cpp
```diff
@@ -0,0 +1,52 @@
...
+// monitoring systems, while another two threads are still used for computing percentiles.
+DSN_DEFINE_uint32("core", timer_service_worker_count, 3, "the number of threads for timer service");
+DSN_DEFINE_validator(timer_service_worker_count,
+                     [](uint32_t worker_count) -> bool { return worker_count > 0; });
+
+shared_io_service::shared_io_service()
+{
+    _workers.reserve(FLAGS_timer_service_worker_count);
+    for (uint32_t i = 0; i < FLAGS_timer_service_worker_count; ++i) {
+        _workers.emplace_back(new std::thread([this]() {
```
acelyc111, line 36: How to deal the situation when it throw exception in some cases?

include/dsn/utility/alloc.h
```diff
@@ -0,0 +1,68 @@
...
+                  "unaligned array element for cache line: array={}, length={}, index={}, elem={}, "
+                  "elem_size={}, mask={}, cacheline_size={}",
+                  fmt::ptr(array),
+                  len,
+                  i,
+                  fmt::ptr(elem),
+                  sizeof(T),
+                  sizeof(T) - 1,
+                  CACHELINE_SIZE);
+    }
```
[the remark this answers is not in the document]
empiredan: > Can we only validate it in DEBUG version?

OK, I think `#ifndef NDEBUG` can be used here.

include/dsn/utility/metrics.h
```diff
@@ -447,4 +479,244 @@ using concurrent_volatile_counter_ptr = counter_ptr<concurrent_long_adder, true>
...
+        dassert_f(type < kth_percentile_type::COUNT,
+                  "kth_percentile_type {} should be < {}",
+                  static_cast<size_t>(type),
+                  static_cast<size_t>(kth_percentile_type::COUNT));
+
+        val = _full_nth_elements[static_cast<size_t>(type)].load(std::memory_order_relaxed);
+        return _kth_percentile_bitset.test(static_cast<size_t>(type));
+    }
+
+    static const size_type kDefaultSampleSize = 5000;
```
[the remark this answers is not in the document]
empiredan: > How about use a value of 2^n as the default sample size? Then we can use bit operator `count & (2^n -1)` for quick calculate the position.

OK, we can restrict the sample size to be power of 2. Therefore the default value will be set to 4096.

src/utils/alloc.cpp
```diff
@@ -0,0 +1,64 @@
...
+              "allocated nullptr by posix_memalign: addr={}, size={}, cacheline_size={}",
+              fmt::ptr(buffer),
+              size,
+              CACHELINE_SIZE);
+    dassert_f((reinterpret_cast<const uintptr_t>(buffer) & (CACHELINE_SIZE - 1)) == 0,
+              "unaligned cache line: addr={}, size={}, cacheline_size={}, mask={}",
+              fmt::ptr(buffer),
+              size,
+              CACHELINE_SIZE,
+              CACHELINE_SIZE - 1);
```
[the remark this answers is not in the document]
empiredan: > If `err` is 0, is it necessary to check these two situations?

Actually upon successful completion of `posix_memalign`, `buffer` shall be a multiple of  `CACHELINE_SIZE`.
On the other hand, `buffer` cannot be `nullptr` provided that `size` is non-zero. Therefore these two assertions can be removed.

include/dsn/utility/metrics.h
```diff
@@ -447,4 +479,244 @@ using concurrent_volatile_counter_ptr = counter_ptr<concurrent_long_adder, true>
...
+    // otherwise, it will always return true with the value corresponding to `type`.
+    bool get(kth_percentile_type type, value_type &val) const
+    {
+        dassert_f(type < kth_percentile_type::COUNT,
+                  "kth_percentile_type {} should be < {}",
+                  static_cast<size_t>(type),
+                  static_cast<size_t>(kth_percentile_type::COUNT));
+
+        val = _full_nth_elements[static_cast<size_t>(type)].load(std::memory_order_relaxed);
+        return _kth_percentile_bitset.test(static_cast<size_t>(type));
```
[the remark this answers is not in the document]
empiredan: > `static_cast<size_t>(type)` called several times in this function, how about define a local variable?

Sure.

src/utils/alloc.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+#include <dsn/utility/safe_strerror_posix.h>
+
+namespace dsn {
+
+#ifdef CACHELINE_SIZE
+
+/* extern */ void *cacheline_aligned_alloc(size_t size)
+{
+    if (size == 0) {
```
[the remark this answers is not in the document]
empiredan: > dsn_unlikely? Or dcheck_gt?

Since it's allowed for `posix_memalign` that `size` can be zero and NULL may be returned, we can use `dsn_unlikely` here.

include/dsn/utility/metrics.h
```diff
@@ -447,4 +479,244 @@ using concurrent_volatile_counter_ptr = counter_ptr<concurrent_long_adder, true>
...
+          _samples(cacheline_aligned_alloc_array<value_type>(sample_size)),
+          _tail(0),
+          _kth_percentile_bitset(),
+          _full_nth_elements(static_cast<size_t>(kth_percentile_type::COUNT)),
+          _nth_element_finder(),
+          _timer()
+    {
+        dassert(sample_size > 0, "sample_sizes should be > 0");
+        dassert(_samples, "_samples should be valid pointer");
+        std::fill(_samples.get(), _samples.get() + _sample_size, value_type{});
```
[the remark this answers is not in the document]
empiredan: > How about move it to template function `cacheline_aligned_ptr<T> cacheline_aligned_alloc_array(size_t len)` ？

Good idea, I've moved this to `template <typename T> cacheline_aligned_ptr<T> cacheline_aligned_alloc_array(size_t len, const T &val)`.

src/utils/shared_io_service.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include "shared_io_service.h"
+
+#include <dsn/utility/flags.h>
+
+namespace dsn {
+namespace tools {
+
+DSN_DEFINE_uint32("core", timer_service_worker_count, 3, "the number of threads for timer service");
+DSN_DEFINE_validator(timer_service_worker_count, [](uint32_t worker_count) -> bool {
+    if (worker_count < 3) {
```
levy5307, line 27: add a const variable for '3'. Avoid to use immediate var

include/dsn/utility/metrics.h
```diff
@@ -447,4 +479,268 @@ using concurrent_volatile_counter_ptr = counter_ptr<concurrent_long_adder, true>
...
+
+std::set<kth_percentile_type> get_all_kth_percentile_types()
+{
+    std::set<kth_percentile_type> all_types;
+    for (size_t i = 0; i < static_cast<size_t>(kth_percentile_type::COUNT); ++i) {
+        all_types.insert(static_cast<kth_percentile_type>(i));
+    }
+    return all_types;
+}
+const std::set<kth_percentile_type> kAllKthPercentileTypes = get_all_kth_percentile_types();
```
levy5307, line 529: what about unordered_set ？

levy5307 (approved):

src/utils/shared_io_service.cpp
empiredan: > 

OK, I'll define a const variable to represent the minimum `timer_service_worker_count.

include/dsn/utility/metrics.h
empiredan: > what about unordered_set ？

Yes, `std::unordered_set` is a choice. The container for `kth_percentiles` is used to stored all the selected kth percentiles. Since elements in `std::set` are not only unique, but also ordered, we can adopt `std::set` firstly.

